PR #31, opened 2020-05-11
Added series description to seg

CPBridge:
Although optional per the standard, the Series Description is commonly used for a variety of practical purposes. E.g. it is displayed underneath the thumbnail in viewers such as OHIF.

Currently the user can add the series description after construction of the dataset, but since it is so common, I believe that this should be handled as an optional argument of segmentation object.

The only question I have is should this be done at the level of the base SOPClass, to allow other SOPs to benefit from it?

### DIFF
--- a/src/highdicom/base.py
+++ b/src/highdicom/base.py
@@ -54,7 +54,8 @@ class SOPClass(Dataset):
             ] = None,
             coding_schemes: Optional[
                 Sequence[CodingSchemeIdentificationItem]
-            ] = None
+            ] = None,
+            series_description: Optional[str] = None
         ):
         """
         Parameters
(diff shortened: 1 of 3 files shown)


CPBridge:
If we do not do this at the level of the SOPClass, I guess I should do the same for the other SOPs

hackermd:
> Currently the user can add the series description after construction of the dataset, but since it is so common, I believe that this should be handled as an optional argument of segmentation object.

Generally, I think we should not just add more and more parameters for optional attributes to the constructor. However, in the case of *Series Description* I agree that it would add value from a user perspective (although one could also argue that one should discourage the use of Series Description, since it's often abused).

Since it's so commonly used, I would suggest adding the parameter to the constructor of the base `SOPClass` instead.

CPBridge:
I definitely agree that we can't just add all optional attributes to the constructor. But like it or loath it series description is used by nearly every viewer to identify the series to the user.

I have moved to the base `SOPClass` as suggested, so now the derived class constructors can just pass it through in the `kwargs`

PR #34, opened 2020-06-30
Multiframe Conversion - JUST TO INFORM NOT FOR MERGE

afshinmessiah:
Dear Markus, 
I've developed a new extensible conversion engine for HighDicom based David Clunie's PixelMed. This piece of code will be our meeting's main topic on July 9th. 
Although it is not necessary, it would be wonderful if you take a look at the structure of the new conversion.   This [file](https://github.com/MGHComputationalPathology/highdicom/files/4854681/pkg01.pdf) also contains the UML class diagram of the new code to get a broader focus on it.

### DIFF
--- a/src/highdicom/legacy/sop.py
+++ b/src/highdicom/legacy/sop.py
@@ -11,7 +11,7 @@ from pydicom.tag import Tag
 from pydicom.dataelem import DataElement
 from pydicom.sequence import Sequence as DicomSequence
 from pydicom.multival import MultiValue
-from datetime import date, datetime, time
+from datetime import date, datetime, time, timedelta
 from pydicom.valuerep import DT, DA, TM
 from copy import deepcopy
 
@@ -309,17 +309,13 @@ def _convert_legacy_to_enhanced(
         except AttributeError:
             pass
 
-        # Cardiac Synchronization (U)
-        # TODO: http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.16.2.html#sect_C.7.6.16.2.7  # noqa
+        # Cardiac Synchronization (U                # TODO: http:#dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.16.2.html#sect_C.7.6.16.2.7  # noqa
 
-        # Contrast/Bolus Usage (U) - MR/CT only
-        # TODO: http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.16.2.html#sect_C.7.6.16.2.12  # noqa
+        # Contrast/Bolus Usage (U) - MR/CT onl              # TODO: http:#dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.16.2.html#sect_C.7.6.16.2.12  # noqa
 
-        # Respiratory Synchronization (U)
-        # TODO: http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.16.2.html#sect_C.7.6.16.2.17  # noqa
+        # Respiratory Synchronization (U                # TODO: http:#dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.16.2.html#sect_C.7.6.16.2.17  # noqa
 
-        # Real World Value Mapping (U) - PET only
-        # TODO: http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.16.2.html#sect_C.7.6.16.2.11  # noqa
+        # Real World Value Mapping (U) - PET onl                # TODO: http:#dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.16.2.html#sect_C.7.6.16.2.11  # noqa
 
         perframe_items.append(perframe_item)
 
@@ -646,6 +642,9 @@ class Abstract_MultiframeModuleAdder(ABC):
         self.EarliestDate = DA('00010101')
         self.EarliestTime = TM('000000')
         self.EarliestDateTime = DT('00010101000000')
+        self.FarthestFutureDate = DA('99991231')
+        self.FarthestFutureTime = TM('235959')
+        self.FarthestFutureDateTime = DT('99991231235959')
     def _is_empty_or_empty_items(self, attribute:DataElement)->bool:
         if attribute.is_empty:
             return True
@@ -657,7 +656,7 @@ class Abstract_MultiframeModuleAdder(ABC):
                     v = item[tg]
                     if not self._is_empty_or_empty_items(v):
                         return False
-        return True
+        return False
                    
     def _mark_tag_as_used(self, tg:Tag):
         if tg in self.SharedTags:
@@ -682,14 +681,14 @@ class Abstract_MultiframeModuleAdder(ABC):
         if src_kw_or_tg in src_ds:
             elem = src_ds[src_kw_or_tg]
             if check_not_to_be_empty:
-                if _is_empty_or_empty_items(elem):
+                if self._is_empty_or_empty_items(elem):
                     return 
             new_elem = deepcopy(elem)
             if dest_kw_or_tg == src_kw_or_tg:
                 dest_ds[dest_kw_or_tg] = new_elem
             else:
                 new_elem1 = DataElement(dest_kw_or_tg,
-                    dictionary_VR(dest_kw_or_tg), newelem.value)
+                    dictionary_VR(dest_kw_or_tg), new_elem.value)
                 dest_ds[dest_kw_or_tg] = new_elem1
             # now mark the attrib as used/done to keep track of every one of it
             self._mark_tag_as_used(src_kw_or_tg)
@@ -732,7 +731,9 @@ class Abstract_MultiframeModuleAdder(ABC):
 
         self._mark_tag_as_used(tg)
         return a
-    def _add_module(self, module_name: str, check_not_to_be_perframe=True):
+    def _add_module(self, module_name: str, excepted_attributes = []
+        , check_not_to_be_perframe=True
+        , check_not_to_be_empty=False):
         # sf_sop_instance_uid = sf_datasets[0]
         # mf_sop_instance_uid = LEGACY_ENHANCED_SOP_CLASS_UID_MAP[sf_sop_instance_uid]
         # iod_name = _SOP_CLASS_UID_IOD_KEY_MAP[mf_sop_instance_uid]
@@ -741,14 +742,17 @@ class Abstract_MultiframeModuleAdder(ABC):
         attribs = MODULE_ATTRIBUTE_MAP[module_name]
         ref_dataset = self.SingleFrameSet[0]
         for a in attribs:
+            if a in excepted_attributes:
+                continue
             if len(a['path']) == 0:
-                self._copy_attrib_if_present(ref_dataset, self.TargetDataset, a['keyword'],
-                                        check_not_to_be_perframe=check_not_to_be_perframe)
+                self._copy_attrib_if_present(ref_dataset, self.TargetDataset, a['keyword']
+                                        , check_not_to_be_perframe=check_not_to_be_perframe
+                                        , check_not_to_be_empty=check_not_to_be_empty)
 
     @abstractmethod
     def AddModule(self):
         pass
-class PixelImageModule(Abstract_MultiframeModuleAdder):
+class ImagePixelModule(Abstract_MultiframeModuleAdder):
     def __init__(self, sf_datasets:Sequence[Dataset]
     , excluded_from_perframe_tags:Sequence[Tag]
     , excluded_from_functional_tags:Sequence[Tag]
@@ -764,7 +768,107 @@ class PixelImageModule(Abstract_MultiframeModuleAdder):
             , multi_frame_output
         )
     def AddModule(self):
-        self._add_module('image-pixel', False) # don't check the perframe set
+        self._add_module('image-pixel',excepted_attributes=[], check_not_to_be_empty=False,
+        check_not_to_be_perframe=False) # don't check the perframe set
+class CompositeInstanceContex(Abstract_MultiframeModuleAdder):
+    def __init__(self, sf_datasets:Sequence[Dataset]
+    , excluded_from_perframe_tags:Sequence[Tag]
+    , excluded_from_functional_tags:Sequence[Tag]
+    , perframe_tags: Sequence[Tag]
+    , shared_tags: Sequence[Tag]
+    , multi_frame_output:Dataset=None):
+        super().__init__(
+            sf_datasets
+            , excluded_from_perframe_tags
+            , excluded_from_functional_tags
+            , perframe_tags
+            , shared_tags
+            , multi_frame_output
+        )
+        self._module_excepted_list = {
+         "patient" : []
+         , "clinical-trial-subject" : []
+         , "general-study" : 
+                [ 
+                        "RequestingService"
+                ]
+         , "patient-study" : 
+                [ 
+                        "ReasonForVisit"
+                        , "ReasonForVisitCodeSequence"
+                ]
+         , "clinical-trial-study" : []
+         , "general-series" : 
+                [ 
+                        "SmallestPixelValueInSeries"
+                        , "LargestPixelValueInSeries"
+                        , "PerformedProcedureStepEndDate"
+                        , "PerformedProcedureStepEndTime"
+                ]
+         , "clinical-trial-series" : []
+         , "general-equipment" : 
+                [ 
+                        "InstitutionalDepartmentTypeCodeSequence"
+                ]
+         , "frame-of-reference" : []
+         , "sop-common" : 
+                [ 
+                        "SpecificCharacterSet"
+                        , "EncryptedAttributesSequence"
+                        , "MACParametersSequence"
+                        , "DigitalSignaturesSequence"
+                ]
+         , "general-image" : 
+                [ 
+                        "ImageType"
+                        , "AcquisitionDate"
+                        , "AcquisitionDateTime"
+                        , "AcquisitionTime"
+                        , "AnatomicRegionSequence"
+                        , "PrimaryAnatomicStructureSequence"
+                        , "IrradiationEventUID"
+                        , "AcquisitionNumber"
+                        , "InstanceNumber"
+                        , "PatientOrientation"
+                        , "ImageLaterality"
+                        , "ImagesInAcquisition"
+                        , "ImageComments"
+                        , "QualityControlImage"
+                        , "BurnedInAnnotation"
+                        , "RecognizableVisualFeatures"
+                        , "LossyImageCompression"
+                        , "LossyImageCompressionRatio"
+                        , "LossyImageCompressionMethod"
+                        , "RealWorldValueMappingSequence"
+                        , "IconImageSequence"
+                        , "PresentationLUTShape"
+                ]
+         , "sr-document-general" : 
+                [ 
+                        "ContentDate"
+                        , "ContentTime"
+                        , "ReferencedInstanceSequence"
+                        , "InstanceNumber"
+                        , "VerifyingObserverSequence"
+                        , "AuthorObserverSequence"
+                        , "ParticipantSequence"
+                        , "CustodialOrganizationSequence"
+                        , "PredecessorDocumentsSequence"
+                        , "CurrentRequestedProcedureEvidenceSequence"
+                        , "PertinentOtherEvidenceSequence"
+                        , "CompletionFlag"
+                        , "CompletionFlagDescription"
+                        , "VerificationFlag"
+                        , "PreliminaryFlag"
+                        , "IdenticalDocumentsSequence"
+                ]
+}
+    def AddModule(self):
+        for module_name, excpeted_a in self._module_excepted_list.items():
+            self._add_module(module_name
+            , excepted_attributes = excpeted_a
+            , check_not_to_be_empty=False
+            , check_not_to_be_perframe=False) # don't check the perframe set
 class CommonCTMRPETImageDescriptionMacro(Abstract_MultiframeModuleAdder):
     def __init__(self, sf_datasets:Sequence[Dataset]
         , excluded_from_perframe_tags:Sequence[Tag]
@@ -783,7 +887,7 @@ class CommonCTMRPETImageDescriptionMacro(Abstract_MultiframeModuleAdder):
         )
         self.Modality = modality
     def _get_value_for_frame_type(self, attrib:DataElement):
-        if type(attrib) == DataElement:
+        if type(attrib) != DataElement:
             return None
         output = ['', '', '', '']
         v = attrib.value
@@ -793,41 +897,53 @@ class CommonCTMRPETImageDescriptionMacro(Abstract_MultiframeModuleAdder):
         output[2] = 'VOLUME' if l<3 else v[2]
         output[3] = 'NONE'
         return output
-
-    def _add_module_to_functional_group(self, src_fg:Dataset, dest_fg:Dataset):
+    def _get_frame_type_seq_tag(self)->int:
         seq_kw = '{}{}FrameTypeSequence'
         if self.Modality == 'PET':
             seq_kw = seq_kw.format(self.Modality, '')
         else:
             seq_kw = seq_kw.format(self.Modality, 'Image')
-        seq_tg = tag_for_keyword(seq_kw)
+        return tag_for_keyword(seq_kw)
+
+
+    def _add_module_to_functional_group(self, src_fg:Dataset, dest_fg:Dataset, level):
+        
         
         FrameType_a = src_fg['ImageType']
+        if level ==0:
+            FrameType_tg = tag_for_keyword('ImageType')
+        else:
+            FrameType_tg = tag_for_keyword('FrameType')
+
         new_val = self._get_value_for_frame_type(FrameType_a)
-        inner_item = Dataset()
-        FrameType_tg = tag_for_keyword('FrameType')
-        inner_item[FrameType_tg] = DataElement(FrameType_tg,
+        dest_fg[FrameType_tg] = DataElement(FrameType_tg,
             FrameType_a.VR, new_val)
-        dest_fg[seq_tg] = DataElement(seq_tg, dictionary_VR(seq_tg),
-            [inner_item]) 
+        element_generator = lambda kw, val:DataElement(tag_for_keyword(kw)
+            , dictionary_VR(tag_for_keyword(kw)), val)
+        dest_fg['PixelPresentation'] = element_generator('PixelPresentation', "MONOCHROME")
+        dest_fg['VolumetricProperties'] = element_generator('VolumetricProperties', "VOLUME")
+        dest_fg['VolumeBasedCalculationTechnique'] = element_generator('VolumeBasedCalculationTechnique', "NONE")
 
     def AddModule(self):
         im_type_tag = tag_for_keyword('ImageType')
         fm_type_tag = tag_for_keyword('FrameType')
+        
+        seq_tg=self._get_frame_type_seq_tag()
 
         if not im_type_tag in  self.PerFrameTags:
-            im_type_a = self.SingleFrameSet[0][im_type_tag]
-            new_val = self._get_value_for_frame_type(im_type_a)
-            self.TargetDataset[im_type_tag] = DataElement(im_type_tag,
-                im_type_a.VR, new_val)
+            self._add_module_to_functional_group(self.SingleFrameSet[0],self.TargetDataset,0) 
             #----------------------------
-            
             item = self._get_shared_item()
-            self._add_module_to_functional_group(self.SingleFrameSet[0],item) 
+            inner_item = Dataset()
+            self._add_module_to_functional_group(self.SingleFrameSet[0],inner_item,1) 
+            item[seq_tg] = DataElement(seq_tg, 'SQ', [inner_item])
         else:
             for i in range(0, len(self.SingleFrameSet)):
                 item = self._get_perframe_item(i)
-                self._add_module_to_functional_group(self.SingleFrameSet[i],item) 
+                inner_item = Dataset()
+                self._add_module_to_functional_group(self.SingleFrameSet[i],inner_item, 1) 
+                item[seq_tg] = DataElement(seq_tg, 'SQ', [inner_item])
+
                 
 class EnhancedCommonImageModule(Abstract_MultiframeModuleAdder):
     def __init__(self, sf_datasets:Sequence[Dataset]
@@ -928,6 +1044,82 @@ class EnhancedCTImageModule(Abstract_MultiframeModuleAdder):
     def AddModule(self):
         pass
         #David's code doesn't hold anything for this module ... should ask him
+class EnhancedPETImageModule(Abstract_MultiframeModuleAdder):
+    def __init__(self, sf_datasets:Sequence[Dataset]
+    , excluded_from_perframe_tags:Sequence[Tag]
+    , excluded_from_functional_tags:Sequence[Tag]
+    , perframe_tags: Sequence[Tag]
+    , shared_tags: Sequence[Tag]
+    , multi_frame_output:Dataset=None):
+        super().__init__(
+            sf_datasets
+            , excluded_from_perframe_tags
+            , excluded_from_functional_tags
+            , perframe_tags
+            , shared_tags
+            , multi_frame_output
+        )
+    def AddModule(self):
+        pass
+        #David's code doesn't hold anything for this module ... should ask him
+class EnhancedMRImageModule(Abstract_MultiframeModuleAdder):
+    def __init__(self, sf_datasets:Sequence[Dataset]
+    , excluded_from_perframe_tags:Sequence[Tag]
+    , excluded_from_functional_tags:Sequence[Tag]
+    , perframe_tags: Sequence[Tag]
+    , shared_tags: Sequence[Tag]
+    , multi_frame_output:Dataset=None):
+        super().__init__(
+            sf_datasets
+            , excluded_from_perframe_tags
+            , excluded_from_functional_tags
+            , perframe_tags
+            , shared_tags
+            , multi_frame_output
+        )
+  
+
+    def AddModule(self):
+        self._copy_attrib_if_present(self.SingleFrameSet[0]
+            , self.TargetDataset
+            , "ResonantNucleus"
+            , check_not_to_be_perframe = True 
+            , check_not_to_be_empty = True)
+            
+        if not 'ResonantNucleus' in self.TargetDataset:
+			# derive from ImagedNucleus, which is the one used in legacy MR IOD, 
+            # but does not have a standard list of defined terms ... (could check these :()
+            self._copy_attrib_if_present(self.SingleFrameSet[0]
+                , self.TargetDataset
+                , "ImagedNucleus"
+                , check_not_to_be_perframe = True 
+                , check_not_to_be_empty = True)
+
+        self._copy_attrib_if_present(self.SingleFrameSet[0]
+            , self.TargetDataset
+            , "KSpaceFiltering"
+            , check_not_to_be_perframe = True 
+            , check_not_to_be_empty = True)
+            
+        self._copy_attrib_if_present(self.SingleFrameSet[0]
+            , self.TargetDataset
+            , "MagneticFieldStrength"
+            , check_not_to_be_perframe = True 
+            , check_not_to_be_empty = True)
+            
+        self._copy_attrib_if_present(self.SingleFrameSet[0]
+            , self.TargetDataset
+            , "ApplicableSafetyStandardAgency"
+            , check_not_to_be_perframe = True 
+            , check_not_to_be_empty = True)
+            
+        self._copy_attrib_if_present(self.SingleFrameSet[0]
+            , self.TargetDataset
+            , "ApplicableSafetyStandardDescription"
+            , check_not_to_be_perframe = True 
+            , check_not_to_be_empty = True)
+            
+   
 class AcquisitionContextModule(Abstract_MultiframeModuleAdder):
     def __init__(self, sf_datasets:Sequence[Dataset]
     , excluded_from_perframe_tags:Sequence[Tag]
@@ -968,20 +1160,28 @@ class FrameAnatomyFunctionalGroup(Abstract_MultiframeModuleAdder):
         #Should check it out later.
         fa_seq_tg = tag_for_keyword('FrameAnatomySequence')
         item = Dataset()
-        self._copy_attrib_if_present(item, src_fg, 'AnatomicRegionSequence'
-                                    , check_not_to_be_perframe=False)
-        self._copy_attrib_if_present(item, src_fg, 'FrameLaterality'
+        self._copy_attrib_if_present(src_fg, item, 'AnatomicRegionSequence'
+                                    , check_not_to_be_perframe=False
+                                    , check_not_to_be_empty=False)
+        self._copy_attrib_if_present(src_fg, item, 'FrameLaterality'
                                     , check_not_to_be_perframe=False
                                     , check_not_to_be_empty=True)
         
         if not 'FrameLaterality' in item:
-            self._copy_attrib_if_present(item, src_fg, 'ImageLaterality'
+            self._copy_attrib_if_present(src_fg, item, 'ImageLaterality'
                                     , 'FrameLaterality'
-                                    , check_not_to_be_perframe=False)
+                                    , check_not_to_be_perframe=False
+                                    , check_not_to_be_empty=True)
         if not 'FrameLaterality' in item:
-            self._copy_attrib_if_present(item, src_fg, 'Laterality'
+            self._copy_attrib_if_present(src_fg, item, 'Laterality'
                                     , 'FrameLaterality'
-                                    , check_not_to_be_perframe=False)
+                                    , check_not_to_be_perframe=False
+                                    , check_not_to_be_empty=True)
+        if not 'FrameLaterality' in item:
+            FrameLaterality_a=self._get_or_create_attribute(src_fg, 'FrameLaterality', "U")
+            item['FrameLaterality'] = FrameLaterality_a
+            
+        
 
         FrameAnatomy_a = DataElement(fa_seq_tg, dictionary_VR(fa_seq_tg),
                                         [item] )
@@ -1113,51 +1313,7 @@ class PlanePositionFunctionalGroup(Abstract_MultiframeModuleAdder):
                 item = self._get_perframe_item(i)
                 self._add_module_to_functional_group(self.SingleFrameSet[i],item)
                     
-class PlanePositionFunctionalGroup(Abstract_MultiframeModuleAdder):
-    def __init__(self, sf_datasets:Sequence[Dataset]
-    , excluded_from_perframe_tags:Sequence[Tag]
-    , excluded_from_functional_tags:Sequence[Tag]
-    , perframe_tags: Sequence[Tag]
-    , shared_tags: Sequence[Tag]
-    , multi_frame_output:Dataset=None):
-        super().__init__(
-            sf_datasets
-            , excluded_from_perframe_tags
-            , excluded_from_functional_tags
-            , perframe_tags
-            , shared_tags
-            , multi_frame_output
-        )
-    def _contains_right_attributes(self, tags:dict) ->bool:
-        ImagePositionPatient_tg = tag_for_keyword('ImagePositionPatient' )
-
-        return ImagePositionPatient_tg in tags
-    def _add_module_to_functional_group(self, src_fg:Dataset, dest_fg:Dataset):
-        item = Dataset()
-        self._copy_attrib_if_present(src_fg
-                                    , item
-                                    ,'ImagePositionPatient'
-                                    , check_not_to_be_perframe=False
-                                    , check_not_to_be_empty=False)
-        PlanePositionSequence_kw = 'PlanePositionSequence'
-        PlanePositionSequence_tg = tag_for_keyword(PlanePositionSequence_kw)
-        seq = DataElement(PlanePositionSequence_tg
-                        , dictionary_VR(PlanePositionSequence_tg)
-                        , [item])
-        dest_fg[PlanePositionSequence_tg] = seq
-
-    def AddModule(self):
-        if (not self._contains_right_attributes(self.PerFrameTags)
-            and (self._contains_right_attributes(self.SharedTags)
-            or self._contains_right_attributes(self.ExcludedFromPerFrameTags))
-            ):
-            item = self._get_shared_item()
-            self._add_module_to_functional_group(self.SingleFrameSet[0],item)
-        elif self._contains_right_attributes(self.PerFrameTags):
-            for i in range(0, len(self.SingleFrameSet)):
-                item = self._get_perframe_item(i)
-                self._add_module_to_functional_group(self.SingleFrameSet[i],item)
-                    
+                  
 class PlaneOrientationFunctionalGroup(Abstract_MultiframeModuleAdder):
     def __init__(self, sf_datasets:Sequence[Dataset]
     , excluded_from_perframe_tags:Sequence[Tag]
@@ -1517,11 +1673,13 @@ class ConversionSourceFunctionalGroup(Abstract_MultiframeModuleAdder):
         item = Dataset()
         self._copy_attrib_if_present(src_fg
                                     , item
+                                    ,'SOPClassUID'
                                     ,'ReferencedSOPClassUID'
                                     , check_not_to_be_perframe=False
                                     , check_not_to_be_empty=True)
         self._copy_attrib_if_present(src_fg
                                     , item
+                                    ,'SOPInstanceUID'
                                     ,'ReferencedSOPInstanceUID'
                                     , check_not_to_be_perframe=False
                                     , check_not_to_be_empty=True)
@@ -1554,7 +1712,7 @@ class FrameContentFunctionalGroup(Abstract_MultiframeModuleAdder):
             , shared_tags
             , multi_frame_output
         )
-        self.EarliestFrameAcquisitionDateTime = DT('99991231235959')
+        self.EarliestFrameAcquisitionDateTime = self.FarthestFutureDateTime
     def _contains_right_attributes(self, tags:dict) ->bool:
         AcquisitionDateTime_tg = tag_for_keyword('AcquisitionDateTime')
         AcquisitionDate_tg = tag_for_keyword('AcquisitionDate')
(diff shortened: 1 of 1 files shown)


hackermd:
Thanks @afshinmessiah! I had a quick look and will review more carefully before we meet.

In the meantime, as a first step I would suggest you look into the [coding style section of the developer guide](https://highdicom.readthedocs.io/en/latest/development.html#coding-style) and make sure the code passes flake8 and mypy checks.

fedorov:
@hackermd the PR is not complete, and is known not to pass the checks. I suggested Afshin shares this incomplete PR with you so you can (if you like) look at the overall approach before the meeting.

hackermd:
@afshinmessiah please fix typing errors locally before pushing to avoid test pipeline failures such as this one
https://travis-ci.com/github/MGHComputationalPathology/highdicom/jobs/430572557

afshinmessiah:
> @afshinmessiah please fix typing errors locally before pushing to avoid test pipeline failures such as this one
> https://travis-ci.com/github/MGHComputationalPathology/highdicom/jobs/430572557

Sorry, these days I'm doing so many tests/debug to covert single frame data collections and push changes in a rush. For now, I corrected the typing errors.

src/highdicom/legacy/sop.py
```diff
@@ -621,21 +530,24 @@ def __init__(
...
-            **kwargs
-        )
+            **kwargs)
         _convert_legacy_to_enhanced(legacy_datasets, self)
+
+
 from abc import ABC, abstractmethod
+
+
 class Abstract_MultiframeModuleAdder(ABC):
```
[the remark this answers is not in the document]
afshinmessiah: @hackermd, I restructured the codes to apply your recommendations. Now there are a couple of key classes I'm going to use in order to convert the single frame. 

- There are classes `FrameSet, FrameSetCollection` responsible for detecting all instances that belong to a multi-frame image and then identify the shared/perframe attributes. 
- The class `LegacyConvertedEnhanceImage` inheriting `SOPClass` and is responsible for adding all necessary modules to multi-frame image. 
- There are `PerframeFunctionalGroup/SharedFunctionalGroup` classes inheriting `DataElementSequence` class which initialize sequences for multiframe images getting populated along the way.

hackermd (changes requested):
Thank you @afshinmessiah! I did not yet review the internal logic of the actual conversion, but would like to focus first on coding style (see also the [coding style section of the developer guide](https://highdicom.readthedocs.io/en/latest/development.html#coding-style) and general design considerations:

1. Please follow PEP8 and the patterns used in this package (lower_case_with_underscores instead of UpperCamelCase, etc.)
2. Add docstrings to all methods, including "private" methods
3. Avoid changing the public API and thereby breaking backwards compatibility (see comments on `FrameSet` and `LegacyConvertedEnhanceImage`)
4. Don't add any "public" attributes to instances of classes derived from `SOPClass`. This shall be reserved for keywords defined in the DICOM dictionary.

src/highdicom/legacy/sop.py
```diff
@@ -23,533 +24,385 @@
...
+    @property
+    def DistinguishingAttributeKeywords(self) -> List[str]:
+        return self._DistinguishingAttributeKeywords[:]
+
+    @property
+    def FrameSets(self) -> List[FrameSet]:
+        return self._FrameSets
+
+
+class LegacyConvertedEnhanceImage(SOPClass):
```
hackermd, line 394: Classes derived from `SOPClass` should actually map to a DICOM SOP Class (and corresponding IOD). There is no DICOM *Legacy Converted Enhance Image* IOD!

This module previously defined three classes: `LegacyConvertedEnhancedMRImage`, `LegacyConvertedEnhancedCTImage` and `LegacyConvertedEnhancedPETImage`. We need those three separate classes.

src/highdicom/legacy/sop.py
```diff
@@ -1,20 +1,21 @@
...
 from highdicom.base import SOPClass
-from highdicom._iods import IOD_MODULE_MAP, SOP_CLASS_UID_IOD_KEY_MAP
+from highdicom._iods import IOD_MODULE_MAP
 from highdicom._modules import MODULE_ATTRIBUTE_MAP
-
-
-logger = logging.getLogger(__name__)
-
-
+# logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
afshinmessiah: There is a problem with declaring logger globally. Usually, the logger is configured by the end-user/progrmmer after importing the modules.  This scenario doesn't work with the global logger because the logger is instantiated before configuration. For instance in user code when they import highdicom (the global logger is instantiated) then configures the logger which would not affect the already instantiated global logger. I tested this myself and the logger failed to log for highdicom.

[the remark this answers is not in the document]
fedorov: Afshin, I do not understand your concern. The way I understand it, there is a separate logger for specific modules, in this case this will be a logger for `highdicom.sr.sop`. The user can access that same logger by name after importing the module. I don't think user application using highdicom would want to log into the same logger defined in the highdicom modules, but the user could, if needed.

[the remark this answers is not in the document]
afshinmessiah: To my understanding, the whole point of using loggers is giving the control of how to log (configuration) to the end-user. Otherwise printing on std output would suffice. The configuration says root logger how(style) and where(file/std/network) to log which is not done by the app developer but by the end-user. Usually, I configure the root logger once and get the logger in each function with the module name. If a specific configuration for the module name exists it is configured in that way otherwise it will be configured as root. 
With highgdicom module logger, when I want to configure the logger, I need to access the instantiated global logger separately and configure it all over again which imposes extra coding and a dirtied control flow. Instead, each function could get the once configured logger while it is being called. I'm also curious what is the benefit of having a global logger in one module over getting a logger with the module name separately in function while it complies by the most recent setting?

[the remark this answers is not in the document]
fedorov: Let's hear from @hackermd, but we cannot revert the decisions made already in the library without first consulting with Markus. If that's how he would like logging to be done, then that's how it should be done. I suggest that changes to logging, if needed, should be revisited later in a separate PR.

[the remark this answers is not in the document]
afshinmessiah: It was just a suggestion. It started when I was not able to get logging from highdicom and change the code, I thought it wouldn't change the coding structure too much. Any way if @hackermd insists on keeping this way, I'll comply.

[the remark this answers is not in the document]
hackermd: The convention in Python is to instantiate a logger instance per module in the library. Logging is then configured only once at the main entry point of the application (e.g., command line program). See [Python logging tutorial](https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial).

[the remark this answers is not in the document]
afshinmessiah: Well, there is just one logger for each module even if I instantiate the logger within the function because I use the module name even in the function. Anyway, when I configure the logger and highdicom's logger is global, I don't get any log from highdicom at all.

src/highdicom/legacy/sop.py
```diff
@@ -23,533 +24,385 @@
...
+        return self._FrameSets
+
+
+class LegacyConvertedEnhanceImage(SOPClass):
     """SOP class for Legacy Converted Enhanced PET Image instances."""
 
     def __init__(
             self,
-            legacy_datasets: Sequence[Dataset],
+            frame_set: FrameSet,
```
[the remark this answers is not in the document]
afshinmessiah: I understand your concern but based on use cases I ran into while converting tcia data collection, there are series with more than one multi-frame image. In order to have control over conversion, we need to separate them prior to conversion. This is the main reason that I used frameset collection externally. Calculating framesets internally will impose extra process overhead because in either case, I need to check the input dicoms externally to make sure I'm not feeding the SOPClass mixed frameset. Anyway, I'll modify the code as you prefer.

src/highdicom/legacy/sop.py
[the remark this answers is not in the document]
hackermd: How do you configure logging?

I generally do something along those lines:

```python
import logging
import sys

if __name__ == '__main__':
    logger = logging.getLogger('highdicom')
    stderr_handler = logging.StreamHandler(stream=sys.stderr)
    logger.addHandler(stderr_handler)
    logger.setLevel(logging.DEBUG)
```

See also [configuring logging for a library](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library).

Note that we use logging sparsely in *highdicom*, so you will generally not get a lot of log messages.

src/highdicom/legacy/sop.py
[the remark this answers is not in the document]
fedorov: @afshinmessiah do you have SOPSeriesInstanceUID for a sample series that has multiple multi-frame instances included?

src/highdicom/legacy/sop.py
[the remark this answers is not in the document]
afshinmessiah: I configure logging by a .json file. I configure each module separately, I just configure the root and every module will instantiate the root logger. I also re-implemented the logging handler to work on multiprocessing fashion. This policy works for my code and a couple of google cloud api but doesn't work for highdicom unless I instantiate logger in the function. No worries, I will find a workaround for this problem of mine and apply your suggestion.

src/highdicom/legacy/sop.py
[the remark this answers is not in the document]
afshinmessiah: This series should have two multi-frames.
`SeriesInstanceUID = 1.3.6.1.4.1.14519.5.2.1.2783.4001.144515889821738216325254106222`
I must add that I tried to exclude them in the query, but it doesn't have anything to do with the code. Code should check how many instances there are and generate a separate UID for each.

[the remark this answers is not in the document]
fedorov: From what I see, all instances in that series have SOPClassUID of CTImageStorage, and I do not see any instance that has 2 frames (none has the "NumberOfFrames" attribute). Are you sure?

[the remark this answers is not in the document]
afshinmessiah: From mixed series I didn't mean them having mixed SOPClassUID, I meant they have different distinguishing attributes which leads to different framsets.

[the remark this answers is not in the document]
fedorov: Thanks for the clarification. I misunderstood the initial comment.

src/highdicom/legacy/sop.py
```diff
@@ -567,44 +420,1442 @@ def __init__(
...
+            "PreliminaryFlag",
+            "IdenticalDocumentsSequence"
+         ]
+        }
+        self.EarliestDate = DA('00010101')
+        self.EarliestTime = TM('000000')
+        self.EarliestDateTime = DT('00010101000000')
+        self.FarthestFutureDate = DA('99991231')
+        self.FarthestFutureTime = TM('235959')
+        self.FarthestFutureDateTime = DT('99991231235959')
```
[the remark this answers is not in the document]
afshinmessiah: No, they are not. I will change their camel case appearance.

tests/test_legacy.py
```diff
@@ -283,3 +141,500 @@ def generate_common_dicom_dataset_series(self, slice_count: int,
...
+#             tmp_dataset.SpecificCharacterSet = 'ISO_IR 100'
+#             tmp_dataset.StudyDate = date_
+#             tmp_dataset.StudyDescription = 'test study'
+#             tmp_dataset.StudyID = ''
+#             if (system == Modality.CT):
+#                 tmp_dataset.RescaleIntercept = 0
+#                 tmp_dataset.RescaleSlope = 1
+#             tmp_dataset.StudyTime = time_
+#             output_dataset.append(tmp_dataset)
+#         return output_dataset
```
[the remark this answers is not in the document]
afshinmessiah: Dicom generation part of the tests now is a separate class. I also removed the mixed SeriesInstanceUID test because based on David Clunie's opinion multi-frame images can span through multiple series.  We also don't consider different transfer syntax as a distinguishing attribute, so I removed the mixed transfer syntax test too. @fedorov, please check out my comment to discuss this issue further.

afshinmessiah:
@hackermd, I applied all modifications you wanted. Please take a look and let me know if anything left.

src/highdicom/legacy/sop.py
```diff
@@ -1,9 +1,7 @@
 """Module for SOP Classes of Legacy Converted Enhanced Image IODs."""
-
 import logging
 from collections import defaultdict
 from typing import Any, Dict, List, Optional, Sequence, Union
-
```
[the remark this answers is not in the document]
fedorov: @afshinmessiah built-in modules are still not separated from third-party by a newline.

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +31,506 @@
...
-        frame_type_item = Dataset()
-        frame_type_item.FrameType = frame_type
-        frame_type_item.PixelRepresentation = pixel_representation
-        frame_type_item.VolumetricProperties = volumetric_properties
-        if frame_type[0] == 'ORIGINAL':
-            frame_type_item.FrameVolumeBasedCalculationTechnique = 'NONE'
+    @classmethod
+    def isequal(cls, v1: Any, v2: Any) -> bool:
+        from pydicom.valuerep import DSfloat
+        float_tolerance = 1.0e-5
```
fedorov, line 62: I think it is better if tolerance is passed as a parameter to the constructor with the default.

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +31,506 @@
...
+
+        def is_equal_float(x1: float, x2: float) -> bool:
+            return abs(x1 - x2) < float_tolerance
+        if type(v1) != type(v2):
+            return False
+        if isinstance(v1, DataElementSequence):
+            for item1, item2 in zip(v1, v2):
+                cls.isequal_dicom_dataset(item1, item2)
```
fedorov, line 70: Is this correct - shouldn't the return value for this function be checked?

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +31,506 @@
...
-        # Cardiac Synchronization (U)
-        # TODO: http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.16.2.html#sect_C.7.6.16.2.7  # noqa
 
-        # Contrast/Bolus Usage (U) - MR/CT only
-        # TODO: http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.16.2.html#sect_C.7.6.16.2.12  # noqa
+class GeometryOfSlice:
 
-        # Respiratory Synchronization (U)
-        # TODO: http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.16.2.html#sect_C.7.6.16.2.17  # noqa
+    """A class for checking dicom slices geomtery/parallelization"""
```
fedorov, line 133: Typo - "geomtery". "Parallelization" should probably be "parallelity".

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +31,506 @@
...
+                 dimensions: tuple) -> None:
+        """
+        Parameters
+        ----------
+        row_vector: ndarray
+            3D vector representing row of the input slice
+        col_vector: ndarray
+            3D vector representing column the input  aslice
+        top_left_corner_pos: ndarray
+            3D point representing top left coner position of the input slice
```
fedorov, line 149: "Coner" typo.

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +31,506 @@
...
+                 top_left_corner_pos: ndarray,
+                 voxel_spaceing: ndarray,
+                 dimensions: tuple) -> None:
+        """
+        Parameters
+        ----------
+        row_vector: ndarray
+            3D vector representing row of the input slice
+        col_vector: ndarray
+            3D vector representing column the input  aslice
```
fedorov, line 147: Typo - "input aslice"

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +31,506 @@
...
+        top_left_corner_pos: ndarray
+            3D point representing top left coner position of the input slice
+        voxel_spaceing: ndarray
```
fedorov, line 150: "spaceing" typo

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +31,506 @@
...
-        for _ in range(len(sf_datasets))
-    ]
-    for tag, dataelements in unassigned_dataelements.items():
-        values = [str(da.value) for da in dataelements]
-        unique_values = set(values)
-        if len(unique_values) == 1:
-            unassigned_shared_ca_item.add(dataelements[0])
+        """
+        n: ndarray = cross(self.RowVector, self.ColVector)
+        n[2] = -n[2]
```
fedorov, line 168: Why is this negation here?

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +31,506 @@
...
+        voxel_spaceing: ndarray
+            Three element array. 1st and 2nd copied from PexelSpacing and the
+            3rd copied from SliceThickness
+        dimensions: tuple
+            3 element tuple holding x as number of cols, y as number of rows
+            and z as 1
+        """
+        self.RowVector = row_vector
+        self.ColVector = col_vector
+        self.TopLeftCornerPosition = top_left_corner_pos
```
fedorov, line 159: Maybe "origin" is a better name to be used here?

Also, as noted earlier, CamelCase is reserved for DICOM keywords.

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +31,506 @@
...
+        """
+        n = self.get_normal_vector()
+        return float(
+            dot(self.TopLeftCornerPosition, n))
+
+    def are_parallel(
+            slice1: Any,
+            slice2: Any,
+            tolerance: float = 0.0001) -> bool:
+        """Returns False if two slices are not prallel else True
```
fedorov, line 183: typo

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +31,506 @@
...
+        """Returns the shortest distince of the slice from the origin
+
+        """
+        n = self.get_normal_vector()
+        return float(
+            dot(self.TopLeftCornerPosition, n))
+
+    def are_parallel(
+            slice1: Any,
+            slice2: Any,
```
fedorov, line 181: Shouldn't the type annotation here be "GeometryOfSlice" instead of "Any"?

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +31,506 @@
...
+
+    def get_distance_along_origin(self) -> float:
+        """Returns the shortest distince of the slice from the origin
+
+        """
+        n = self.get_normal_vector()
+        return float(
+            dot(self.TopLeftCornerPosition, n))
+
+    def are_parallel(
```
fedorov, line 179: It looks like you are using the same approach to determining whether slices are parallel as in David Clunie Pixelmed Toolkit (I was wondering if he would check tolerance for the angle between normals instead, and that's why I checked). If this is indeed the case, it could be helpful to add a comment whenever certain approach is borrowed from David's code, just because it would be helpful for the reviewer of the code, and future user.

fedorov (changes requested):
I didn't finish the review, but I thought I submit the comments I have so far.

tests/test_legacy.py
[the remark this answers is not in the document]
fedorov: I agree SeriesInstanceUID should not be used as a distinguishing attribute. One most common example is dynamic contrast imaging, which some vendors encode as a separate series per time point, but which can be encoded as a single MF.

About transfer syntax, I am not so sure.

src/highdicom/legacy/sop.py
afshinmessiah: You are right, corrected.

src/highdicom/legacy/sop.py
afshinmessiah: Supposing row vector is along the y-axis and col vector is along the x-axis, it makes sense.

src/highdicom/legacy/sop.py
afshinmessiah: Sure, corrected. But the naming convention is copied from David's code.

src/highdicom/legacy/sop.py
afshinmessiah: It was already GeometryOfSlice but got an error from mypy saying that if the type is set as GeometryOfSlice why the function is doing extra type checking in "if clause" which never gets into.

src/highdicom/legacy/sop.py
afshinmessiah: Sure I will add comments pointing to the fact that the code logic is borrowed from David Clunie's PixelMed.

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +31,506 @@
...
+                    return False
+        return True
 
-        # Derivation Image (C)
-        try:
-            perframe_item.SourceImageSequence = ds.SourceImageSequence
-        except AttributeError:
-            pass
+    @classmethod
+    def isequal_dicom_dataset(cls, ds1: Dataset, ds2: Dataset) -> bool:
```
hackermd, line 89: I wonder to which extent this class and the individual methods are necessary. The *pydicom* library seems to support (at least a subset of) them, for example [pydicom.dataset.Dataset.__eq__](https://github.com/pydicom/pydicom/blob/master/pydicom/dataset.py#L694).

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +31,506 @@
...
-        conv_src_attr_item = Dataset()
-        conv_src_attr_item.ReferencedSOPClassUID = ds.SOPClassUID
-        conv_src_attr_item.ReferencedSOPInstanceUID = ds.SOPInstanceUID
-        perframe_item.ConversionSourceAttributesSequence = [
-            conv_src_attr_item,
-        ]
+        """
+        if not isinstance(tg, BaseTag):
+            tg = Tag(tg)
+        return '(0x{:0>4x}, 0x{:0>4x})'.format(tg.group, tg.element)
```
hackermd, line 120: See [pydicom.tag.BaseTag.__str__](https://github.com/pydicom/pydicom/blob/master/pydicom/tag.py#L193)

src/highdicom/legacy/sop.py
```diff
@@ -417,66 +545,1911 @@ def __init__(
...
+            accession_number=None if 'AccessionNumber' not in ref_ds
+            else ref_ds.AccessionNumber,
+            study_id=None if 'StudyID' not in ref_ds
+            else ref_ds.StudyID,
+            study_date=None if 'StudyDate' not in ref_ds
+            else ref_ds.StudyDate,
+            study_time=None if 'StudyTime' not in ref_ds
+            else ref_ds.StudyTime,
+            referring_physician_name=None if 'ReferringPhysicianName' not in
+            ref_ds else ref_ds.ReferringPhysicianName,
```
hackermd, line 590: These are either type 1 or type 2 attributes and should be present in `ref_ds`, shouldn't they?

src/highdicom/legacy/sop.py
```diff
@@ -417,66 +545,1911 @@ def __init__(
...
+        elif tg in self._perframe_tags:
+            self._perframe_tags[tg] = True
+
+    def _copy_attrib_if_present(self, src_ds: Dataset, dest_ds: Dataset,
+                                src_kw_or_tg: str, dest_kw_or_tg: str = None,
+                                check_not_to_be_perframe: bool = True,
+                                check_not_to_be_empty: bool = False) -> None:
+        """Copies a dicom attribute value from a keyword in the source Dataset
+        to a similar or different keyword in the destination Dataset
+        Parameters
```
hackermd, line 760: Couple of points regarding docstrings:

* Keep space between sections and at the end of the docstring
* Use the absolute namespace for documentation of custom types from external packages, e.g., `pydicom.dataset.Dataset` instead of `Dataset`
* Document type of return value (not relevant to this particular method)

src/highdicom/legacy/sop.py
```diff
@@ -417,66 +545,1911 @@ def __init__(
...
+                dest_ds[dest_kw_or_tg] = new_elem
+            else:
+                new_elem1 = DataElement(dest_kw_or_tg,
+                                        dictionary_VR(dest_kw_or_tg),
+                                        new_elem.value)
+                dest_ds[dest_kw_or_tg] = new_elem1
+            # now mark the attrib as used/done to keep track of every one of it
+            self._mark_tag_as_used(src_kw_or_tg)
+
+    def _get_or_create_attribute(
```
hackermd, line 804: I wonder whether this could be drastically simplified by relying on *pydicom* using `setattr()` instead of working with "raw" `pydicom.dataelem.DataElement` instances.

src/highdicom/legacy/sop.py
```diff
@@ -417,66 +545,1911 @@ def __init__(
...
+            try:
+                t_tmp = TM(a.value)
+                a.value = TM(default) if t_tmp is None else t_tmp
+            except BaseException:
+                a.value = TM(default)
+
+        self._mark_tag_as_used(tg)
+        return a
+
+    def _add_module(self, module_name: str, excepted_attributes: list = [],
```
hackermd, line 853: Please follow the style of the package and place each parameter on a separate line if the length of the function definition would exceed the character limit.

hackermd, line 853: Are you sure you want to use `list = []`? Mutable default arguments can cause unexpected behavior in Python and should generally be avoided in my opinion.

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +31,506 @@
...
+        """Construncts a shared frame functional group holding only one item
 
-    """SOP class for Legacy Converted Enhanced MR Image instances."""
+        """
+        super().__init__()
+        item = Dataset()
+        self.append(item)
+
+
+class FrameSet:
```
hackermd, line 233: Please make this class "private", i.e., rename to `_FrameSet`

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +31,506 @@
...
+
+    """
+        A class containing the dicom frames that hold equal distinguishing
+        attributes to detect all perframe and shared dicom attributes
+    """
+
+    def __init__(self, single_frame_list: list,
+                 distinguishing_tags: list) -> None:
```
hackermd, line 241: Use more specific type annotations, e.g., `List[str]`

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +31,506 @@
...
+    @property
+    def perframe_tags(self) -> List[Tag]:
+        return self._perframe_tags[:]
+
+    @property
+    def shared_tags(self) -> List[Tag]:
+        return self._shared_tags[:]
+
+    @property
+    def SeriesInstanceUID(self) -> UID:
```
hackermd, line 284: If the class is not derived from `pydicom.dataset.Dataset`, use the Python naming convention for attributes -> `series_instance_uid`

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +31,506 @@
...
+            del rough_shared[t]
+        for t, v in rough_shared.items():
+            self._shared_tags.append(t)
+            self._perframe_tags.remove(t)
+
+    def _istag_excluded_from_perframe(self, t: BaseTag) -> bool:
+        return t in self.excluded_from_perframe_tags
+
+
+class FrameSetCollection:
```
hackermd, line 360: Please make this class "private", i.e., rename to `_FrameSetCollection`

src/highdicom/legacy/sop.py
```diff
@@ -417,66 +545,1911 @@ def __init__(
...
+            else ref_ds.StudyDate,
+            study_time=None if 'StudyTime' not in ref_ds
+            else ref_ds.StudyTime,
+            referring_physician_name=None if 'ReferringPhysicianName' not in
+            ref_ds else ref_ds.ReferringPhysicianName,
         )
-        _convert_legacy_to_enhanced(legacy_datasets, self)
-        self.PresentationLUTShape = 'IDENTITY'
+        self._legacy_datasets = legacy_datasets
+        self._perframe_functional_group = ـPerframeFunctionalGroup(
```
hackermd, line 593: I would use the plural for `_perframe_functional_groups` and `_shared_functional_groups`, since these represent sequences of functional groups

src/highdicom/legacy/sop.py
```diff
@@ -417,66 +545,1911 @@ def __init__(
...
+            len(legacy_datasets))
+        tg = tag_for_keyword('PerFrameFunctionalGroupsSequence')
+        self[tg] = DataElement(tg, 'SQ', self._perframe_functional_group)
+        self._shared_functional_group = ـSharedFunctionalGroup()
+        tg = tag_for_keyword('SharedFunctionalGroupsSequence')
+        self[tg] = DataElement(tg, 'SQ', self._shared_functional_group)
+        self.distinguishing_attributes_tags = self._get_tag_used_dictionary(
+            frame_set.distinguishing_attributes_tags)
+        self.excluded_from_perframe_tags = self._get_tag_used_dictionary(
+            frame_set.excluded_from_perframe_tags)
```
hackermd, line 603: Please make these attributes "private". They are only used for construction of the object and we don't want to maintain them as part of the public API.

src/highdicom/legacy/sop.py
```diff
@@ -417,66 +545,1911 @@ def __init__(
...
+        self._perframe_tags = self._get_tag_used_dictionary(
+            frame_set.perframe_tags)
+        self._shared_tags = self._get_tag_used_dictionary(
+            frame_set.shared_tags)
+        self.ExcludedFromFunctionalGroupsTags = {
```
hackermd, line 608: Is this a DICOM attribute? If not, please use Python naming convention and make "private".

src/highdicom/legacy/sop.py
afshinmessiah: You are right. The main reason though for me is to have control over floating-point comparison considering a tolerance.

src/highdicom/legacy/sop.py
afshinmessiah: Sounds trivial, but I wanted to include '0x' for hex representation :). Don't get why they didn't include hex signe.

src/highdicom/legacy/sop.py
afshinmessiah: Yes, they should, but I don't want to stop conversion just because one of them missing in the input single-frame sets (while converting huge collections, there were cases that missed Modality or some other attributes causing problems to the whole conversion process)

src/highdicom/legacy/sop.py
afshinmessiah: Well, we want a deep copy of an attribute from a source if available and not empty, otherwise, create a new one with a default value. In some sense, I could use setattr, but given my intent to control the Date and Time and empty attributes I would wind up writing some code more or less similar.

tests/test_legacy.py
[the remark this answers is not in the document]
hackermd: I am not sure about transfer syntax either, but I think we could get into trouble if differently lossy compressed single-frame images should be combined into one multi-frame image.

Could we use Frame of Reference UID as distinguishing attribute?

src/highdicom/legacy/sop.py
afshinmessiah: I'm afraid we can't make FrameSet and FrameSetCollection private. We use both classes to detect potential multi-frame sets within every candidate collection before feeding them into LegacyConvertedEnhanced images.

src/highdicom/legacy/sop.py
afshinmessiah: Same as [previeous](https://github.com/MGHComputationalPathology/highdicom/pull/34#discussion_r613715803).

tests/test_legacy.py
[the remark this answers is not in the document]
afshinmessiah: We are using FrameOfReferenceUID as a distinguishing attribute. I think should discuss this with @dclunie. Here is the [PixelMode](https://github.com/afshinmessiah/PixelMed/blob/master/pixelmed/dicom/FrameSet.java#L52) code which is the source of our code's logic.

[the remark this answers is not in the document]
fedorov: You mean we should discuss with David why he is not using transfer syntax as a distinguishing attribute?

afshinmessiah:
Just finished the second round of code modification and tried to apply the required changes and corrections.

tests/test_legacy.py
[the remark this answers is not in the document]
afshinmessiah: Yes I do.

src/highdicom/legacy/sop.py
hackermd: The overall pattern throughout the library is that constructors of classes derived from `SOPClass` accept the source images as `Sequence[pydicom.dataset.Dataset`. That's also how the `LegacyConvertedEnhanced*Image` classes were implemented. It looks like you updated the code to remain backwards compatible with the public API. It should now be possible to make these classes private, shouldn't it?

src/highdicom/legacy/sop.py
hackermd: That may make sense from a conversion perspective, but we want all DICOM objects generated by *highdicom* to be fully compliant with the standard and we enforce compliance via the API. If a legacy image instance has compliance issues (in particular if a type 1 attribute is missing), these would need to be fixed beforehand. We can handle missing type 2 attributes and set the value to `None` if the attribute is missing in the source image. However, I would suggest rewriting this using the [getattr()](https://docs.python.org/3/library/functions.html#getattr) function for improved readability. For example

```python
study_id = getattr(ref_df, 'StudyID', None)
```

src/highdicom/legacy/sop.py
```diff
@@ -417,66 +574,2089 @@ def __init__(
...
+            self._excluded_from_perframe_tags[tg] = True
+        elif tg in self._perframe_tags:
+            self._perframe_tags[tg] = True
+
+    def _copy_attrib_if_present(
+            self,
+            src_ds: Dataset,
+            dest_ds: Dataset,
+            src_kw_or_tg: str,
+            dest_kw_or_tg: str = None,
```
hackermd, line 795:
```suggestion
            dest_kw_or_tg: Optional[str] = None,
```

src/highdicom/legacy/sop.py
```diff
@@ -417,66 +574,2089 @@ def __init__(
...
+        to a similar or different keyword in the destination Dataset
+        Parameters
+        ----------
+        src_ds: pydicom.dataset.Dataset
+            Source Dataset to copy the attribute from.
+        dest_ds: pydicom.dataset.Dataset
+            Destination Dataset to copy the attribute to.
+        src_kw_or_tg: str
+            The keyword from the source Dataset to copy its value.
+        dest_kw_or_tg: str = None
```
hackermd, line 809:
```suggestion
        dest_kw_or_tg: str, optional
```

hackermd: Please apply this to the other parameters and other docstrings as well

src/highdicom/legacy/sop.py
hackermd: Please apply this to the other function definitions as well

src/highdicom/legacy/sop.py
afshinmessiah: The common use case to set apart all frames that belong to a potential MF dicom instance before feeding the list into SOPClass to avoid any failure. If we don't do that the conversion will not proceed and will crash. Of course, it is not necessary to do this when it is guaranteed that the input single-frame list belongs to one MF instance, in our case though, it is not.

afshinmessiah:
Just pushed updates for the recent comments. All recent comments were applied except making the FrameSet and FrameSetCollection classes private which was not possible due to their external usage.

src/highdicom/legacy/sop.py
CPBridge: It would be clearer to simply reverse the order of the cross-product instead. I.e. `cross(self.col_vector, self.row_vector)`, which has the same effect

src/highdicom/legacy/sop.py
CPBridge: It is an unusual pattern to require a certain type parameter but then return a value correctly if the type constraint is not satisfied. This is why mypy is complaining. If you use `GeometryOfSlice` as the type annotation and then raise a `TypeError` if the inputs are the wrong type, this mypy error will go away and the behaviour of the method will be more intuitive. Type hinting using `Any` is highly misleading

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +36,2630 @@
...
+                logger.debug('\t\t{}/{})\t {}'.format(
+                    dicom_i, len(x[0]), dicom_ds['SOPInstanceUID']))
+        frames = ''
+        for i, f_count in enumerate(frame_counts, 1):
+            frames += '{: 2d}){:03d}\t'.format(i, f_count)
+        frames = '{: 2d} frameset(s) out of all {: 3d} instances:'.format(
+            len(frame_counts), len(self.mixed_frames)) + frames
+        logger.info(frames)
+        for kw in to_be_removed_from_distinguishing_attribs:
+            self.distinguishing_attribute_keywords.remove(kw)
```
CPBridge, line 473: I cannot follow all of the logic here but there is definitely something awry. Here are some things that seem wrong:
1. I cannot see a situation in which `to_be_removed_from_distinguishing_attribs` would contain anything, since it is initialised to an empty set at the start of the method and then never altered.
2. The line `self.distinguishing_attribute_keywords.remove(kw)` first calls the `distinguishing_attribute_keywords` property, which returns a *copy* of the `_distinguishing_attribute_keywords` attribute and then removes the tag from the copy, so the change will not persist.

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +36,2630 @@
...
+
+    def _istag_excluded_from_perframe(self, t: BaseTag) -> bool:
+        return t in self._excluded_from_perframe_tags
+
+
+class FrameSetCollection:
+
+    """A class to extract framesets based on distinguishing dicom attributes"""
+
+    def __init__(self, single_frame_list: Sequence[Any]) -> None:
```
CPBridge, line 389: Why not `Sequence[Dataset]`?

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +36,2630 @@
...
+        for kwkw in self.distinguishing_attribute_keywords:
+            self._excluded_from_perframe_tags[tag_for_keyword(kwkw)] = False
+        self._excluded_from_perframe_tags[
+            tag_for_keyword('AcquisitionDateTime')] = False
+        self._excluded_from_perframe_tags[
+            tag_for_keyword('AcquisitionDate')] = False
+        self._excluded_from_perframe_tags[
+            tag_for_keyword('AcquisitionTime')] = False
+        self.excluded_from_functional_groups_tags = {
+            tag_for_keyword('SpecificCharacterSet'): False}
```
CPBridge, line 484: Seems like this could be written much more succinctly:
```suggestion
        self._excluded_from_perframe_tags = {
            tag_for_keyword(kwkw): False for kwkw in self.distinguishing_attribute_keywords
        }
        special_exclusions = [
            'AcquisitionDateTime',
            'AcquisitionDate',
            'AcquisitionTime',
            'SpecificCharacterSet'
        ]
        for kwkw in special_exclusions:
            self._excluded_from_perframe_tags[tag_for_keyword(kwkw)] = False
```

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +36,2630 @@
...
+        dimensions: tuple
+            3 element tuple holding x as number of cols, y as number of rows
+            and z as 1
+        """
+
+        self.row_vector = row_vector
+        self.col_vector = col_vector
+        self.top_left_corner_position = top_left_corner_pos
+        self.voxel_spacing = voxel_spacing
+        self.Dim = dimensions
```
CPBridge, line 170:
```suggestion
        self.dim = dimensions
```
Capitalisation. But also this attribute seems to be unused so maybe we can just delete it? In that case the `dimensions` parameter to the `__init__` would also be unused.

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +36,2630 @@
...
+
+        """
+
+        if (not isinstance(slice1, GeometryOfSlice) or
+                not isinstance(slice2, GeometryOfSlice)):
+            logger.warning(
+                'slice1 and slice2 are not of the same '
+                'type: type(slice1) = {} and type(slice2) = {}'.format(
+                    type(slice1), type(slice2)))
+            return False
```
CPBridge, line 205: Summarising various comments about this method here in this suggestion. Mypy does not complain about this
```suggestion
    @staticmethod
    def are_parallel(
            slice1: GeometryOfSlice,
            slice2: GeometryOfSlice,
            tolerance: float = 0.0001,
        ) -> bool:
        """Returns False if two slices are not parallel else True."""
        if (not isinstance(slice1, GeometryOfSlice) or
                not isinstance(slice2, GeometryOfSlice)):
            raise TypeError("Inputs slices must be of type 'GeometryOfSlice'")
```

CPBridge (changes requested):
Hi @afshinmessiah , @hackermd and @fedorov have asked me to take a look at this PR also. It's clear that a huge amount of work has gone into this, thank you it's going to be a great resource! There are however still several things that need to be addressed.

I did not go through the logic of all the parts of the code yet but I'm posting this review anyway because I'm paranoid that my browser will lose all the comments I've built up if I don't post now...

There are a few general comments that apply to various pieces of the code:
- Docstrings: on the whole these are great but there are a few spacing issues. I have changed a large number of these with `suggestions` which you should just be able to accept. But in general docstrings should have a single line at the top, then optionally other comments on multiple lines, followed by sections (Parameters, Returns, Raises, Notes), which have blank lines above them to separate them. Then there should be a blank before the end, followed immediate by the code (no further blank line). For example:

```python
def add_1(x: int) -> int:
    """Add one to a number.
    
    The output is 1 greater than the input. Wow.

    Parameters
    ----------
    x: int
        A number

    Returns
    -------
    int
        Another number

    """
    return x + 1
``` 

- Type hints. There are a few places where you type hints with "container" types" such as dict or list. In these cases, there the type of the contained item needs to be specified. E.g. `List[int]` or `Dict[str, Dataset]` (note capital letter types are imported from the standard library `typing` module), we should avoid using the overly general `list` and `dict` as type hints.

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +36,2630 @@
...
+    def get_distance_along_origin(self) -> float:
+        """Returns the shortest distance of the slice from the origin
+
+        """
+
+        n = self.get_normal_vector()
+        return float(
+            dot(self.top_left_corner_position, n))
+
+    def are_parallel(
```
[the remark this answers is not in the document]
CPBridge: (see suggestion well below for my suggested rewriting of this)

src/highdicom/legacy/sop.py
afshinmessiah: Removed that method.

src/highdicom/legacy/sop.py
afshinmessiah: Well, sure that it's fine for the conversion to crash and it does. But, what should users do if they want to convert a mixed series? Don't they need a separate tool to figure out which single frames are potential multi-frame?? If they develop this tool on their own they need to duplicate the logic of the Frameset class.

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +36,2630 @@
...
+        rough_shared: dict = {}
+        sfs = self.frames
+        for ds in sfs:
+            for ttag, elem in ds.items():
+                if (not ttag.is_private and not
+                    ـDicomHelper.istag_file_meta_information_group(ttag) and not
+                        ـDicomHelper.istag_repeating_group(ttag) and not
+                        ـDicomHelper.istag_group_length(ttag) and not
+                        self._istag_excluded_from_perframe(ttag) and
+                        ttag != tag_for_keyword('PixelData')):
```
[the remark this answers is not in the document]
afshinmessiah: The code logic is borrowed from David Clunie's PixelMed and I barely changed it. To my understanding, the PixelData is always a per-frame attribute and we just wanted to get around loading and comparing the big chunk of data stored in PixelData. The excluded tags from per-frame are being handled differently than the normal shared/per-frame attributes.

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +36,2630 @@
...
+                    elem = ds[ttag]
```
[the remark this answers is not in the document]
afshinmessiah: The reason is to avoid getting RawDataElement. The items iterator sometimes gives RawDataElement while ds[ttag] doesn't.

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +36,2630 @@
...
+                for v_i in v:
+                    if not ـDicomHelper.isequal(v_i, v[0]):
+                        all_values_are_equal = False
+                        break
+                if not all_values_are_equal:
+                    to_be_removed_from_shared.append(ttag)
+        from pydicom.datadict import keyword_for_tag
+        for t, v in rough_shared.items():
+            if keyword_for_tag(t) != 'PatientSex':
+                continue
```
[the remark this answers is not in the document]
afshinmessiah: I might have put it in for some debugging purpose.

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +36,2630 @@
...
+                    elem = ds[ttag]
+                    if ttag not in self._perframe_tags:
+                        self._perframe_tags.append(ttag)
+                    if ttag in rough_shared:
+                        rough_shared[ttag].append(elem.value)
+                    else:
+                        rough_shared[ttag] = [elem.value]
+        to_be_removed_from_shared = []
+        for ttag, v in rough_shared.items():
+            v = rough_shared[ttag]
```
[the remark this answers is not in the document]
afshinmessiah: Same answer as [above](https://github.com/MGHComputationalPathology/highdicom/pull/34#discussion_r628733378).

src/highdicom/legacy/sop.py
```diff
@@ -25,384 +36,2630 @@
...
+            'ProtocolName',
+            'ImageOrientationPatient',
+            'PixelSpacing',
+            'SliceThickness',
+            'AcquisitionContextSequence']
+        to_be_removed_from_distinguishing_attribs: set = set()
+        self._frame_sets: List[FrameSet] = []
+        frame_counts = []
+        frameset_counter = 0
+        while len(self.mixed_frames_copy) != 0:
```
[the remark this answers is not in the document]
afshinmessiah: At each iteration _find_all_similar_to_first_datasets finds all frames identical to the first frame and removes them from mixed_frames_copy list. Each iteration at least removes one item.

src/highdicom/legacy/sop.py
afshinmessiah: Thank you Chris, you are totally right. The to_be_removed_from_distinguishing_attribs is redundant and belongs to the code's older logic. I removed it from the code.

src/highdicom/legacy/sop.py
afshinmessiah: Agreed.

afshinmessiah:
Thank you @CPBridge for all your wonderful comments. I did my best to apply all changes required by the comments and not to leave out any. Except few comments to which I provided some explanation, I've applied all comments. As I explained [before](https://github.com/MGHComputationalPathology/highdicom/pull/34#discussion_r613187466) @fedorov and I do not concur with @hackermd and @CPBridge on making the FrameSet and FrameSetCollection classes private. I also currently use those classes externally to prepare framesets prior to conversion of the TCIA collection data.

dclunie:
I don't think that "Swapped voctors [sic] in normal calculation" is correct - instead you might want to just remove the negation of Z, which was copied from pixelmed, which I think is wrong about this (and I will fix); this would reverse the sort order of transverse (perpendicular to Z) slices (not that there is any "right" sort order).

fedorov:
@hackermd @CPBridge the last day for @afshinmessiah in his current BWH position is effectively May 17 (he will then take the last 2 weeks off as vacation, before moving to his new job). Is there anything I can do to help with this PR?

CPBridge:
Hi @afshinmessiah ,

> Thank you @CPBridge for all your wonderful comments.

You are most welcome. Thank you for working through them. I hope you were able to just click "Commit suggestion" on most of them rather than reproduce by hand!

> As I explained before @fedorov and I do not concur with @hackermd and @CPBridge on making the FrameSet and FrameSetCollection classes private.

This appears to be a particular point of disagreement/confusion holding this all back, so I want to try and sort this out. My position, and I think this mostly reflects @hackermd's also, is that I am not opposed outright to making `FrameSet` and `FrameSetCollection` public if they serve a purpose to the end suer, but I am just confused by why you want to/need to given the rest of the API. If we consider the "main entry points" of the code in this module to be the constructors of the three `LegacyConvertedEnhancedXImage` classes then these constructors cannot accept arguments of type `FrameSet` or `FrameSetCollection` - they just pass in a plain python list of legacy instances. Therefore, what would a user of `highdicom` do with `FrameSet` or `FrameSetCollection` if these classes were made public? Maybe you could provide a code snippet of a toy example of what user code using these classes to construct a Legacy Converted Enhanced image would actually look like. That would really help me understand your proposal.

afshinmessiah:
>This appears to be a particular point of disagreement/confusion holding this all back, so I want to try and sort this out. My position, and I think this mostly reflects @hackermd's also, is that I am not opposed outright to making `FrameSet` and `FrameSetCollection` public if they serve a purpose to the end suer, but I am just confused by why you want to/need to given the rest of the API. If we consider the "main entry points" of the code in this module to be the constructors of the three `LegacyConvertedEnhancedXImage` classes then these constructors cannot accept arguments of type `FrameSet` or `FrameSetCollection` - they just pass in a plain python list of legacy instances. Therefore, what would a user of `highdicom` do with `FrameSet` or `FrameSetCollection` if these classes were made public? Maybe you could provide a code snippet of a toy example of what user code using these classes to construct a Legacy Converted Enhanced image would actually look like. That would really help me understand your proposal.

FrameSet class gives the list of matched Dicom datasets which will be fed into `LegacyConvertedEnhancedXImage`. I've attached an image of a simple conversion function (sorry I couldn't/didn't know how to/ copy the multiline code into this comment)

image

fedorov:
What if we make the FrameSet class publicly available to the user as a utility to identify legacy frames that belong to a single frame set. The user would then apply this class to select frame sets that would be passed to the converter class. Inside the converter class there could be another check, or it could just throw an exception when any assumption is violated. Afshin, I am sure we discussed this before, but I do not recall why this would not be an option. @CPBridge are you concerned about changing the API for the converter, or you are opposed to make the FrameSet class public in any form, even as a utility class?

fedorov:
> having `FrameSetCollection` and `FrameSet` classes both accessible to the user

@CPBridge and @hackermd would this be acceptable to you?

dclunie:
FYI, by way of background ...

I introduced the term "FrameSet" in the [IHE Basic Image Review (BIR) profile](http://www.ihe.net/uploadedFiles/Documents/Radiology/IHE_RAD_Suppl_BIR.pdf) in order to treat similarly (for display purposes) a bunch of related single frame image instances and a single multi-frame image instance, so that the arbitrary difference in encoding did not manifest itself to the user of the application, as well as to deal with some specific partitioning issues (like dual-echo MR series). Nobody suggested a better term at the time, or was too concerned about this.

DICOM doesn't actually formally define a frame, but from the definition in PS3.3 of [Multi-frame Image](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/chapter_3.html#para_97cbf9b3-31cd-4be3-b7be-4aca4c74b783), it can probably be taken to mean a "two-dimensional pixel plane".

I grant you that in this context the "Frame" does have accompanying metadata, and is not just the array of pixel data, which one could argue is misleading, but it never occurred to me it would be misconstrued in this highly specific context of use.

So I think @afshinmessiah is using the terms "Frame" and "FrameSet" in a reasonable manner, and certainly the same way as I re-used them in PixelMed for doing set of single frame to multi-frame conversions (inspired by the BIR profile), so he inherited this (possibly poor choice of) terminology from me.

And I meant them to be a "set" not a "list" because members are not intended to be duplicated, which I believe is the distinction between the two. Ideally the code would check that there are no duplicates, if the input set is actually provided as a list.

PS. For our purposes of conversion, we further narrow what constitutes a FrameSet to have more commonality that merely being part of the same Series and indeed need not constrain Instances to be from the same Series in the general case. The BIR constrains were simplified for a specific purpose that may not always align with our legacy conversion use cases.

hackermd:
Thanks @dclunie for the additional information and background.

Within the *highdicom* library we generally refer to a "frame" as an Item of an encapsulated Pixel Data element of an Image instance. This is in line with the usage with the term in PS 3.3 (e.g., [Per-Frame Functional Groups Sequence](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.16.html#sect_C.7.6.16.1.2)) and PS 3.5 (e.g., [Transfer Syntaxes For Encapsulation of Encoded Pixel Data](dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_A.4.html)). We sometimes also use it more broadly to refer the decoded pixel matrix (NumPy array), i.e., the pixels of an individual image slice/plane/tile. I would avoid the term here in this context when referring to single-frame image instances.

Since the order of single-frame image instances is significant in this context, I would further avoid the term "set", which could easily be misunderstood in Python given the built-in type [set](https://docs.python.org/3/library/stdtypes.html#set-types-set-frozenset).

Irrespective of the terminology, I would prefer if this logic (figuring out which frames to use, sorting them, removing duplicates, performing runtime checks, etc.) would be applied within the constructor of the `LegacyConvertedEnhancedXImage` classes rather than in a separate class. We can factor that logic out into separate utilities (preferably using different names as elaborated above) and provide them as part of the library (e.g., in `highdicom.legacy.utils`) so that they can be re-used for managing collections of single-frame image instances beyond the legacy conversion use case.

dclunie:
A "sorted set" is still a "set", not a "list".

Further, the order may need to be determined by sorting based on metadata (even if it is only by InstanceNumber within Series), not the the order in which the instances are supplied in the input, which may depend on how they are listed in the filesystem, which may vary between platforms.

hackermd:
> A "sorted set" is still a "set", not a "list".

Python doesn't have an ordered set and the interface of the built-in set differs significantly from that of a [sequence](https://docs.python.org/3/glossary.html#term-sequence), for example it cannot be indexed using the `[]` operator and contained elements must be hashable.
I am not saying the term "set" is wrong from an abstract type theory perspective, but potentially misleading to Python developers. Therefore, I would avoid the term "set" and `Union[Sequence[Dataset], FrameSet]`.

We could instead use "unique" to emphasize that items of the sequence are not duplicated (see also [numpy.unique](https://numpy.org/doc/stable/reference/generated/numpy.unique.html)).

> Further, the order may need to be determined by sorting based on metadata (even if it is only by InstanceNumber within Series), not the the order in which the instances are supplied in the input, which may depend on how they are listed in the filesystem, which may vary between platforms.

Python provides several mechanisms for [sorting](https://docs.python.org/3/howto/sorting.html) sequences (lists), for example using [sorted](https://docs.python.org/3/library/functions.html#sorted), which accepts a function as an argument to obtain an attribute for comparing items and determining their order. This could be readily applied to `Sequence[Dataset]`.

fedorov:
> I would still prefer the constructor to just accept `Sequence[pydicom.dataset.Dataset]` and avoid introducing a new type. 

@afshinmessiah, can you implement this change, making the facilitating classes available as utility classes?

> although they would need to be renamed, because "frame" has a specific meaning in DICOM and is very misleading in this context

Afshin was using those terms since he did not want to change the conventions used in Pixelmed. But I don't think this is worth an argument - can you suggest an alternative name that you like Markus, so Afshin can fix this?

hackermd:
> Afshin was using those terms since he did not want to change the conventions used in Pixelmed. But I don't think this is worth an argument - can you suggest an alternative name that you like Markus, so Afshin can fix this?

The main logic seems to be implemented in the `FrameSet._find_per_frame_and_shared_tags()` method, which updates the `FrameSet._shared_tags` and `FrameSet._perframe_tags` attributes. Could we  just create a utility function (in `highdicom.sr.utils`) instead with the following signature?

```python
def group_per_frame_and_shared_attributes(
    instances: Sequence[Dataset]
) -> Tuple[Set[BaseTag], Set[BaseTag]]:
    """Assign attributes to per-frame or shared functional groups.

    Parameters
    ----------
    instances: Sequence[pydicom.dataset.Dataset]
        DICOM SOP instances

    Returns
    -------
    Tuple[Set[pydicom.tag.BaseTag], Set[pydicom.tag.BaseTag]]
        Tags of attributes assigned to per-frame and shared functional groups

    """
```

I would prefer the use of keywords instead of tags, but I assume that this function should be capable of dealing with private attributes as well. Should it?

afshinmessiah:
> I would prefer the use of keywords instead of tags, but I assume that this function should be capable of dealing with private attributes as well. Should it?
It shouldn't I guess. I just preferred to have a list of tags instead of kws to speed up the code.
@hackermd, just pushed the new changes to code, moving the FrameSet and FrameSetCollection to highdicom.utils :)

fedorov:
>> I assume that this function should be capable of dealing with private attributes as well. Should it?
> It shouldn't I guess. I just preferred to have a list of tags instead of kws to speed up the code.

@afshinmessiah why should that function NOT be capable to operate on private attributes, can you explain?

afshinmessiah:
> > > I assume that this function should be capable of dealing with private attributes as well. Should it?
> 
> > It shouldn't I guess. I just preferred to have a list of tags instead of kws to speed up the code.
> 
> @afshinmessiah why should that function NOT be capable to operate on private attributes, can you explain?

Well, let me rephrase: 'it is not necessary', according to the current logic of the code, both per-frame and shared attributes are not allowed to be private.

hackermd:
In that case, I would prefer the function to have the following interface:

```python
def group_per_frame_and_shared_attributes(
    instances: Sequence[Dataset]
) -> Tuple[Set[str], Set[str]]:
    """Assign attributes to per-frame or shared functional groups.

    Parameters
    ----------
    instances: Sequence[pydicom.dataset.Dataset]
        DICOM SOP instances

    Returns
    -------
    Tuple[Set[str], Set[str]]
        Keywords of attributes assigned to per-frame and shared functional groups

    """
```

The function be able to replace the `FrameSet` class.

afshinmessiah:
> The main logic seems to be implemented in the `FrameSet._find_per_frame_and_shared_tags()` method, which updates the `FrameSet._shared_tags` and `FrameSet._perframe_tags` attributes. Could we just create a utility function (in `highdicom.sr.utils`) instead with the following signature?

@hackermd, I am more of a fan of objects and classes than of functions since they are easy to extend and well organized. Of course, we can rewrite any class and object in function form but I don't get this preference of function over the class. 
In this specific case, there are other functionalities of FrameSet class that I use throughout the conversion process. I believe this is too much of a change for this single PR. I suggest we settle for the current FrameSet and FrameSetCollection classes since they are optional and users don't have to use them. I the future versions you can refactor the structure to better deal with future usecases.

hackermd:
> I suggest we settle for the current FrameSet and FrameSetCollection classes since they are optional and users don't have to use them. I the future versions you can refactor the structure to better deal with future use cases.

That will not be that easy if we make them part of the public API. That's why I suggested keeping them private and only using them internally in the constructors for now. We can later revisit if additional use cases come up.

fedorov:
@afshinmessiah I don't necessarily agree with Markus, but let's just go ahead with this suggestion - can you make those private and use them only internally in the constructors?

afshinmessiah:
@hackermd and @CPBridge, I made both FrameSet and FrameSetCollection classes private. Let me know if anything else is left.

CPBridge (changes requested):
Hi @afshinmessiah ,

I managed to go through the rest of the code, and left a bunch of minor things to be addressed. Also please note my replies to a few of the unresolved previous threads.

Additionally, I notice that the tests occur entirely on synthetic data. Ideally there would be some tests that are on something closer to real world data. Fortunately, pydicom has a bunch of test files that we can use because they will already be available on any machine if pydicom is available.

```python
from pydicom.data import get_testdata_file

# Each of these will return a path to a directory containing an MR series
directory = get_testdata_file('MR700')
directory = get_testdata_file('MR1')
directory = get_testdata_file('MR2')

# Each of these will return a path to a directory containing a CT series
directory = get_testdata_file('CT2')
directory = get_testdata_file('CT5N')
```

I think it would be a really good idea if we used these more realistic files to test the conversion process, and could be done quickly quickly.

In my opinion after these are addressed, we should be ok to merge.

src/highdicom/legacy/sop.py
```diff
@@ -25,385 +37,391 @@
...
+                        _DicomHelper.istag_group_length(ttag) and not
+                        self._istag_excluded_from_perframe(ttag) and
+                        ttag != tag_for_keyword('PixelData')):
+                    elem = ds[ttag]
+                    if ttag not in self._perframe_tags:
+                        self._perframe_tags.append(ttag)
+                    if ttag in rough_shared:
+                        rough_shared[ttag].append(elem.value)
+                    else:
+                        rough_shared[ttag] = [elem.value]
```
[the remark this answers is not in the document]
afshinmessiah: I don't believe using fancy data containers would make the code more readable, since this awkward checking has long been the common way of handling similar use cases in python and not difficult to follow. But anyway, I will go ahead and apply what you suggest.

src/highdicom/legacy/sop.py
```diff
@@ -25,385 +37,391 @@
...
+            for ttag, elem in ds.items():
+                if (not ttag.is_private and not
+                    _DicomHelper.istag_file_meta_information_group(ttag) and not
+                        _DicomHelper.istag_repeating_group(ttag) and not
+                        _DicomHelper.istag_group_length(ttag) and not
+                        self._istag_excluded_from_perframe(ttag) and
+                        ttag != tag_for_keyword('PixelData')):
+                    elem = ds[ttag]
+                    if ttag not in self._perframe_tags:
+                        self._perframe_tags.append(ttag)
```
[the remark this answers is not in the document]
afshinmessiah: You are right, the order is not important and it is a good idea to use set instead of list. Since I don't want to do too much of a change to the code, I'll collect the tags in sets and finally convert them into lists.

src/highdicom/legacy/sop.py
```diff
@@ -25,385 +37,391 @@
...
-
-    mf_dataset.AcquisitionContextSequence = []
-
-    # TODO: Encapsulated Pixel Data with compressed frame items.
+                all_values_are_equal = all(
+                    _DicomHelper.isequal(v_i, v[0]) for v_i in v)
+                if not all_values_are_equal:
+                    to_be_removed_from_shared.append(ttag)
+        for t in to_be_removed_from_shared:
+            del rough_shared[t]
```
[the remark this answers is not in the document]
afshinmessiah: Good point, you are right.

src/highdicom/legacy/sop.py
```diff
@@ -417,69 +435,2075 @@ def __init__(
...
+            frame_set.distinguishing_attributes_tags)
+        self._excluded_from_perframe_tags = self._get_tag_used_dictionary(
+            frame_set.excluded_from_perframe_tags)
+        self._perframe_tags = self._get_tag_used_dictionary(
+            frame_set.perframe_tags)
+        self._shared_tags = self._get_tag_used_dictionary(
+            frame_set.shared_tags)
+        self.excluded_from_functional_groups_tags = {
+            tag_for_keyword('SpecificCharacterSet'): False}
+        self.__build_blocks: List[Any] = []
```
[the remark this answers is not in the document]
afshinmessiah: This is due to the previous design of the code based on an abstract interface to force subclasses to implement their own configuration of build blocks. Now it doesn't have nay use and I will change it.

src/highdicom/legacy/sop.py
```diff
@@ -417,69 +435,2075 @@ def __init__(
...
+        Returns
+        -------
+        Union[list, None]
+            A new list of FrameType value is returned. If attrib is not of type
+            DataElement None is returned.
+
+        """
+        if not isinstance(attrib, DataElement):
+            return None
+        output = ['', '', '', '']
```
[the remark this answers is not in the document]
afshinmessiah: I don't get it. This will be the output of the function.

src/highdicom/legacy/sop.py
```diff
@@ -417,69 +435,2075 @@ def __init__(
...
+                self._legacy_datasets[i], item)
+
+            self.earliest_frame_acquisition_date_time =\
+                self.farthest_future_date_time
+
+    def _build_slices_geometry_frame_content(self) -> None:
+        """Instantiates an object of _GeometryOfSlice for each slice."""
+        frame_count = len(self._legacy_datasets)
+        for i in range(0, frame_count):
+            curr_frame = self._legacy_datasets[i]
```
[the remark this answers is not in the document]
afshinmessiah: Why less efficient??

src/highdicom/legacy/sop.py
```diff
@@ -417,69 +435,2075 @@ def __init__(
...
+        slice_count = len(self._slices)
+        if slice_count >= 2:
+            last_slice = self._slices[0]
+            for i in range(1, slice_count):
+                curr_slice = self._slices[i]
+                if not _GeometryOfSlice.are_parallel(
+                        curr_slice, last_slice, self._tolerance):
+                    return False
+                last_slice = curr_slice
+            return True
```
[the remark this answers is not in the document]
afshinmessiah: Although it rarely happens, you are totally correct.

afshinmessiah:
Thanks, @CPBridge for the comments. I tried to address almost all comments.
>I think it would be a really good idea if we used these more realistic files to test the conversion process, and could be done quite quickly.

Well, the good thing about synthetic test cases is that I have control over the number of frames, framesets, and geometry while generating the data. I agree with you on the preference for real data over synthetic data. However, it is complicated to design test cases for real data (to address all modules being added through the conversion process) when the expected multi-frame is unknown. Especially when the whole structure of FrameSet, FrameSetCollection, and GeometryOfSlice classes are subject to change. I'd rather stick with current tests and for this PR since I'm short of time cleaning up the code a getting it ready to merge.

CPBridge:
Thanks @afshinmessiah for addressing these issues so quickly. Understood regarding the tests - I will try to address this in a future PR

fedorov:
@hackermd as part of this PR we should add a statement to the README perhaps that this work was supported by IDC. Let me know if you agree, and I can add the corresponding text to the PR.

hackermd:
@fedorov I agree. Please note that I recently added a line to the README (see [here](https://github.com/MGHComputationalPathology/highdicom/blob/17d9580571d7416541b101c4d9bc7161c3e7b174/README.md)) to highlight that the development of the library has been supported by the IDC. Please feel free to either rephrase or add an additional statement.

malaterre:
@afshinmessiah Would it be possible to rebase your branch on git/master ? thanks

phistep:
Hello, can someone summarize the state of this issue? I came here from https://github.com/pydicom/pydicom/discussions/1705#discussioncomment-3792994, since I want to get rid of our `dcm4che` dependency and use pure python tools for the jobs.

It seems to me that this feature is mostly done, some minor issues were being addressed and then the PR was abandoned. @afshinmessiah, do you have any intentions of continuing your work here? I might be interested in taking over and getting this PR merged. Do you have any estimate of how long it might take?

CPBridge:
Hi @phistep. Thanks for reaching out about this. Ashfin contributed this PR some time ago but the core team (@hackermd  and myself) had some disagreements about style and API design, and then he changed roles and could no longer dedicate time to it.

Since then it has always been my intention to pick this up again and get it merged because there are a lot of valuable changes here only a few minor sticking point in the way, but then there has always been something more urgent... However, now is as good a time as any I suppose. I will have to take some time to go back through the comments and remind myself of the points that need to be addressed. But from memory they are along these lines:

- Unpythonic coding style in several places (grouping functions into classes unnecessarily etc)
- Making the geometry related objects and code more general and putting into a separate module. This has only grown in importance as I plan to make some changes to seg soon that will require these changes. It may be best if I deal with this part.
- Backwards compatibility with released versions of the library. This PR badly breaks backwards compatibility, which we would like to avoid without very good reason. We were hoping to find a way to enable the new logic without breaking API compatibility but couldn't agree on one. I would have to review things again to remember the specifics.

It is worth noting that the legacy conversion process in the released version of the library does work, but that this pull request contains more sophisticated sorting of frames and a few fixes.

If you are willing to help, that could certainly be very useful. Let me find some time in the next few days to review properly and follow up with some more detail

fedorov:
It would be great to have this integrated - I am not the one who wrote the code, but together with @dclunie we did supervise Afshin, so should be able to address at least some questions. Happy to meet to chat about this Chris!

phistep:
Okay, really cool! Then I will start reading the changeset in-depth and maybe get back to you with some questions and a roadmap.

Issue #38, opened 2021-01-19
Incorrectly named IOD_MODULE_MAP key for positron emission tomography


vanossj:
https://github.com/MGHComputationalPathology/highdicom/blob/a9f3916a9b98e09ff5f359fe77f508746e56b730/src/highdicom/_iods.py#L7825

Which SOPInstanceUID is the key `pet-image` supposed to link to? I don't see a pet image storage on the official [list](http://dicom.nema.org/dicom/2013/output/chtml/part04/sect_B.5.html).

Is it referring to _Positron Emission Tomography Image Storage_ SOP 1.2.840.10008.5.1.4.1.1.128?

If so the key should be changed to `positron-emission-tomography-image`

If this isn't the issue what key in `IOD_MODULE_MAP` does SOP 1.2.840.10008.5.1.4.1.1.128 map to?

hackermd:
@vanossj  thank you for reporting the issue and for fixing it. We will draft a new release within the next days and make the new version available on PyPI.

PR #39, opened 2021-01-19
look up IOD key from SOP Class UID

vanossj:
Added SOP Class UID to IOD key lookup map instead of generating IOD key from SOP Class Name, which is not always consistent, as shown in #38.


hackermd (approved):
Thanks @vanossj!

PR #40, opened 2021-02-23
Make Algoritithm Identification Sequence optional when using manual segmentations

CPBridge:
Currently, the `algorithm_identification` parameter is always a required parameter for a `SegmentDescription` object. (notably the the docstring lists it as optional and therefore does not match the actual `__init__` method's signature).

The SegmentationAlgorithmIdentificationSequence is a type 3 (optional) tag within a SegmentSequence ([see here](https://dicom.innolitics.com/ciods/segmentation/segmentation-image/00620002/00620007)). Currently highdicom also deduces the value of the SegmentAlgorithmName tag from the `algorithm_identifier` parameter. The SegmentAlgorithmName is a type 1C tag, required on the condition that the algorithm type is not `MANUAL` ([see here](https://dicom.innolitics.com/ciods/segmentation/segmentation-image/00620002/00620009)).

In the spirit of being opinionated, I think it is reasonable to require the full algorithm identification parameter for automatic or semi-automatic segmentations even though this is stricter than the standard. However it probably does not make sense to require this parameter when the segmentation is a manual one.

Therefore this PR implements the logic that the `algorithm_identification` is required unless the `algorithm_type` is `MANUAL`.

### DIFF
--- a/src/highdicom/seg/content.py
+++ b/src/highdicom/seg/content.py
@@ -28,7 +28,7 @@ class SegmentDescription(Dataset):
             segmented_property_category: Union[Code, CodedConcept],
             segmented_property_type: Union[Code, CodedConcept],
             algorithm_type: Union[SegmentAlgorithmTypeValues, str],
-            algorithm_identification: AlgorithmIdentificationSequence,
+            algorithm_identification: Optional[AlgorithmIdentificationSequence] = None,
             tracking_uid: Optional[str] = None,
             tracking_id: Optional[str] = None,
             anatomic_regions: Optional[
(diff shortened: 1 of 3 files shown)


hackermd (approved):

hackermd:
@CPBridge there are a couple of PEP8 style compliance issues (see [unit test output](https://travis-ci.org/github/MGHComputationalPathology/highdicom/jobs/760206996)). Could you kindly address those?

CPBridge:
Should be fixed. Tests are still failing because of coverage, but I don't believe that this is due to the changes in this PR @hackermd

hackermd (approved):

hackermd:
> Should be fixed. Tests are still failing because of coverage, but I don't believe that this is due to the changes in this PR

Hmm...maybe because extra lines where added and the coverage dropped. Shall we just lower the threshold for now?

PR #42, opened 2021-02-26
CID Links in Documentation

CPBridge:
Every time a docstring refers to specific CID, I have added a hyperlink to the DICOM standard's page for that CID. I hope this makes using the docs considerably more convenient.

I checked every link and in so doing found and corrected a mistake: Time Point Type (CID 6146) was mistakenly listed as CID 646 "Preclinical Small Animal Imaging Procedures". Oopsie

In case it's ever useful again, here is a handy vim substitution command:
```
:%s#CID \([0-9]*\)#`CID \1 <http://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_CID_\1.html>`_\r           #c
```


hackermd (approved):

hackermd:
> In case it's ever useful again, here is a handy vim substitution command:

Nice :)

PR #43, opened 2021-02-26
Implement constructors directly from datasets for reference objects

CPBridge:
In the SR module there are a number of classes that represent various types of references to source images:

- highdicom.sr.content.ReferencedSegment
- highdicom.sr.content.ReferencedSegmentationFrame
- highdicom.sr.content.SourceImageForMeasurement
- highdicom.sr.content.SourceImageForRegion
- highdicom.sr.content.SourceImageForSegmentation
- highdicom.sr.content.SourceSeriesForSegmentation

The constructors for these classes require that the user manually pass the relevant UIDs. This is by no means a large burden but is a little low-level and increases the probability of a mistake.

I propose implementing alternative constructors as class methods called `from_dataset` that allow constructing objects by simply passing the relevant dataset and have highdicom take care of the rest.

For example

```python
import pydicom
from highdicom.sr.content import SourceImageForMeasurement

ct_dataset = pydicom.dcmread(...)

# Currently a source image object must be constructed like this
source_image = SourceImageForMeasurement(
    referenced_sop_class_uid=ct_dataset.SOPClassUID,
    referenced_sop_instance_uid=ct_dataset.SOPInstanceUID
)

# Under the proposal, it could also alternatively be constructed like this
source_image = SourceImageForMeasurement.from_dataset(ct_dataset)
```

I personally think this is a bit higher level and a nice reduction in boilerplate.

For now, I have just implemented this for the `SourceImageForMeasurement` class. If you like the idea, I will implement for the other classes and write tests.


hackermd:
@CPBridge I like the idea!

What do you think about calling the classmethod `from_source()` instead of `from_dataset()`? The constructor of these classes will also create a `Dataset`, therefore this could get confusing.

CPBridge:
Cool, I'll get to work on the rest then. I agree that dataset is a little confusing as all these objects are datasets. I could get behind `from_source()`. Or maybe the slightly more verbose `from_source_dataset()`, since `source` is also a bit vague?

CPBridge:
@hackermd this is now implemented and ready for review. The referenced segment and referenced segmentation frame versions were more complex than I thought since there are a few different ways that source information may appear in a segmentation dataset. But this also means it will be more useful. Hopefully I've covered most eventualities.

CPBridge:
@hackermd this is awaiting rewiew

src/highdicom/sr/content.py
```diff
@@ -103,6 +105,34 @@ def __init__(
             relationship_type=RelationshipTypeValues.INFERRED_FROM
         )
 
+    @classmethod
+    def from_source_dataset(
```
hackermd, line 109: Given that we are considering adding a `from_dataset()`, it may be better to call this method just `from_source_image()`. What do you think?

P.S.: If we do this, we should apply the same logic to the other methods as well.

src/highdicom/sr/content.py
```diff
@@ -599,3 +838,163 @@ def __init__(
...
+        # groups sequence
+        # This gives more specific information, but depends upon optional
+        # attributes so may fail with some segmentation objects
+        # Those created with highdicom should have this information
+        referenced_frame_info = []
+        if frame_numbers is not None:
+            # Use only the provided frames
+            for f in frame_numbers:
+                # Check that the provided frame number is valid
+                if f < 0 or f >= dataset.NumberOfFrames:
```
hackermd, line 900: Frame numbers are one-based.

```suggestion
                if f < 1 or f >= dataset.NumberOfFrames:
```

src/highdicom/sr/content.py
```diff
@@ -599,3 +838,163 @@ def __init__(
...
+                    raise ValueError(
+                        f'Frame {f} is an invalid frame number within the '
+                        'provided dataset'
+                    )
+
+                frame_info = dataset.PerFrameFunctionalGroupsSequence[f]
```
hackermd, line 906:
```suggestion
                i = f - 1
                frame_info = dataset.PerFrameFunctionalGroupsSequence[i]
```

src/highdicom/sr/content.py
```diff
@@ -599,3 +838,163 @@ def __init__(
...
+
+                if ref_segment != segment_number:
+                    raise ValueError(
+                        f'The provided frame number {f} does not refer to '
+                        f'segment number {segment_number}'
+                    )
+                referenced_frame_info.append(frame_info)
+        else:
+            # Use every frame that applies to this segment
+            for frame_info in dataset.PerFrameFunctionalGroupsSequence:
```
hackermd, line 921: Since there may be many frames in a Segmentation instance, I would rewrite this into a list comprehension for improved performance.

src/highdicom/sr/content.py
```diff
@@ -599,3 +838,163 @@ def __init__(
...
+                'Input dataset should be a segmentation storage instance'
+            )
+
+        source_images = []
+        source_series = None
+
+        # Attempt to retrieve a list of source images for the specific segment
+        # and frames specified using information in the per-frame functional
+        # groups sequence
+        # This gives more specific information, but depends upon optional
```
hackermd, line 892: Which attributes are optional? Based on my understanding, all the used attributes are required. Am I missing something?

hackermd: Ah...this seems to not only refer to the next block, but also to the subsequent block.

src/highdicom/sr/content.py
```diff
@@ -599,3 +838,163 @@ def __init__(
...
+                raise ValueError(
+                    f'No frame information found referencing segment '
+                    f'{segment_number}'
+                )
+
+        # Gather up references using the Derivation Image Sequences of the
+        # referenced frames
+        refd_insuids = set()
+        for frame_info in referenced_frame_info:
+            if hasattr(frame_info, 'DerivationImageSequence'):
```
hackermd, line 937: The two lines could be combined into one:

```python
for drv_image in getattr(frame_info, 'DerivationImageSequence', []):
    ...
```

src/highdicom/sr/content.py
```diff
@@ -599,3 +838,163 @@ def __init__(
...
+                for drv_image in frame_info.DerivationImageSequence:
+                    if hasattr(drv_image, 'SourceImageSequence'):
+                        for src_image in drv_image.SourceImageSequence:
```
hackermd, line 940: The two lines could be combined into one:

```python
for src_image in getattr(drv_image, 'SourceImageSequence', []):
    ...
```

src/highdicom/sr/content.py
```diff
@@ -599,3 +838,163 @@ def __init__(
...
+                            # Check to avoid duplication of instances
+                            ins_uid = src_image.ReferencedSOPInstanceUID
+                            cls_uid = src_image.ReferencedSOPClassUID
+                            if ins_uid not in refd_insuids:
+                                if hasattr(src_image, 'ReferencedFrameNumber'):
+                                    ref_frames = src_image.ReferencedFrameNumber
+                                else:
+                                    ref_frames = None
```
hackermd, line 948:
```suggestion
                                ref_frames = getattr(src_image, 'ReferencedFrameNumber', None)
```

src/highdicom/sr/content.py
```diff
@@ -599,3 +838,163 @@ def __init__(
...
+             applies to all segments and frames present in the segmentation
+             instance.
+
+        """
+        if dataset.SOPClassUID != SegmentationStorage:
+            raise ValueError(
+                'Input dataset should be a segmentation storage instance'
+            )
+
+        source_images = []
```
hackermd, line 886: Why not `source_images = None`?

src/highdicom/sr/content.py
```diff
@@ -599,3 +838,163 @@ def __init__(
...
+                            s.ReferencedSOPClassUID,
+                            s.ReferencedSOPInstanceUID
+                        )
+                        for s in ref_series.ReferencedInstanceSequence
+                    ]
+                elif hasattr(ref_series, 'SeriesInstanceUID'):
+                    source_series = SourceSeriesForSegmentation(
+                        ref_series.SeriesInstanceUID
+                    )
+                    source_images = []
```
hackermd, line 981: Is this necessary? This does not (and is not supposed to) change the value assigned to the variable, does it?

src/highdicom/sr/content.py
```diff
@@ -599,3 +838,163 @@ def __init__(
...
+                    "Dataset does not contain a ReferencedSeriesSequence "
+                    "containing information about the source images"
+                )
+
+        return cls(
+            sop_class_uid=dataset.SOPClassUID,
+            sop_instance_uid=dataset.SOPInstanceUID,
+            segment_number=segment_number,
+            frame_numbers=frame_numbers,
+            source_images=source_images if source_images else None,
```
hackermd, line 998: This may not be necessary. See above comments.

src/highdicom/sr/content.py
```diff
@@ -599,3 +838,163 @@ def __init__(
...
+                    )
+                    source_images = []
+                else:
+                    raise AttributeError(
+                        "The ReferencedSeriesSequence in the segmentation "
+                        "dataset does not contain the expected information"
+                    )
+            else:
+                raise AttributeError(
+                    "Dataset does not contain a ReferencedSeriesSequence "
```
hackermd, line 989: I would clarify that the **segmentation** dataset does not contain the attribute (similar to the above error message):

```suggestion
                    "Segmentation dataset does not contain a ReferencedSeriesSequence "
```

tests/test_sr.py
```diff
@@ -819,6 +821,660 @@ def test_finding_site(self):
...
+class TestSourceImageForSegmentation(unittest.TestCase):
+
+    def setUp(self):
+        super().setUp()
+        file_path = Path(__file__)
+        data_dir = file_path.parent.parent.joinpath('data')
+        self._src_dataset = dcmread(
+            str(data_dir.joinpath('test_files', 'ct_image.dcm'))
+        )
+        self._ref_frames = [0, 4, 5]
```
hackermd, line 833: Frame numbers shall start at 1:

```suggestion
        self._ref_frames = [1, 4, 5]
```

hackermd: Applies to the other test cases as well.

tests/test_sr.py
```diff
@@ -819,6 +821,660 @@ def test_finding_site(self):
...
+
+    def test_basic(self):
```
hackermd, line 835: I would suggest naming the method `test_construction` instead for clarity.

Applies to the other test cases as well.

tests/test_sr.py
```diff
@@ -819,6 +821,660 @@ def test_finding_site(self):
...
+        assert (
+            src_image.ReferencedSOPSequence[0].ReferencedSOPClassUID ==
+            self._src_dataset.SOPClassUID
+        )
+        assert (
+            src_image.ReferencedSOPSequence[0].ReferencedSOPInstanceUID ==
+            self._src_dataset.SOPInstanceUID
+        )
+
+    def test_basic_frames(self):
```
hackermd, line 850: I would suggest naming the method `test_construction_with_frame_reference` instead for clarity.

hackermd (changes requested):

src/highdicom/sr/content.py
```diff
@@ -169,6 +170,34 @@ def __init__(
...
+            referenced image is a multi-frame image
+
+        Returns
+        -------
+        SourceImageForRegion
+            Content item representing a reference to the image dataset
+
+        """
+        return cls(
+            referenced_sop_class_uid=dataset.SOPClassUID,
```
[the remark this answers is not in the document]
CPBridge: Is there a well-defined list of SOP Classes that would be acceptable here? Seems like we don't want to be overly restrictive. Or would we instead create a blacklist containing obvious mistakes such as Segmentation, SRs, GSPS etc?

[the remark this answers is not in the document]
hackermd: If it walks like a duck and talks like a duck...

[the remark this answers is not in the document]
CPBridge: ...then we can say with a high degree of certainty that (0010,2201) contains "Anatidae"?

src/highdicom/sr/content.py
```diff
@@ -523,6 +655,113 @@ def __init__(
...
+        Returns
+        -------
+        ReferencedSegment
+            Content item representing a reference to the segment
+
+        Notes
+        -----
+             This method will attempt to deduce source image information
+             from information provided in the segmentation instance. If
+             available, it will used information specific to the segment
```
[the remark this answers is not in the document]
CPBridge: Adressed in 4bea49d

src/highdicom/sr/content.py
```diff
@@ -523,6 +655,113 @@ def __init__(
...
+             and frame numbers (if any) provided using the Derivation
+             Image Sequence information in the frames. If this information
```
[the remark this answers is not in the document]
CPBridge: Addressed in 4bea49d

src/highdicom/sr/content.py
```diff
@@ -523,6 +655,113 @@ def __init__(
...
+
+        if frame_number < 0 or frame_number >= dataset.NumberOfFrames:
+            raise ValueError(
+                f'Frame {frame_number} is an invalid frame number within the '
+                'provided dataset'
+            )
+
+        frame_info = dataset.PerFrameFunctionalGroupsSequence[frame_number]
+
+        segment_number = frame_info.SegmentIdentificationSequence[0]\
```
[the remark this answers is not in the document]
CPBridge: Addressed in 4bea49d

src/highdicom/sr/content.py
```diff
@@ -523,6 +655,113 @@ def __init__(
...
+                'Input dataset should be a segmentation storage instance'
+            )
+
+        if frame_number < 0 or frame_number >= dataset.NumberOfFrames:
+            raise ValueError(
+                f'Frame {frame_number} is an invalid frame number within the '
+                'provided dataset'
+            )
+
+        frame_info = dataset.PerFrameFunctionalGroupsSequence[frame_number]
```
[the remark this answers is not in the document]
CPBridge: Good point. Eugh one-based indexing... 
Addressed in 4bea49d

src/highdicom/sr/content.py
```diff
@@ -523,6 +655,113 @@ def __init__(
...
+                raise ValueError(
+                    'Could not deduce a single source image from the '
+                    'provided dataset'
+                )
+            drv_image = frame_info.DerivationImageSequence[0]
+
+            if hasattr(drv_image, 'SourceImageSequence'):
+                if len(drv_image.SourceImageSequence) != 1:
+                    raise ValueError(
+                        'Could not deduce a single source image from the '
```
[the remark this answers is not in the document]
CPBridge: Addressed in 4bea49d

src/highdicom/sr/content.py
```diff
@@ -523,6 +655,113 @@ def __init__(
...
+                        src.ReferencedSOPInstanceUID
+                    )
+                else:
+                    raise AttributeError(
+                        "The ReferencedSeriesSequence in the segmentation "
+                        "dataset does not contain the expected information"
+                    )
+            else:
+                raise AttributeError(
+                    "Dataset does not contain a ReferencedSeriesSequence "
```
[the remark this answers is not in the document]
CPBridge: Addressed in 4bea49d

src/highdicom/sr/content.py
```diff
@@ -599,3 +838,163 @@ def __init__(
...
+            of the given segment number
+
+        Returns
+        -------
+        ReferencedSegment
+            Content item representing a reference to the segment
+
+        Notes
+        -----
+             This method will attempt to deduce source image information
```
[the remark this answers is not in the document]
CPBridge: Nope, probably a vim error...
Addressed in 4bea49d

src/highdicom/sr/content.py
CPBridge: Yeah, agreed. Changed for all relevant classes

src/highdicom/sr/content.py
CPBridge: Addressed in 4bea49d

src/highdicom/sr/content.py
CPBridge: Addressed in 4bea49d

src/highdicom/sr/content.py
CPBridge: Yeah exactly. Have clarified the comments somewhat in  4bea49d

src/highdicom/sr/content.py
CPBridge: Neat. Addressed in 4bea49d

src/highdicom/sr/content.py
CPBridge: Addressed in 4bea49d

src/highdicom/sr/content.py
CPBridge: Addressed in 4bea49d

src/highdicom/sr/content.py
CPBridge: Nope, should be redundant. Removed in 4bea49d

src/highdicom/sr/content.py
CPBridge: Because we need to append to it later in the loop beginning:

```python
for frame_info in referenced_frame_info:
```
currently on line 1006

src/highdicom/sr/content.py
CPBridge: Addressed in 4bea49d

tests/test_sr.py
CPBridge: Addressed in 4bea49d. Also added a test on frame number (previously there was no check)

tests/test_sr.py
CPBridge: Addressed in 4bea49d

tests/test_sr.py
CPBridge: Addressed in 4bea49d

src/highdicom/sr/content.py
CPBridge: Addressed in 4bea49d

CPBridge:
I have addressed most comments. Will look into checks on the validity of source images (as suggested in the comments) later.

src/highdicom/sr/content.py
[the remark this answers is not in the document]
hackermd: [PS3.3 A.1.2.6 Image IE](http://dicom.nema.org/medical/dicom/current/output/html/part03.html#sect_A.1.2.6)

> The Image IE defines the Attributes that describe the pixel data of an image. 

However, that definition includes Segmentation images.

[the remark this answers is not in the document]
hackermd: > Or would we instead create a blacklist containing obvious mistakes such as Segmentation, SRs, GSPS etc?

Maybe we should just assert that the `PixelData` attribute (and its friends for floating point values) is present and that `Modality` is not obviously wrong (`SR`, `SEG`, etc.) without being overly restrictive or comprehensive.

hackermd:
> I have addressed most comments.

Could you kindly respond to the comments with `Addressed with #COMMIT_HASH`. This will make it easier for me to keep track of changes and determine whether a comment can be resolved.

CPBridge:
Didn't I do that?

hackermd:
> Didn't I do that?

You did indeed. Excellent! Sorry I didn't notice. The comments were already resolved and hidden. Please keep comments open and let me close them. That makes it easier for me to keep track of changes.

src/highdicom/sr/content.py
```diff
@@ -22,6 +24,53 @@
 )
 
 
+def _check_valid_source_image_dataset(dataset: Dataset) -> None:
+    """Raise an error if the image is not a valid for a source image reference.
```
hackermd, line 28:
```suggestion
    """Raise an error if the image is not a valid source image reference.
```

src/highdicom/sr/content.py
```diff
@@ -103,6 +163,35 @@ def __init__(
             relationship_type=RelationshipTypeValues.INFERRED_FROM
         )
 
+    @classmethod
+    def from_source_image(
+            cls,
+            image_dataset: Dataset,
+            referenced_frame_numbers: Optional[Sequence[int]] = None
+        ) -> 'SourceImageForMeasurement':
```
hackermd, line 171: Let's start doing this consistently throughout the library and in compliance with PEP8:

```suggestion
        cls,
        image_dataset: Dataset,
        referenced_frame_numbers: Optional[Sequence[int]] = None
    ) -> 'SourceImageForMeasurement':
```

src/highdicom/sr/content.py
```diff
@@ -103,6 +163,35 @@ def __init__(
...
+        ----------
+        image_dataset: pydicom.dataset.Dataset
+            Dataset representing the image to be referenced
+        referenced_frame_numbers: Sequence[int], optional
+            numbers of the frames to which the reference applies in case the
+            referenced image is a multi-frame image
+
+        Returns
+        -------
+        SourceImageForMeasurement
```
hackermd, line 184:
```suggestion
        highdicom.sr.content.SourceImageForMeasurement
```

src/highdicom/sr/content.py
```diff
@@ -140,6 +240,35 @@ def __init__(
             relationship_type=RelationshipTypeValues.SELECTED_FROM
         )
 
+    @classmethod
+    def from_source_image(
+            cls,
+            image_dataset: Dataset,
+            referenced_frame_numbers: Optional[Sequence[int]] = None
+        ) -> 'SourceImageForRegion':
```
hackermd, line 248:
```suggestion
        cls,
        image_dataset: Dataset,
        referenced_frame_numbers: Optional[Sequence[int]] = None
    ) -> 'SourceImageForRegion':
```

src/highdicom/sr/content.py
```diff
@@ -140,6 +240,35 @@ def __init__(
...
+        ----------
+        image_dataset: pydicom.dataset.Dataset
+            Dataset representing the image to be referenced
+        referenced_frame_numbers: Sequence[int], optional
+            numbers of the frames to which the reference applies in case the
+            referenced image is a multi-frame image
+
+        Returns
+        -------
+        SourceImageForRegion
```
hackermd, line 261:
```suggestion
        highdicom.sr.content.SourceImageForRegion
```

src/highdicom/sr/content.py
```diff
@@ -177,6 +317,35 @@ def __init__(
             relationship_type=RelationshipTypeValues.CONTAINS
         )
 
+    @classmethod
+    def from_source_image(
+            cls,
+            image_dataset: Dataset,
+            referenced_frame_numbers: Optional[Sequence[int]] = None
+        ) -> 'SourceImageForSegmentation':
```
hackermd, line 325:
```suggestion
        cls,
        image_dataset: Dataset,
        referenced_frame_numbers: Optional[Sequence[int]] = None
    ) -> 'SourceImageForSegmentation':
```

src/highdicom/sr/content.py
```diff
@@ -177,6 +317,35 @@ def __init__(
...
+        ----------
+        image_dataset: pydicom.dataset.Dataset
+            Dataset representing the image to be referenced
+        referenced_frame_numbers: Sequence[int], optional
+            numbers of the frames to which the reference applies in case the
+            referenced image is a multi-frame image
+
+        Returns
+        -------
+        SourceImageForSegmentation
```
hackermd, line 338:
```suggestion
        highdicom.sr.content.SourceImageForSegmentation
```

src/highdicom/sr/content.py
```diff
@@ -202,6 +371,30 @@ def __init__(self, referenced_series_instance_uid: str):
             relationship_type=RelationshipTypeValues.CONTAINS
         )
 
+    @classmethod
+    def from_source_image(
+            cls,
+            image_dataset: Dataset,
+        ) -> 'SourceSeriesForSegmentation':
```
hackermd, line 378:
```suggestion
        cls,
        image_dataset: Dataset,
    ) -> 'SourceSeriesForSegmentation':
```

src/highdicom/sr/content.py
```diff
@@ -202,6 +371,30 @@ def __init__(self, referenced_series_instance_uid: str):
...
+
+        Parameters
+        ----------
+        image_dataset: pydicom.dataset.Dataset
+            dataset representing a single image from the series to be
+            referenced
+
+        Returns
+        -------
+        SourceSeriesForSegmentation
```
hackermd, line 389:
```suggestion
        highdicom.sr.content.SourceSeriesForSegmentation
```

src/highdicom/sr/content.py
```diff
@@ -414,6 +607,33 @@ def __init__(self, referenced_sop_instance_uid: str):
...
+
+        Parameters
+        ----------
+        value_map_dataset: pydicom.dataset.Dataset
+            dataset representing the real world value map to be
+            referenced
+
+        Returns
+        -------
+        RealWorldValueMap
```
hackermd, line 625:
```suggestion
        highdicom.sr.content.RealWorldValueMap
```

src/highdicom/sr/content.py
```diff
@@ -414,6 +607,33 @@ def __init__(self, referenced_sop_instance_uid: str):
             relationship_type=RelationshipTypeValues.CONTAINS
         )
 
+    @classmethod
+    def from_source_value_map(
+            cls,
+            value_map_dataset: Dataset,
+        ) -> 'RealWorldValueMap':
```
hackermd, line 614:
```suggestion
        cls,
        value_map_dataset: Dataset,
    ) -> 'RealWorldValueMap':
```

src/highdicom/sr/content.py
```diff
@@ -524,6 +744,120 @@ def __init__(
             )
         self.append(source_image)
 
+    @classmethod
+    def from_seg_dataset(
+            cls,
+            dataset: Dataset,
+            frame_number: int
+        ) -> 'ReferencedSegmentationFrame':
```
hackermd, line 752:
```suggestion
        cls,
        dataset: Dataset,
        frame_number: int
    ) -> 'ReferencedSegmentationFrame':
```

src/highdicom/sr/content.py
```diff
@@ -524,6 +744,120 @@ def __init__(
...
+        dataset: pydicom.dataset.Dataset
+            Dataset representing a segmentation containing the referenced
+            segment.
+        frame_number: int
+            Number of the frame to which the reference applies. Note that
+            one-based indexing is used to index the frames.
+
+        Returns
+        -------
+        ReferencedSegment
```
hackermd, line 766:
```suggestion
        highdicom.sr.content.ReferencedSegment
```

src/highdicom/sr/content.py
```diff
@@ -524,6 +744,120 @@ def __init__(
...
+        This method will attempt to deduce source image information
+        from information provided in the segmentation instance. If
+        available, it will use information specific to the segment
+        and frame numbers (if any) provided using the Derivation
+        Image Sequence item for the given frame. If this information
+        is not present in the segmentation dataset, it will instead
+        use the information in the Referenced Series Sequence, which
+        applies to all segments and frames present in the segmentation
+        instance.
+
```
hackermd, line 780: Please add a `Raises` section to indicate under what circumstances a `ValueError` will be raised

src/highdicom/sr/content.py
```diff
@@ -600,3 +934,176 @@ def __init__(
...
                 '"source_images" or "source_series".'
             )
+
+    @classmethod
+    def from_seg_dataset(
+            cls,
+            dataset: Dataset,
+            segment_number: int,
+            frame_numbers: Optional[Sequence[int]] = None
+        ) -> 'ReferencedSegment':
```
hackermd, line 944:
```suggestion
        cls,
        dataset: Dataset,
        segment_number: int,
        frame_numbers: Optional[Sequence[int]] = None
    ) -> 'ReferencedSegment':
```

src/highdicom/sr/content.py
```diff
@@ -600,3 +934,176 @@ def __init__(
...
+            number of the segment to reference within the provided dataset
+        frame_numbers: Optional[Sequence[int]], optional
+            list of frames in the segmentation dataset to reference. If
+            not provided, the reference is assumed to apply to all frames
+            of the given segment number. Note that frame numbers are
+            indexed with 1-based indexing.
+
+        Returns
+        -------
+        ReferencedSegment
```
hackermd, line 962:
```suggestion
        highdicom.sr.content.ReferencedSegment
```

src/highdicom/sr/content.py
hackermd: I see. It's potentially error prone due to the `source_images if source_images else None`, but I don't have a better suggestion either.

src/highdicom/sr/content.py
```diff
@@ -103,6 +163,35 @@ def __init__(
             relationship_type=RelationshipTypeValues.INFERRED_FROM
         )
 
+    @classmethod
+    def from_source_image(
+            cls,
+            image_dataset: Dataset,
```
hackermd, line 169: Do you have strong feelings about using `image_dataset` instead of `image`? I generally avoid including the type in the name. We have a type annotation and a docstring to inform the user.

hackermd (changes requested):
Nice! Only a few fixes and suggestions for docstrings.

src/highdicom/sr/content.py
CPBridge: Resolved in 4cc5f0e

src/highdicom/sr/content.py
CPBridge: Just that 'image' is very vague and could be a numpy pixel array or something. On balance though it is probably better. Have changed to `image` in this class and all other `SourceXForX` classes, and additionally changed the `dataset` parameter of `ReferencedSegment.from_seg_dataset()` to `segmentation` to match. Changes in 4cc5f0e

On a related note, should we change `from_seg_dataset` to simply `from_seg` or `from_segmentation` to better match `from_source_image`?

CPBridge:
I have addressed all comments and also added:
- Checks that ensure that referenced frame numbers are valid for the specific dataset passed (previously this was only done for frame numbers that are always incorrect, i.e. < 1)
- Tests for the checks on dataset appropriateness and frame number correctness

src/highdicom/sr/content.py
```diff
@@ -938,15 +989,15 @@ def __init__(
     @classmethod
     def from_seg_dataset(
```
hackermd, line 990: Given that we call the related methods `from_source_image()` and `from_source_value_map()`, I wonder whether we should call this method `from_segmentation()` for consistency. What do you think?

hackermd (commented):
Thanks @CPBridge! Looks great! I have just one additional comment regarding the name of one of the methods. Please let me know what you think.

src/highdicom/sr/content.py
CPBridge: I asked you the same question above! I'll make the change

CPBridge: Changed as suggested in 9aca15a @hackermd

hackermd: > I asked you the same question above! I'll make the change

I just realized that, too. We are on the same page 👍

hackermd (approved):

Issue #45, opened 2021-02-28
Simple TimePointContext constructor fails


CPBridge:
Trying to use the constructor for the `TimePointContext` object seems to result in an exception regardless of the input parameters.

```python
from highdicom.sr.templates import TimePointContext

TimePointContext('whatever')
```

Results in:
```
Traceback (most recent call last):
  File "test.py", line 3, in <module>
    TimePointContext('whatever')
  File "/Users/christopher.bridge/Developer/highdicom/src/highdicom/sr/templates.py", line 214, in __init__
    self.append(time_point_item)
  File "/Users/christopher.bridge/Developer/highdicom/src/highdicom/sr/value_types.py", line 139, in append
    super(ContentSequence, self).append(item)
  File "/Users/christopher.bridge/.pyenv/versions/highdicom/lib/python3.8/site-packages/pydicom/sequence.py", line 62, in append
    super().append(val)
  File "/Users/christopher.bridge/.pyenv/versions/highdicom/lib/python3.8/site-packages/pydicom/multival.py", line 66, in append
    self._list.append(self.type_constructor(val))
AttributeError: 'TimePointContext' object has no attribute '_list'
```
 I will keep thinking about it but for now this has me stumped. Any ideas appreciated.

I'm using pydicom version 2.1.2, python 3.8.7, highdicom master branch

hackermd:
@CPBridge there was just a call to `super` missing in the constructor. Created a draft pull request. Will add a couple of unit tests and then publish the PR.

CPBridge:
Ahhh of course

hackermd:
@CPBridge I added tests (and fixed additional bugs). The PR is now ready for review.

PR #47, opened 2021-03-01
Add missing super call to constructor

hackermd:
Addresses #45 

```python
from highdicom.sr.templates import TimePointContext

TimePointContext('whatever')
```


CPBridge (approved):
Looks good

PR #50, opened 2021-03-02
Bug/sc laterality

CPBridge:
Add back laterality as an optional parameter for SCImage in order to avoid the break in backwards compatibility introduced by #49


CPBridge:
Closing due to merge conflict

PR #55, opened 2021-03-21
Prevent invalid segment numbers

CPBridge:
I was recently reading the standard and realised that highdicom is currently too leniant in what it allows for valid segment numbers. The [standard](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.20.2.4.html#sect_C.8.20.2.4) states:

> Segment Number (0062,0004) shall be unique within each Instance, start at a value of 1, and increase monotonically by 1

As a result, if the segmentation instance is to be valid at all points after its construction, the segments must be added in strict ascending order (1, 2, 3, 4, 5, ....).

Currently we enforce that segment numbers are unique and sorted within a given `add_segments` call but all the following invalid possibilities can currently happen:
- Segment numbers that are negative
- Segment numbers equal to zero (this is actually very likely to happen in practice, with python programmers used to zero-based indexing. In fact this is how I noticed this issue and opened pandora's box)
- Segment numbers that are sorted but not consecutive (e.g. `[1, 2, 45, 126]`)
- Segment numbers that over the course of multiple calls to `add_segments` give rise to an unsorted list. E.g. it currently possible to call with a segment number 2 and then later with segment number 1, giving rise to an invalid segmentation instance.

Any one of these would give rise to non-compliant segmentation instances produced by highdicom, so fixing this should be fairly high priority.

In this PR, I have added checks to enforce that all of the above situations are now disallowed and raise errors with hopefully useful error messages. I have added appropriate documentation and tests.

Given that the standard is so prescriptive in how segment numbers are used, I am questioning whether it makes sense to make the user specify the segment numbers manually in the segment descriptions in most situations. I imagine that >95% of uses of the segmentation instance will fall into one of the following two cases:
- A single call to the constructor with single or multiple segments in the pixel array
- A call to the constructor with a single segment, followed by subsequent calls to the `add_segments` method, each with a single additional segment.

In other words, multiple calls to `add_segments` each with multiple segments is a situation that I imagine will only arise very seldomly.

In both of the common cases, having the user specify the segment numbers manually is arguably just boilerplate and a possibility for annoying errors to occur. Therefore in this PR I am suggesting the following: that the segment_number argument to SegmentDescription now accepts `None` value. In this case `add_segments` will automatically determine the segment numbers such that they occur in the correct pattern. In time, I would suggest that the `segment_number` argument is moved to be an optional argument to `SegmentDescription`, but this is a breaking change that we may want to be more careful about.

Alternatively, if you prefer, we have two other options:
- Make `segment_number` an optional parameter in this PR (breaking change)
- Continue to require `segment_number` to be explicitly specified by the user and accept that this is boilerplate in most cases.

Thoughts?

### DIFF
--- a/src/highdicom/seg/content.py
+++ b/src/highdicom/seg/content.py
@@ -23,11 +23,11 @@ class SegmentDescription(Dataset):
 
     def __init__(
             self,
+            segment_number: Optional[int],
             segment_label: str,
             segmented_property_category: Union[Code, CodedConcept],
             segmented_property_type: Union[Code, CodedConcept],
             algorithm_type: Union[SegmentAlgorithmTypeValues, str],
-            segment_number: Optional[int] = None,
             algorithm_identification: Optional[
                 AlgorithmIdentificationSequence
             ] = None,
@@ -43,8 +43,11 @@ class SegmentDescription(Dataset):
         """
         Parameters
         ----------
-        segment_number: int
-            Number of the segment
+        segment_number: Optional[int]
+            Number of the segment. If set to ``None``, no segment number will
+            assigned at this time, and a number will be automatically
+            determined when the description is added to a segmentation
+            instance.
         segment_label: str
             Label of the segment
         segmented_property_category: Union[pydicom.sr.coding.Code, highdicom.sr.coding.CodedConcept]
@@ -79,6 +82,8 @@ class SegmentDescription(Dataset):
         """  # noqa
         super().__init__()
         if segment_number is not None:
+            if segment_number < 1:
+                raise ValueError("Segment number must be a positive integer")
             self.SegmentNumber = segment_number
         self.SegmentLabel = segment_label
         self.SegmentedPropertyCategoryCodeSequence = [
--- a/src/highdicom/seg/content.py
+++ b/src/highdicom/seg/content.py
@@ -23,7 +23,7 @@ class SegmentDescription(Dataset):
 
     def __init__(
             self,
-            segment_number: Optional[int],
+            segment_number: int,
             segment_label: str,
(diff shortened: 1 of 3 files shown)


hackermd:
>> Segment Number (0062,0004) shall be unique within each Instance, start at a value of 1, and increase monotonically by 1

> As a result, if the segmentation instance is to be valid at all points after its construction, the segments must be added in strict ascending order (1, 2, 3, 4, 5, ....).

Great that you caught that!

> Make segment_number an optional parameter in this PR (breaking change)

I would not do that, since it would also require re-ordering the parameters (optional parameters with default argument values cannot be before required parameters).

> Continue to require segment_number to be explicitly specified by the user and accept that this is boilerplate in most cases.

I would prefer this approach. The `SegmentDescription` class is a bit of unusual, since it doesn't map directly onto a DICOM attribute (i.e., contains only a subset of the attributes of *Segment Sequence*). In hindsight, it would have probably been better to not include `segment_number` and just automatically populated the `SegmentNumber` attribute in `Segmentation.add_segments()`. However, I feel that it would be better to stick to the chosen approach at this pointer rather than introducing a breaking change or adding further confusion by making it "optional" (i.e., `Union[int, None]` but not `Optional[int] = None`). Having to explicitly specify the value doesn't do much harm in my opinion and actually ensures that segments are added in the intended order.

hackermd (changes requested):
Let's not make `segment_number` "optional" in the constructor of `SegmentDescription`, but just assert the value of the `SegmentNumber` attribute upon `Segmentation.add_segments()`.

CPBridge:
Ok I have removed the automatic segment numbering. All the checks for incorrect numbering are still in place

hackermd (approved):

PR #57, opened 2021-03-29
Add function to correctly format a DS string

CPBridge:
According to the standard, attributes with the decimal string (DS) value representation must contain a maximum of 16 characters (see [here](http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html)). Unfortunately pydicom does not enforce this, and it looks like there is no movement on @hackermd's [open issue](https://github.com/pydicom/pydicom/issues/1264) to rectify this.

A quick fix was previously implemented in the `PlanePositionSequence`, but this was sub-optimal as it truncated (rather than correctly rounding) and did not handle certain cases correctly (e.g. values that python would write in scientific notation when calling the str() method on a float).

Furthermore, there are several other instances around the codebase where this fix was not implemented, meaning that invalid DS could be created. For example in NumContentItem, which is used for recording nearly all numeric values in an SR document. I found that this was causing many SRs I created to fail the checks in the dciodvfy tool.

This PR implements a new function that I believe correctly handles all edge cases, and rounds properly. It also checks for nans/infs, which would previously be silently stored as strings despite being invalid values for DS. Also added tests and docs.

I have tried to find all the places in the codebase where DS elements are created and switch to using the new function, but I'm not 100% sure I found that I found them all.

### DIFF
--- a/src/highdicom/content.py
+++ b/src/highdicom/content.py
@@ -12,6 +12,7 @@ from highdicom.enum import (
     CoordinateSystemNames,
     UniversalEntityIDTypeValues,
 )
+from highdicom.valuerep import get_ds_string
 from highdicom.sr.coding import CodedConcept
 from highdicom.sr.value_types import (
     CodeContentItem,
@@ -103,10 +104,10 @@ class PixelMeasuresSequence(DataElementSequence):
         """
         super().__init__()
         item = Dataset()
-        item.PixelSpacing = list(pixel_spacing)
-        item.SliceThickness = slice_thickness
+        item.PixelSpacing = [get_ds_string(ps) for ps in pixel_spacing]
+        item.SliceThickness = get_ds_string(slice_thickness)
         if spacing_between_slices is not None:
-            item.SpacingBetweenSlices = spacing_between_slices
+            item.SpacingBetweenSlices = get_ds_string(spacing_between_slices)
         self.append(item)
 
 
@@ -145,9 +146,6 @@ class PlanePositionSequence(DataElementSequence):
         super().__init__()
         item = Dataset()
 
-        def ds(num: float) -> float:
-            return float(str(num)[:16])
-
         coordinate_system = CoordinateSystemNames(coordinate_system)
         if coordinate_system == CoordinateSystemNames.SLIDE:
             if pixel_matrix_position is None:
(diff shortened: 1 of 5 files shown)


src/highdicom/valuerep.py
```diff
@@ -0,0 +1,68 @@
+from math import log10, floor
+
+from numpy import isfinite
+
+
+def get_ds_string(f: float) -> str:
```
pieper, line 6: Wouldn't it make sense for this to be in pydicom instead?

hackermd: @pieper I agree, this would ideally go into pydicom. There is an option issue (see https://github.com/pydicom/pydicom/issues/1264). @CPBridge have you considered creating a PR to update `pydicom.valuerep.DS`?

hackermd (changes requested):
Thanks @CPBridge. This looks great. However, I agree with @pieper that this would ideally go into pydicom. Is this an option?

src/highdicom/content.py
```diff
@@ -261,7 +265,7 @@ def __eq__(self, other: Any) -> bool:
         Returns
         -------
         bool
-            Whether the two image planes have the same orientation
+            Whether the two image planes ha[
```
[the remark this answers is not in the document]
CPBridge: Ooops

CPBridge:
I would certainly agree with @hackermd  and @pieper that this should be fixed at the pydicom level and this was my first thought also, but given this response in the [existing issue](https://github.com/pydicom/pydicom/issues/1264) in pydicom:

> "...pydicom should leave it to the calling code to figure out how to truncate the value."

I got the impression that this would not be welcomed.

However I suppose I night be reading too much into this and any code that does not force pydicom users to truncate their numbers in a certain way might still be welcomed. I could try adding it as a utility function in pydicom that users of pydicom can choose to use explicitly.

Let's park this PR here for now and I will submit a version to pydicom and see where that leads.

hackermd:
> However I suppose I night be reading too much into this and any code that does not force pydicom users to truncate their numbers in a certain way might still be welcomed. I could try adding it as a utility function in pydicom that users of pydicom can choose to use explicitly.

Handling of DS is tricky and can cause really nasty bugs. I understand that we need to workaround compliance issues, but it shouldn't go as far as making it hard for people to actually create valid DICOM in the first place.

An approach as proposed in https://github.com/MGHComputationalPathology/highdicom/issues/56#issuecomment-810679595 may help with that.

CPBridge:
My approach here is to:
- Get the functionality for correctly truncating a float into a 16 char string into pydicom, but have it entirely optional for users of pydicom
- Have users of highdicom continue to pass floats into the highdicom API, and then call the truncation routine from pydicom whenever the attribute is set within highdicom code (the same places I updated in the original PR)

Sound like a good approach?

hackermd:
> Sound like a good approach?

Sounds good to me. @darcymason what do you think? Would that be a good compromise?

darcymason:
> Sounds good to me. @darcymason what do you think? Would that be a good compromise?

Given that it is proposed as an explicit call, I can't see any problem with that.

CPBridge:
See https://github.com/pydicom/pydicom/pull/1334

CPBridge:
This functionality has just been merged into pydicom master branch https://github.com/pydicom/pydicom/pull/1334 and should be in the next release of pydicom (looks like this will be 2.2.0). I will adapt this PR to use that function instead

src/highdicom/valuerep.py
hackermd: @CPBridge given that this has been merged into pydicom, can we remove this module?

tests/test_valuerep.py
```diff
@@ -0,0 +1,36 @@
+import pytest
```
hackermd, line 1: @CPBridge I assume we no longer need these tests in highdicom, do we?

CPBridge: No this PR was originally written with the assumption that the functionality would be in highdicom and I haven't adapted it to at all yet for the fact that the functionality has moved to pydicom. A lot will need to change. I might just create a new PR from scratch, not sure which will end up being easier

src/highdicom/valuerep.py
CPBridge: yes, there's not sense reviewing this PR in it's current form

CPBridge:
Closing in favour of #65

PR #58, opened 2021-03-29
Fixed small encoding errors

seandoyle:
Three small encoding errors - all caught by Dave Clunie's SR validator and checked against DICOM spec.


hackermd (approved):
Thanks @seandoyle

PR #59, opened 2021-03-30
Add required VerifyingObserverIdentificationCodeSequence

CPBridge:
In the SR CIODs we are using, the Verifying Observer Identification Code Sequence is a Type 2 attribute of the Verifying Observer Sequence, meaning it must be present, even if empty ([reference](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.17.2.html#table_C.17-2)).

Currently this is completely omitted. The dciodvfy tool is flagging this as an error for any SRs constructed with a verifying obersver.

This PR simply adds an empty sequence to bring the files into compliance with the standard.

Another option would be to give the user the possibility to pass the information necessary to populate this sequence.

### DIFF
--- a/src/highdicom/sr/sop.py
+++ b/src/highdicom/sr/sop.py
@@ -7,6 +7,7 @@ from typing import Any, Dict, List, Optional, Sequence, Union
 
 from pydicom.sr.coding import Code
 from pydicom.dataset import Dataset
+from pydicom.sequence import Sequence as pdcm_Sequence
 from pydicom.valuerep import DT
 from pydicom._storage_sopclass_uids import (
     ComprehensiveSRStorage,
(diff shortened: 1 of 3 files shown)


hackermd:
> Another option would be to give the user the possibility to pass the information necessary to populate this sequence.

I don't know what that attribute is actually indented for. So I think it's fine to just include an empty sequence for now. We can revisit in case need arises in the future.

src/highdicom/sr/sop.py
```diff
@@ -164,6 +165,8 @@ def __init__(
             observer_item.VerifyingObserverName = verifying_observer_name
             observer_item.VerifyingOrganization = verifying_organization
             observer_item.VerificationDateTime = DT(now)
+            observer_item.VerifyingObserverIdentificationCodeSequence = \
```
hackermd, line 168: Is there an advantage of using `pydicom.sequence` over `[]`?

```python
            observer_item.VerifyingObserverIdentificationCodeSequence = []
```

CPBridge: I don't think so, have changed as suggested

hackermd (approved):

PR #60, opened 2021-03-30
Prevent empty content sequence in Finding Site

CPBridge:
If a FindingSite object is constructed with neither optional parameter specified (`laterality=None` and `topographical_modifier=None`), the item will be constructed with an empty `ContentSequence` (i.e. a sequence of length 0).

However, `ContentSequence` is a type 1C attribute, which I believe means that it cannot be empty. Further more [table C.17-6](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.17.3.html#table_C.17-6) states explicitly that

> One or more Items shall be included in this Sequence

I conclude that the expected behaviour in this situation is that the `ContentSequence` attribute should not be present. I was previously encountering (somewhat crytic) errors with dciodvfy before implementing this fix, and these have now gone.

### DIFF
--- a/src/highdicom/sr/content.py
+++ b/src/highdicom/sr/content.py
@@ -447,29 +447,30 @@ class FindingSite(CodeContentItem):
             value=anatomic_location,
             relationship_type=RelationshipTypeValues.HAS_CONCEPT_MOD
         )
-        self.ContentSequence = ContentSequence()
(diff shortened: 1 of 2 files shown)


hackermd (approved):
Good catch! Thanks @CPBridge

Issue #61, opened 2021-03-30
It is possible to create qualitative evaluations with no relationship types


CPBridge:
I created an SR which contained qualitative evaluations, which are passed in as a sequence of `CodeContentItems` to the `qualitative_evaluations` parameter of the various SRs. I constructed these `CodeContentItems` without the `relationship_type` parameter, and highdicom did not issue any warnings or complaints. However the dciodvfy tool raises an error saying that the `RelationshipType` is required and therefore the SR objects are invalid.

I am not sure exactly what the right level to solve this issue is.

The simplest way is simply to check the elements of `qualitative_evaluations` when constructing `MeasurementsAndQualitativeEvaluations` objects to ensure they have `RelationshipType`s. However after thinking about this more I think that this specific issue may be a symptom of a broader problem. Currently the `relationship_type` is an optional parameter for a `ContentItem` and all its subclasses (e.g. `CodeContentItem`, `TimeContentItem`,`NumContentItem`, ...). I'm not sure that this is correct. I think it should probably be required.

My reading of table [C.17-6](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.17.3.html#table_C.17-6) is that RelationshipType is a type 1 (required) attribute within a `ContentSequence`. I tried making `relationship_type` a required parameter of `ContentItem` and all subclasses, and the only tests that failed were tests where these objects were directly constructed without the relationship type parameter by test code. I.e. there is no place in the main codebase where an item is constructed without a relationship type (or at least not one with test coverage I suppose...). I haven't been able to think of a situation where a relationship type wouldn't be needed. The one place I'm unsure about is the very root of the content tree, which doesn't have a relationship to its parent. I'm not sure how to understand the recursive standard docs on this point. But even there it seems like highdicom is currently placing a `RelationshipType` attribute at the root of the tree anyway

Another option is that maybe the append method of `ContentSequence` should have a check to ensure that the `ContentItem` being appended has a `RelationshipType`. 

@hackermd can you help please me understand at what level this issue exists?

hackermd:
Hmm... That's a tricky one!

> Currently the relationship_type is an optional parameter for a ContentItem and all its subclasses (e.g. CodeContentItem, TimeContentItem,NumContentItem, ...). I'm not sure that this is correct. I think it should probably be required.

[...]

PR #66, opened 2021-04-29
Add fallback for missing ReferringPhysicianName, add qualitative_evaluations to test_sr

fedorov:


fedorov:
On a related note, it took me quite some time to figure out that I need to initialize `RelationshipType` manually (as in https://github.com/MGHComputationalPathology/highdicom/pull/66/files#diff-346e4cd4f40101300bc7800229e1371f1428b2d395ec4207d3ec788e7d4f98e3R1875) when passing qualitative evaluations (as in this example here: https://github.com/MGHComputationalPathology/highdicom/pull/66/files#diff-346e4cd4f40101300bc7800229e1371f1428b2d395ec4207d3ec788e7d4f98e3R1894).

In this particular case, due to the template constraints, the relationship of that `CODE` content item will always be `CONTAINS`. It would make a lot more sense if this relationship was automatically initialized somewhere here: https://github.com/MGHComputationalPathology/highdicom/blob/master/src/highdicom/sr/templates.py#L1360-L1366. Hopefully this addition to the test will make it easier for the next person trying to do what I was doing!

src/highdicom/legacy/sop.py
```diff
@@ -457,7 +457,8 @@ def __init__(
             study_id=ref_ds.StudyID,
             study_date=ref_ds.StudyDate,
             study_time=ref_ds.StudyTime,
-            referring_physician_name=ref_ds.ReferringPhysicianName,
+            referring_physician_name=getattr(
+                ref_ds, "ReferringPhysicianName", None),
```
hackermd, line 461:
```suggestion
            referring_physician_name=getattr(
                ref_ds, 'ReferringPhysicianName', None
            ),
```

src/highdicom/seg/sop.py
```diff
@@ -255,7 +255,8 @@ def __init__(
             study_id=src_img.StudyID,
             study_date=src_img.StudyDate,
             study_time=src_img.StudyTime,
-            referring_physician_name=src_img.ReferringPhysicianName,
+            referring_physician_name=getattr(
+                src_img, "ReferringPhysicianName", None),
```
hackermd, line 259:
```suggestion
            referring_physician_name=getattr(
                src_img, 'ReferringPhysicianName', None
            ),
```

src/highdicom/sr/sop.py
```diff
@@ -134,7 +134,8 @@ def __init__(
             study_id=evidence[0].StudyID,
             study_date=evidence[0].StudyDate,
             study_time=evidence[0].StudyTime,
-            referring_physician_name=evidence[0].ReferringPhysicianName,
+            referring_physician_name=getattr(
+                evidence[0], "ReferringPhysicianName", None),
```
hackermd, line 138:
```suggestion
            referring_physician_name=getattr(
                evidence[0], 'ReferringPhysicianName', None
            ),
```

hackermd (changes requested):
Thanks @fedorov. Good catch! I have only a few minor comments regarding coding style (single quotes instead of double quotes and closing brackets on new line).

hackermd:
Regarding relationship type, see related discussion https://github.com/MGHComputationalPathology/highdicom/issues/61

fedorov:
Thanks Markus - your suggestions applied.

Indeed, I actually even commented in #61, but did not dig into the main topic of the issue. I will follow up in that thread. Thanks for making the reference!

hackermd (approved):

Issue #62, opened 2021-04-02
Should inclusion of referenced images in PertinentOtherEvidenceSequence be enforced?


CPBridge:
It is my understanding that the PertinentOtherEvidenceSequence should contain all instances referenced anywhere in the content tree, and failing to do so would technically be non-compliant. See [sect_C.17.2.3](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.17.2.3.html#sect_C.17.2.3)

> [evidence in PertinentOtherEvidenceSequence] shall include, but is not limited to, all current evidence referenced in the content tree

When constructing an SR object, highdicom has all the available information to ensure that this has been done correctly (i.e. no referenced instance in the content tree has been omitted from the `evidence` parameter), but it currently does not do so.

I would suggest that one improvement could be to add code to the SR constructor that walks the content tree and raises an exception if any evidence is missing. Unfortunately I think it will not be possible to automatically add the evidence as some information (such as series instance uid) may be missing from the references in the content tree.

Thoughts? @hackermd

hackermd:
Good catch!

There are several aspects:

1. Recording all evidence that is referenced in the content tree.
2. Extracting the relevant information automatically from items contained in `ContentSequence` or from the passed `evidence`.
3. Deciding whether the evidence should go into `CurrentRequestedProcedureEvidenceSequence` or `PertinentOtherEvidenceSequence`.

We should be able to determine all evidence based on the value type of content items (`IMAGE`, `COMPOSITE`, etc.) passed via `content`. You are right that the `SeriesInstanceUID` is not contained in the content items, but is required to reference the SOP instances via `CurrentRequestedProcedureEvidenceSequence` or `PertinentOtherEvidenceSequence`. Therefore, the corresponding SOP instances (`pydicom.dataset.Dataset` objects) have to be passed via `evidence`. I agree with you that it would be beneficial to add checks that ensure that's actually the case.

However, we currently also use `evidence` to copy the corresponding study-level metadata. Can we assume that all SOP instances referenced in the content tree are part of the same study? @dclunie could you kindly confirm this contraint?

hackermd:
[...]

PR #67, opened 2021-04-29
Fix referencing of current requested procedure or pertinent other evidence

hackermd:
Closing #62.

Includes any data set provided via `evidence` into *Current Requested Procedure Evidence Sequence* if a given data set is referenced in the document content tree via a content item of value type IMAGE or COMPOSITE. If the data set is not referenced, it will be included in *Pertinent Other Evidence Sequence* if `record_evidence` is set to `True`.


CPBridge:
Haven't done a full review yet but one thing that I immediately noticed is that, unless I've missed something, it is still possible to construct an SR where a particular SOP instance is referenced in the content tree but not included in the evidence parameter. In this case highdicom will still not complain and produce an SR that is technically invalid. I think there needs to be an extra loop that checks whether all `ref_uids` are present in `evidence` and raises an exception of something is missing?

hackermd:
> Haven't done a full review yet but one thing that I immediately noticed is that, unless I've missed something, it is still possible to construct an SR where a particular SOP instance is referenced in the content tree but not included in the evidence parameter. In this case highdicom will still not complain and produce an SR that is technically invalid. I think there needs to be an extra loop that checks whether all ref_uids are present in evidence and raises an exception of something is missing?

Very good point! Will add the check

hackermd:
@CPBridge I added the check. The constructor will now raise a ValueError if any of the referenced data sets is not provided as `evidence`. Please note that I removed the requirement that all referenced data sets are part of the same study. As you pointed out in #62 that is probably too restrictive. I further noticed that the Current Requested Procedure Evidence Sequence, the Pertinent Other Evidence Sequence, and the Predecessor Documents Sequence can contain more than one item, which means that series and instances from multiple studies may be referenced.

src/highdicom/sr/sop.py
```diff
@@ -180,67 +180,168 @@ def __init__(
...
+        ]
+        evd_uids = []
+        ref_collection: Dict[str, List[Dataset]] = defaultdict(list)
+        unref_collection: Dict[str, List[Dataset]] = defaultdict(list)
+        for evd in evidence:
+            if evd.StudyInstanceUID != evidence[0].StudyInstanceUID:
+                raise ValueError(
+                    'Data sets provided as evidence must all belong to the '
+                    'same study.'
+                )
```
CPBridge, line 324: I think you maybe intended to remove this check and exception?

src/highdicom/sr/sop.py
```diff
@@ -180,67 +180,168 @@ def __init__(
...
+        )
+        references += find_content_items(
+            content,
+            value_type=ValueTypeValues.COMPOSITE,
+            recursive=True
+        )
+        ref_uids = [
+            ref.ReferencedSOPSequence[0].ReferencedSOPInstanceUID
+            for ref in references
+        ]
```
CPBridge, line 315: Minor preference probably - but I would tend to make this a set instead of a list since order is unimportant, there may be duplicates references that only need to be in this container once, and it will be used for lots of look-up operations. That would also allow you to use set operations between `ref_uids` and `evd_uids` later, rather than another loop.

src/highdicom/sr/sop.py
```diff
@@ -180,67 +180,168 @@ def __init__(
...
+        ----------
+        evidence: List[pydicom.dataset.Dataset]
+            Metadata of instances that serve as evidence for the SR document
+            content
+        content: pydicom.dataset.Dataset
+            SR document content
+
+        Returns
+        -------
+        Tuple[List[pydicom.dataset.Dataset], List[pydicom.dataset.Dataset]]
```
CPBridge, line 291: As discussed previously, my preferred style is to list elements of the output tuple as individual named outputs as this leads to clearer docs in my opinion. Take it or leave it.

tests/test_sr.py
```diff
@@ -2336,6 +2336,12 @@ def setUp(self):
     def test_sop_class_uid(self):
         assert self._report.SOPClassUID == '1.2.840.10008.5.1.4.1.1.88.33'
 
+    def test_evidence(self):
+        ref_evd_items = self._report.CurrentRequestedProcedureEvidenceSequence
+        assert len(ref_evd_items) == 1
+        with pytest.raises(AttributeError):
+            assert self.PertinentOtherEvidenceSequence
+
```
CPBridge, line 2344: A test for evidence from different studies wouldn't be a bad idea?

src/highdicom/sr/sop.py
```diff
@@ -180,67 +180,168 @@ def __init__(
...
+                )
+            evd_item = Dataset()
+            evd_item.ReferencedSOPClassUID = evd.SOPClassUID
+            evd_item.ReferencedSOPInstanceUID = evd.SOPInstanceUID
+            key = (evd.StudyInstanceUID, evd.SeriesInstanceUID)
+            if evd.SOPInstanceUID in ref_uids:
+                ref_collection[key].append(evd_item)
+            else:
+                unref_collection[key].append(evd_item)
+            evd_uids.append(evd.SOPInstanceUID)
```
CPBridge, line 333: Is it worth checking that the item is not already in the sequence, in case the user passes duplicate items in `evidence`?

CPBridge (changes requested):
A few further comments - nothing blocking merge except the vestigial study UID check. Also there seem to be some mypy failures

hackermd:
Addressed mypy failures with 58351d685307ffb0bebebdd2d1617c88c7f6823a

src/highdicom/sr/sop.py
hackermd: Addressed with 8e6ad6259f7553cbd708bbc392d2b2f41df540ae

src/highdicom/sr/sop.py
hackermd: Addressed with 8e6ad6259f7553cbd708bbc392d2b2f41df540ae

src/highdicom/sr/sop.py
hackermd: I did. Somehow it made it back in. Addressed with 8e6ad6259f7553cbd708bbc392d2b2f41df540ae

src/highdicom/sr/sop.py
hackermd: I agree and I think we should adopt this style of documenting the return value. However, I would like the same style to be consistently applied throughout the library and suggest updating docstrings via a separate PR.

tests/test_sr.py
hackermd: Addressed with dbe4c7707a002b3c9e7698d2e5d509cff33a331f

src/highdicom/sr/sop.py
CPBridge: I did some grepping tosee whether there are in fact any other instances of this in the codebase, and I found a couple. Dumping here for reference:
- `highdicom.seg.content.DimensionIndexSequence.get_index_values`
- `highdicom.spatial.map_pixel_into_coordinate_system` and `highdicom.spatial.map_coordinate_into_pixel_matrix`. These may actually be better left as they are as I imagine the output will typically not be destructered

CPBridge (approved):
Looks good! Just one more flake8 bug to squash

hackermd:
> Looks good! Just one more flake8 bug to squash

Addressed with 7284f0c4f255ad654e232572e36a9013c078dbca

PR #71, opened 2021-05-11
Constructors for ImageRegion

CPBridge:
@hackermd 

This is my working branch for implementing more user-friendly constructors for `ImageRegion`. For now just `ImageRegion` and not `ImageRegion3D` is implemented. This PR is just an example - as discussed we may instead choose a different path such as implementing as sub-classes. However the logic in this branch should be re-usable, it may just end up being refactored

### DIFF
--- a/src/highdicom/sr/content.py
+++ b/src/highdicom/sr/content.py
@@ -468,6 +468,20 @@ class ImageRegion(ScoordContentItem):
             of the source image
             (default: ``highdicom.sr.enum.PixelOriginInterpretationValues.VOLUME``)
 
+        Raises
+        ------
+        ValueError
+            If the shape of the graphic data array is invalid for the graphic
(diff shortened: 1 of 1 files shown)


src/highdicom/sr/content.py
```diff
@@ -500,6 +514,228 @@ def __init__(
...
+            graphic_data=graphic_data,
+            source_image=source_image,
+            pixel_origin_interpretation=pixel_origin_interpretation
+        )
+
+    @classmethod
+    def from_ellipse(
+        cls,
+        xc: float,
+        yc: float,
```
hackermd, line 654: We should avoid `x` and `y` here and use `row` and `col` instead, because the axis of the coordinate system (pixel matrix) are Rows and Columns.

hackermd (changes requested):
Nice! Just a minor comment regarding terminology.

PR #72, opened 2021-05-19
Fix import PR docstrings

CPBridge:
Fixing docstrings and sphinx pages for #70 and setting up this PR to merge into #70. See comments in #70 for why this is necessary. Steps involve:

- Removing all packages whose contents moved "up the tree" from the `package.rst` so their docs are not duplicated.
- Changing all references to these types in the docstrings to use the "higher level" version (with copious `sed -i`) so that hyperlinks in the docs build correctly.

I also fixed an issue in which `highdicom.seg.content` was not included in `highdicom.seg.__init__.py`

### DIFF
--- a/docs/package.rst
+++ b/docs/package.rst
@@ -23,15 +23,6 @@ highdicom.color module
    :undoc-members:
    :show-inheritance:
 
-highdicom.content module
-++++++++++++++++++++++++
-
-.. automodule:: highdicom.content
-   :members:
-   :special-members: __call__
-   :undoc-members:
-   :show-inheritance:
-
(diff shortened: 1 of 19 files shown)


CPBridge:
One further thought - if and when these changes are merged into master, readthedocs will get an update with this new version of the docs with all the new shorter import paths, which will no longer match the latest release on pypi and could cause quite a lot of confusion. We want to consider making a release as soon as this is merged or holding back the readthedocs hook for a while.

hackermd (approved):
Thanks  @CPBridge for going through all of the files and updating the docs!!

PR #74, opened 2021-05-30
Segmentation Parsing

CPBridge:
Implementation of substantial parts of the planned segmentation parsing.

This begins with implementation of `from_dataset`/`from_sequence` classmethods and "getter" properties for several classes:
  * AlgorithmIdentificationSequence
  * PixelMeasuresSequence
  * PlanePositionSequence
  * PlaneOrientationSequence
  * SegmentDescription
  * Segmentation

In order to check each of the above sequences for the required attributes in a less verbose, manual way, I implemented checks that a dataset contains the required attributes from a certain module automatically using the module definitions contained within highdicom. For now this functionality is private and lives in `highdicom._module_utils`.

Beyond this, there are several new methods on the Segmentation class that allow users to get various properties of the segmentation image, search for segments and tracking id/uids using filters, and get pixel arrays. There are currently three different public methods for constructing pixel arrays:
1. `get_pixels_by_source_instance` - This is intended for segmentations derived from (multiple) single-frame instances, where frames are indexed by the SOP Instance UID of the source frame. 
2. `get_pixels_by_source_frame` - This is intended for segmentations derived from multiple-frame intances, where frames are indexed by the the frame number of the source frame.
3. `get_pixels_by_dimension_index_values` - Allows users to index by arbitrary dimension index values. This allows for a more general case where the segmentation may de derived from resampled source images.

Considered out of scope for this PR but planned for future work:
* Lazy decoding of segmentation frames via subclassing ImageFileReader
* `from_dataset` method for DimensionIndexSequence - awaiting changes to the encoding behaviour in light of recent discussions around dimension organization uids.
* Indexing segmentations by spatial locations (without interpolation)

Everything seems to be working fine but this PR is currently provisional until tests are written.

@hackermd  could you please comment on the general approach and the interface before tests are written?


src/highdicom/seg/sop.py
```diff
@@ -944,3 +966,1458 @@ def _encode_pixels(self, planes: np.ndarray) -> bytes:
...
+
+        matches = [
+            desc.segment_number
+            for desc in self.SegmentSequence
+            if all(f(desc) for f in filter_funcs)
+        ]
+
+        return matches
+
+    def search_for_tracking_ids(
```
hackermd, line 1400:
```suggestion
    def get_tracking_ids(
```

src/highdicom/seg/sop.py
```diff
@@ -944,3 +966,1458 @@ def _encode_pixels(self, planes: np.ndarray) -> bytes:
...
+                SegmentAlgorithmTypeValues(desc.algorithm_type) == algo_type
+            )
+
+        return list({
+            desc.tracking_id for desc in self.SegmentSequence
+            if desc.tracking_id is not None and
+            all(f(desc) for f in filter_funcs)
+        })
+
+    def search_for_tracking_uids(
```
hackermd, line 1451:
```suggestion
    def get_tracking_uids(
```

src/highdicom/seg/sop.py
hackermd: I am wondering whether it would make sense to combine `get_tracking_ids()` and `get_tracking_uids()` into one function and return `Tuple[str, str]`. What do you think?

src/highdicom/seg/sop.py
```diff
@@ -944,3 +966,1458 @@ def _encode_pixels(self, planes: np.ndarray) -> bytes:
...
+                pixel_value_map = segment_numbers
+            scaled_array = pixel_array * pixel_value_map.reshape(1, 1, 1, -1)
+
+            # Combine segments by taking maximum down final dimension
+            max_array = scaled_array.max(axis=-1)
+            pixel_array = max_array
+
+        return pixel_array
+
+    def get_source_sop_instances(self) -> List[Tuple[hd_UID, hd_UID, hd_UID]]:
```
hackermd, line 1682: Since the method returns not only the SOP Instance UIDs of source images, I find the name misleading.

```suggestion
    def get_source_instance_uids(self) -> List[Tuple[hd_UID, hd_UID, hd_UID]]:
```

src/highdicom/seg/sop.py
```diff
@@ -944,3 +966,1458 @@ def _encode_pixels(self, planes: np.ndarray) -> bytes:
...
+
+        Returns
+        -------
+        List[pydicom.tag.BaseTag]
+            List of tags used as the default dimension index pointers.
+
+        """
+        return self._dim_ind_pointers[:]
+
+    def dimension_indices_are_unique(
```
hackermd, line 1716: I prefer function names to start with the verb

```suggestion
    def are_dimension_indices_unique(
```

src/highdicom/seg/sop.py
```diff
@@ -944,3 +966,1458 @@ def _encode_pixels(self, planes: np.ndarray) -> bytes:
...
+                    'ignore_spatial_locations' parameter.
+                    """
+                )
+        if not self._single_source_frame_per_seg_frame:
+            raise RuntimeError(
+                'Indexing via source frames is not permissible since some '
+                'frames in the segmentation specify multiple source frames.'
+            )
+
+    def get_pixels_by_source_instance(
```
hackermd, line 1820: I am wondering whether we should introduce two functions to handle the cases of single-frame and multi-frame source images separately.

```python
def get_pixels_by_reference(
    self,
    references: Sequence[Tuple[str, int]]
    ...
) -> np.ndarray
```

where `references` provides SOP Instance UID and Frame Number to uniquely identify each frame (see also [Image SOP Instance Reference Macro Attributes](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_10.3.html#table_10-3)). In case of a series of single-frame source images, the frame number shall always be `1` and in case of a multi-frame source image the SOP Instance UID shall be the same for each frame. We could also support `Sequence[Union[str, Tuple[str, int]]]` to allow omitting the Frame Number in case of a series of single-frame images.

Have you considered this approach?

hackermd (commented):
Looks nice @CPBridge. I have added a few suggestions and comments here and there, but my two main concerns are:

1. **Handling of single-frame and multi-frame source images**: The main question is whether we should try to take a general approach that would work for both cases, or implement specific methods for each case. I am ambivalent and see pros and cons for either approach.
2. **Types of return value of properties**: Several of the properties assume that the values stored on the corresponding attributes have a specific highdicom type (e.g., CodedConcept). The type is enforced by the constructor methods (either via `__init__()` or `from_dataset()`/`from_sequence()`). However, I am wondering whether the properties should ensure that the returned values have the expected type (if it doesn't introduce too much overhead), because it's easy to miss the type casting in the alternative constructor methods and because the objects are mutable. We may also want to revisit the SR pull request.

src/highdicom/seg/sop.py
```diff
@@ -944,3 +966,1458 @@ def _encode_pixels(self, planes: np.ndarray) -> bytes:
...
+    def iter_segments(self):
+        return iter_segments(self)
+
+    @property
+    def number_of_segments(self) -> int:
+        """int: The number of segments in this SEG image."""
+        return len(self.SegmentSequence)
+
+    @property
+    def segment_numbers(self) -> range:
```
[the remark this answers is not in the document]
CPBridge: It's a type
```python
In [1]: print(type(range(10)))
<class 'range'>
```

src/highdicom/seg/sop.py
```diff
@@ -944,3 +966,1458 @@ def _encode_pixels(self, planes: np.ndarray) -> bytes:
...
+        tracking_id: Optional[str] = None,
+    ) -> List[int]:
+        """Get a list of segment numbers matching provided criteria.
+
+        Any number of optional filters may be provided. A segment must match
+        all provided filters to be included in the returned list.
+
+        Parameters
+        ----------
+        segment_label: Optional[str], optional
```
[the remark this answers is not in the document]
CPBridge: Have we decided to keep them in docstrings now?

src/highdicom/content.py
```diff
@@ -73,6 +75,86 @@ def __init__(
...
+    @classmethod
+    def from_sequence(
+        cls,
+        sequence: DataElementSequence
+    ) -> 'AlgorithmIdentificationSequence':
+        """Construct instance from an existing data element sequence.
+
+        Parameters
+        ----------
+        sequence: pydicom.sequence.Sequence
```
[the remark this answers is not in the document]
CPBridge: I don't think this makes sense... The syntax of using square brackets on a type to denote the type of the contained item is not standard python. It is only correct for the "abstract" (I don't think that's the right word but I'm not sure what the right word is) types in the standard library's `typing` module, e.g. `typing.Sequence[pydicom.dataset.Dataset]`. This syntax is not valid python for the `pydicom.sequence.Sequence` and therefore I don't think we should use it in the docstring.

src/highdicom/seg/content.py
```diff
@@ -152,6 +154,146 @@ def __init__(
...
+        """Construct instance from an existing dataset.
+
+        Parameters
+        ----------
+        dataset: pydicom.dataset.Dataset
+            Dataset representing an item of the Segment Sequence.
+
+        Returns
+        -------
+        highdicom.seg.content.SegmentDescription
```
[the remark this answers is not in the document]
CPBridge: No, we should prefer the simplified pattern. In fact I think that is needed for sphinx to understand the types. Resolved in aabfe08f9f239a050a7e645a1793b83b12d41e8d

src/highdicom/seg/content.py
```diff
@@ -152,6 +154,146 @@ def __init__(
...
+        """Optional[highdicom.content.AlgorithmIdentificationSequence]
+            Information useful for identification of the algorithm, if any.
+
+        """
+        if hasattr(self, 'SegmentationAlgorithmIdentificationSequence'):
+            return self.SegmentationAlgorithmIdentificationSequence
+        return None
+
+    @property
+    def tracking_uid(self) -> Optional[str]:
```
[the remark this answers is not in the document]
CPBridge: I don't really agree that this is better, but changed as suggested in d57e950e704fc8faf27a2310a5f655ce5ff86e9a

src/highdicom/seg/sop.py
```diff
@@ -944,3 +966,1458 @@ def _encode_pixels(self, planes: np.ndarray) -> bytes:
...
+
+        """
+        if segment_number < 1 or segment_number > self.number_of_segments:
+            raise IndexError(
+                f'{segment_number} is an invalid segment number for this '
+                'dataset.'
+            )
+        return self.SegmentSequence[segment_number - 1]
+
+    def search_for_segments(
```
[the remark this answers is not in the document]
CPBridge: I don't really have a preference between `find` and `get`, but we should be consistent. Since the SR parsing branch uses get, I changed to `get_segment_numbers`. Resolved in 9f466a42080777e8943d4e00f5f2593a1977adaf

src/highdicom/seg/sop.py
CPBridge: Resolved in 9f466a42080777e8943d4e00f5f2593a1977adaf

src/highdicom/seg/sop.py
CPBridge: Resolved in 9f466a42080777e8943d4e00f5f2593a1977adaf

src/highdicom/seg/sop.py
CPBridge: I actually started down that path and then split into two functions after a while because I think the single-method approach leads to quite a confusing function signature/API. In the case of a multiframe, unless something quite complex is going on that I don't think we need to support, all the frames with have the same UID leading to a lot of redundancy in the input. The two-method approach perhaps leads to a small amount of code redundancy but i think gives a cleaner interface and simpler implementation

src/highdicom/seg/sop.py
CPBridge: Perhaps. The issue with this is what happens if the mapping of tracking ID to tracking UID is not 1-to-1? Would it return all unique pairs? But then some IDs or UIDs would be repeated. Additionally I think most users will be referencing segments by either tracking ID or tracking UID, and not both so will normally throw away half the returned information

src/highdicom/seg/sop.py
CPBridge: Resolved in df7cb4801c8e090870e30e90d29e992e0e62efb3

src/highdicom/seg/sop.py
CPBridge: Changed to `get_tracking_ids` in df7cb4801c8e090870e30e90d29e992e0e62efb3

src/highdicom/seg/content.py
```diff
@@ -152,6 +154,146 @@ def __init__(
...
+    @property
+    def anatomic_regions(self) -> List[CodedConcept]:
+        """List[highdicom.sr.coding.CodedConcept]:
+            List of anatomic regions into which the segment falls.
+            May be empty.
+
+        """
+        if not hasattr(self, 'AnatomicRegionSequence'):
+            return []
+        return self.AnatomicRegionSequence
```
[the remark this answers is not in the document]
CPBridge: We definitely need to be careful about this. However I think this "eager" type conversion is the right approach, and we should enforce via testing rather than runtime conversion in the parsing methods.

The reason is that I think that we should be careful to ensure that `from_dataset` methods convert all objects under their hierarchy to the relevant `highdicom` type where applicable so that objects returned from `from_dataset` are indistinguishable from those created via constructors and thereafter we don't have to worry about the "history" of the object. If we stick to this, it is clear where the burden lies for this type conversion (always in `from_dataset`) and I think this will lead to fewer bugs. I also discuss this on my [comment](https://github.com/MGHComputationalPathology/highdicom/pull/69#pullrequestreview-671889771) on the SR parsing API.

src/highdicom/seg/sop.py
CPBridge: After reading up on the standard and realising that there should be a one-to-one mapping between tracking id and tracking uid, I changed my mind and decided to merge into a single function as suggested

CPBridge:
@hackermd I have added full tests and this is ready for a full review

CPBridge:
I'm not sure why the travis pipeline isn't running, and I can't see how to trigger it

src/highdicom/content.py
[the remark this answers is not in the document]
hackermd: Agreed.

src/highdicom/seg/content.py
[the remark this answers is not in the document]
hackermd: Fully agreed. We should probably document this (and other fundamental design decision) in the [developer guide](https://highdicom.readthedocs.io/en/latest/development.html). What do you think?

src/highdicom/seg/sop.py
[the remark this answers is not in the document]
hackermd: Interesting! Thanks for clarifying

src/highdicom/seg/sop.py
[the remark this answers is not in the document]
hackermd: If I recall correctly, we decided to use `Optional[str] = None` for the type annotation and `Union[str, None], optional` for the docstrings.

src/highdicom/seg/sop.py
```diff
@@ -940,3 +962,1491 @@ def _encode_pixels(self, planes: np.ndarray) -> bytes:
...
+            All unique segmented property categories referenced in segment
+            descriptions in this SEG image.
+
+        """
+        categories = []
+        for desc in self.SegmentSequence:
+            if desc.segmented_property_category not in categories:
+                categories.append(desc.segmented_property_category)
+
+        return categories
```
hackermd, line 1486: Do we want to rewrite this into a list comprehension for (potentially) improved performance?

```suggestion
        return [
            desc.segmented_property_category
            for desc in self.SegmentSequence
            if desc.segmented_property_category not in categories
        ]
```

src/highdicom/seg/sop.py
```diff
@@ -940,3 +962,1491 @@ def _encode_pixels(self, planes: np.ndarray) -> bytes:
...
+            All unique segmented property types referenced in segment
+            descriptions in this SEG image.
+
+        """
+        types = []
+        for desc in self.SegmentSequence:
+            if desc.segmented_property_type not in types:
+                types.append(desc.segmented_property_type)
+
+        return types
```
hackermd, line 1504:
```suggestion
        return [
            desc.segmented_property_type
            for desc in self.SegmentSequence
            if desc.segmented_property_type not in types
        ]
```

hackermd (approved):
Very nice @CPBridge! I only have a couple of suggestions for your consideration, but am also fine with merging as is.

src/highdicom/seg/sop.py
CPBridge: This can't be written as a list comprehension because the check is applied to the list that is being constructed. It appends the segmented property to the list only if it is not already in the list.

(It is also not possible to use a set here because CodedConcept is not hashable).

I think we are stuck with it the way it is, even if it's a bit ugly.

src/highdicom/seg/sop.py
CPBridge: This can't be written as a list comprehension because the check is applied to the list that is being constructed. It appends the segmented property to the list only if it is not already in the list.

src/highdicom/seg/sop.py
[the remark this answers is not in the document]
CPBridge: Addressed in 9f449dc

src/highdicom/seg/content.py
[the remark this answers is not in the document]
CPBridge: See #88

CPBridge:
@hackermd a few changes since the last review:

- Added more docstring examples for the more complicated methods
- Removed `Optional` from the docstrings as requested, in favour of `Union[x, None]`
- Added a new `segread` function (analogous to `dcmread`) to read from a file into a segmentation object in a single line of code. I wanted to put this in `hd.seg.utils` but couldn't because of circular imports, so put in `hd.seg.sop` instead. We may want to add `srread` similarly?

src/highdicom/seg/sop.py
hackermd: > This can't be written as a list comprehension because the check is applied to the list that is being constructed. It appends the segmented property to the list only if it is not already in the list.

Right...

> It is also not possible to use a set here because CodedConcept is not hashable

This is something we should consider adding. We may also want to add `__hash__`to `pydicom.sr.coding.Code` so that we can use instances as dictionary keys..

src/highdicom/seg/sop.py
hackermd: Sorry...didn't look carefully enough at it.

hackermd:
Thanks @CPBridge. I would suggest we merge this into master, but wait for #69 before drafting a new release.

CPBridge:
I agree on waiting for #69 to release. Feel free to merge this to master now if you like, or we can try to combine this with #69 in a dev branch first, up to you

CPBridge:
Actually, since it seems like there are still a few potentially gnarly things to resolve on #69, I think I recommend holding this here until we finalize both PRs. We may make some decisions on #69 with broad implications for this PR too, and we should aim for consistency of approach between the two.

CPBridge:
Merging into dev branch to continue refinement there alongside #69

PR #77, opened 2021-06-02
Feat/image library entries

seandoyle:
Small changes for ImageLibrary entries. The current output passes the pixelmed validator. 

TBD - adding modality specific descriptors in Image Library Entries (TID 1602)


hackermd (approved):
Thanks @seandoyle. I defer to your judgment of whether and how we should implement modality-specific descriptors. However, that's something we can address with a separate PR.

hackermd (changes requested):
I just noticed that the tests fail due to style and type issues (https://travis-ci.org/github/MGHComputationalPathology/highdicom/jobs/773313211). @seandoyle you kindly fix those?

seandoyle:
Yes - I'll address those. I also need to add some unit tests.

seandoyle:
The style issues are now fixed but there are other errors in templates.py which I believe are not related to my changes. Should I look at these as well? 

BTW: 

The DICOM SR validator needs a lot of resources. `java -Xmx6g -Xms5g -XX:-UseGCOverheadLimit -cp "${PIXELMEDDIR}/pixelmed.jar:${PIXELMEDDIR}/lib/additional/commons-compress-1.12.jar:${PIXELMEDDIR}/lib/additional/commons-codec-1.3.jar" com.pixelmed.validate.DicomSRValidator $*` - 6GB is a lot of memory.

hackermd (approved):
Thanks @seandoyle. We can address the other typing errors in the target branch.

PR #80, opened 2021-06-04
Add QualitativeEvaluation, fix relationship types

### LINKED ISSUE #61: It is possible to create qualitative evaluations with no relationship types

CPBridge:
Resolves #61 as discussed on that thread by:

- Creating a `highdicom.sr.content.QualitativeEvaluation` as a subclass of `CodeContentItem` that correctly establishes the relationship type of `CONTAINS` (please confirm that's correct) and is otherwise identical to `CodeContentItem`
- Allow users to pass `QualitativeEvaluations` to the `XMeasurementsAndQualitativeEvaluations` constructors. Plain `CodeContentItems` are also acceptable for backwards compatibility but their relationship type is now checked and enforced.
- Add a deprecation warning that creating any `ContentItem` without a relationship type is deprecated behaviour.
- Fix all tests to remove the deprecation warning introduced above. This was mostly just changing the test code itself to pass relationship type parameters when constructing items, however there was one instance in the package code: the `ContainerContentItem` at the root of the `MeasurementReport`. I'm a bit unclear about what should happen here, since it's at the root of the content tree. I added a `CONTAINS` relationship type. Please confirm that this is correct.


src/highdicom/sr/templates.py
```diff
@@ -1227,7 +1228,9 @@ def __init__(
         finding_sites: Optional[Sequence[FindingSite]] = None,
         session: Optional[str] = None,
         measurements: Sequence[Measurement] = None,
-        qualitative_evaluations: Optional[Sequence[CodeContentItem]] = None
+        qualitative_evaluations: Optional[
+            Sequence[Union[CodeContentItem, QualitativeEvaluation]]
```
hackermd, line 1232: A `QualitativeEvaluation` is still a `CodeContentItem`. Therefore, we technically wouldn't need the `Union`.

src/highdicom/sr/templates.py
```diff
@@ -1254,7 +1257,7 @@ def __init__(
             Description of the session
         measurements: Sequence[highdicom.sr.Measurement], optional
             Numeric measurements
-        qualitative_evaluations: Sequence[highdicom.sr.CodeContentItem], optional
+        qualitative_evaluations: Sequence[Union[highdicom.sr.CodeContentItem, highdicom.sr.QualitativeEvaluation]], optional
             Coded name-value pairs that describe measurements in qualitative
             terms
```
hackermd, line 1262: A measurement (discrete or continuous numeric feature) and a qualitative evaluation (categorical feature) are conceptually two different things

```suggestion
            Coded name-value pairs that describe qualitative evaluations
```

src/highdicom/sr/templates.py
```diff
@@ -1362,7 +1365,15 @@ def __init__(
                 if not isinstance(evaluation, CodeContentItem):
                     raise TypeError(
                         'Items of argument "qualitative_evaluations" must have '
-                        'type CodeContentItem.'
+                        'type CodeContentItem or QualitativeEvaluation.'
+                    )
+                rel_type = evaluation.relationship_type
+                expected_rel_type = RelationshipTypeValues.CONTAINS
+                if rel_type is None or \
+                   RelationshipTypeValues(rel_type) != expected_rel_type:
```
hackermd, line 1373: This shouldn't be necessary, since `evaluation.relationship_type` should already return a value of type `RelationshipTypeValues`

```suggestion
                if rel_type is None or rel_type != expected_rel_type:
```

src/highdicom/sr/templates.py
```diff
@@ -1426,7 +1439,7 @@ def __init__(
             description of the session
         measurements: Sequence[highdicom.sr.Measurement], optional
             numeric measurements
-        qualitative_evaluations: Sequence[highdicom.sr.CodeContentItem], optional
+        qualitative_evaluations: Sequence[Union[highdicom.sr.CodeContentItem, highdicom.sr.QualitativeEvaluation]], optional
             coded name-value (question-answer) pairs that describe the
             measurements in qualitative terms
```
hackermd, line 1444:
```suggestion
            coded name-value (question-answer) pairs that describe
            qualitative evaluations
```

src/highdicom/sr/templates.py
```diff
@@ -1570,7 +1585,7 @@ def __init__(
             description of the session
         measurements: Sequence[highdicom.sr.Measurement], optional
             measurements for a region of interest
-        qualitative_evaluations: Sequence[highdicom.sr.CodeContentItem], optional
+        qualitative_evaluations: Sequence[Union[highdicom.sr.CodeContentItem, highdicom.sr.QualitativeEvaluation]], optional
             coded name-value (question-answer) pairs that describe the
             measurements in qualitative terms for a region of interest
```
hackermd, line 1590:
```suggestion
            coded name-value (question-answer) pairs that describe
            qualitative evaluations of a region of interest
```

src/highdicom/sr/templates.py
```diff
@@ -1677,7 +1694,7 @@ def __init__(
             description of the session
         measurements: Sequence[highdicom.sr.Measurement], optional
             measurements for a volume of interest
-        qualitative_evaluations: Sequence[highdicom.sr.CodeContentItem], optional
+        qualitative_evaluations: Sequence[Union[highdicom.sr.CodeContentItem, highdicom.sr.QualitativeEvaluation]], optional
             coded name-value (question-answer) pairs that describe the
             measurements in qualitative terms for a volume of interest
```
hackermd, line 1699:
```suggestion
            coded name-value (question-answer) pairs that describe
            qualitative evaluations of a volume of interest
```

src/highdicom/sr/templates.py
```diff
@@ -1837,7 +1854,8 @@ def __init__(
             )
         item = ContainerContentItem(
             name=title,
-            template_id='1500'
+            template_id='1500',
+            relationship_type=RelationshipTypeValues.CONTAINS
```
hackermd, line 1858: I am not sure if this is correct in case of a root template, which per definition doesn't have a parent.

src/highdicom/sr/value_types.py
```diff
@@ -55,7 +56,14 @@ def __init__(
...
         self.ConceptNameCodeSequence = [name]
-        if relationship_type is not None:
+        if relationship_type is None:
+            warnings.warn(
+                'Constructing a content item without specifying a '
+                'relationship type is deprecated behaviour and will be '
+                'removed in a future version of highdicom.',
+                DeprecationWarning
+            )
+        else:
```
hackermd, line 66: I am not sure whether we can do this. As far as I understand it, root content items of value type CONTAINER may not have a Relationship Type (e.g., [root content item of root template TID 1500](http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#para_85dac8a4-19f9-4283-bd59-d51aa905412d)).

Sections [SR Document Content Tree](dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.17.3.html) and [Content Sequence and Relationship Type](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.17.3.2.4.html) are not explicit in this regard.

However, since [Relationship Type](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.17.3.html#para_f4165794-8736-4c66-84f0-39f64a7fc05d) is an attribute of [Content Sequence](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.17.3.html#para_4f32997d-7567-4a1c-b2d1-ce8fc6795299) and the root content item is not contained in a Content Sequence, but the SR Document itself, it cannot have a Relationship Type.

hackermd: We could add this check to all classes derived from `ContentItem` but `ContainerContentItem`.

hackermd (changes requested):
Thanks @CPBridge. Looks great! I have a few minor comments/suggestions regarding docstrings and type annotations and one major comment regarding the question whether `relationship_type` should be an optional parameter.

src/highdicom/sr/templates.py
CPBridge: I thought so too but in fact it returns a `str`... bit of a mistake in my opinion as I would definitely expect a `RelationshipTypeValues` in line with other properties around the library. We could fix in a non backwards compatible way?

src/highdicom/sr/value_types.py
CPBridge: I was very unsure about this, what you are saying certainly makes intuitive sense to me. I was basing this on @dclunie 's comment in #61 in which he said:

> I dont have suggestions about the API as long as when serialized a value for Relationship Type is always supplied, so that invalid objects are not be created ... see http://dicom.nema.org/medical/dicom/current/output/chtml/part03> /sect_C.17.3.2.4.html <http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.17.3.2.4.html>
> I.e., at some point it needs to be required, not optional.

Suggesting that relationship type is always required, although he doesn't say anything explicitly about the root. Perhaps he would be able to clarify here whether the root content item is an exception to the rule that a relationship type is always required.

> We could add this check to all classes derived from `ContentItem` but `ContainerContentItem`.

@hackermd on the assumption that the root content item is the one exception to the rule, I think this approach, while a little duplicative, is the right one.

src/highdicom/sr/templates.py
hackermd: I agree. That's a good idea. It may be slightly confusing if the type annotations differ between the function signature and the docstring, but we can deprecate support of `CodeContentItem` in a future version and only allow `QualitativeEvaluation`. What do you think about adding a deprecation warning?

src/highdicom/sr/templates.py
hackermd: Oh..that's bad. I would consider this a bug. The property should return a `RelationshipTypeValues` instance, similar to the other properties. Is this something we should fix?

src/highdicom/sr/value_types.py
hackermd: > @hackermd on the assumption that the root content item is the one exception to the rule, I think this approach, while a little duplicative, is the right one.

The problem is that when we include the `RelationshipType` attribute into the document, the object may not be compliant with the standard and may fail validation tests. What do you think about making the `relationship_type` parameter required for all classes derived from `ContentItem`, but making it "conditionally required" for `ContainerContentItem`? We could add an `is_root: bool` parameter to `ContainerContentItem` and raise an exception if `is_root == False` and `relationship_type is None`. This would allow the `relationship_type` parameter to remain optional on the `ContentItem` base class.

fedorov: @dclunie SR book to the rescue!

image

fedorov: I think it would make a lot of sense if RelationshipType was "1C IFF not Root content item" here: http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.17.3.html#sect_C.17.3.1.

hackermd: > I think it would make a lot of sense if RelationshipType was "1C IFF not Root content item"

The conceptual hurdle is that the Relationship Type attribute is defined in [Document Relationship Macro Attributes](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.17.3.html#para_4f32997d-7567-4a1c-b2d1-ce8fc6795299) under Content Sequence, while all the other attributes are defined the [Document Content Macro Attributes](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.17.3.html#table_C.17-5). The root content item is not contained in a Content Sequence, therefore attribute type 1 is correct in this context I guess.

dclunie: Yes ... the Document Relationship Macro is designed that way for that reason, and no condition is necessary.

src/highdicom/sr/templates.py
CPBridge: Ok implemented as suggested in 217b4e0 :
- Type annotations just say `CodeContentItem`
- Docstrings just say `QualitativeEvaluation`
- Deprecation warning is raised if you pass a `CodeContentItem`

src/highdicom/sr/templates.py
CPBridge: I think so, I have fixed this (and the `value_type` property) in 259ae46. This is not backwards compatible, but I doubt much if any user code is using these properties directly. We should probably add something in the release notes though

src/highdicom/sr/value_types.py
CPBridge: Thank you @fedorov  and @dclunie  for the clarification!

> What do you think about making the relationship_type parameter required for all classes derived from ContentItem, but making it "conditionally required" for ContainerContentItem? We could add an is_root: bool parameter to ContainerContentItem and raise an exception if is_root == False and relationship_type is None. This would allow the relationship_type parameter to remain optional on the ContentItem base class.

This definitely seems like the right approach and I was thinking about something very similar myself. However, as I'm trying to implement it, I have hit upon a rather nasty problem: changing the relationship type parameters of the subclasses from optional to required attributes in many cases means having to move their positions in the function signatures ahead of other optional attributes (since in python, optional parameters must follow required attributes). This is especially unfortunate given that the `relationship_type` parameters have been placed at the end of the signature by convention. For example, consider the constructor of `NumContentItem`:

```python
class NumContentItem(ContentItem):

    """DICOM SR document content item for value type NUM."""

    def __init__(
        self,
        name: Union[Code, CodedConcept],
        value: Optional[Union[int, float]] = None,
        unit: Optional[Union[Code, CodedConcept]] = None,
        qualifier: Optional[Union[Code, CodedConcept]] = None,
        relationship_type: Optional[Union[str, RelationshipTypeValues]] = None,  # UH OH
    ) -> None:
```

This would be a potentially nasty breaking change (although I doubt that many or even any users are directly interacting with these content item types). I don't see any other good option though, apart from leave it as an optional parameter but have it raise an error if the caller actually leaves it at the default value, which seems very awkward also.

Preference for how to proceed @hackermd ?

src/highdicom/sr/templates.py
CPBridge: See discussion [below](https://github.com/MGHComputationalPathology/highdicom/pull/80#discussion_r645236633)

src/highdicom/sr/value_types.py
hackermd: Hmm...that's indeed a tricky one. Maybe leaving it as is and asserting the presence of the Relationship Type attribute only when the item gets added to Content Sequence? Since all classes derived from `Template` inherit from `ContentSequence`, we could add the check to [ContentSequence.append()](https://github.com/MGHComputationalPathology/highdicom/blob/1a904e58bc3685eef190e626a0052a21831d8168/src/highdicom/sr/value_types.py#L125). It will require a bit more discipline for the nested content items, but I feel this would still be cleanest option at this point (and it would be in agreement with how it is specified in the standard). We could also add a deprecation warning to keep the option of refactoring the content item interface for the 1.0 release. @CPBridge what do you think?

CPBridge:
Ok @hackermd, I think we might finally have arrived at something workable!

I have implemented as you suggested above:
- Relationship types are still optional for content items, and their function signatures have not changed. However constructed a content item (except a container content item) without a relationship type will now raise a deprecation warning.
- I have added checks to the constructor, `append`, `extend` and `insert` methods of `ContentSequence` that enforces the requirement for all items to have relationship types. Unless I have missed something, this is sufficient to ensure that all DICOM objects produced by the library have relationships established in all content items (except the root, as below).
- To make the above work, there is now a new parameter of the `ContentSequence` constructor called `is_root` that tells the class that it is intended to be used as the root content item of an SR document tree. This reverses all the above checks - i.e. now items passed in the constructor, or appended/extended/inserted must have *no* relationship type.
- The constructor of `MeasurementReport` now passes `is_root=True` to the superclass constructor. From my understanding this is the only template in the library that may be used at the root, but I'm not sure about this. Please confirm? We may need to add this to other templates otherwise.
- Added very simple unit tests for all the above

hackermd (changes requested):
I like the idea of adding a deprecation warning and updating docstring and type annotations to better reflect the future interface. However, I tend argue that the current implementation is not correct.

hackermd (approved):

PR #82, opened 2021-06-07
Compute plane positions only once

hackermd:
Compute plane positions for frames in source image(s) only once upon construction of the Segmentation image rather than upon every `add_segments()`. This should improve performance when many segments are added.


cgorman:
Looks good to me and there's a huge speedup, thanks!

cgorman (approved):

PR #83, opened 2021-06-13
Feat/add image library tid1601

seandoyle:
Added support for Image Library [TID 1601](http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1601) , Image LibraryEntryDescriptors  [TID 1602](http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1602), Image Library Entry Descriptors for Cross-Sectional Modalities [TID 1604](http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1604) and partial support for Image Library Entry Descriptors for Projection Radiography  [TID 1603](http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1603).


CPBridge:
Hi @seandoyle,

In all things SR defer to @hackermd  over me as I’m still getting my head around these things. However I have had a look at this and made some observations that may help.

Firstly I believe that the concept name being blank in the table 1601 means that there should be no concept name for this content item. I believe that IMAGE content items are not required to have concept names, as shown by the dashed border on page 88 of the [SR book](http://www.dclunie.com/pixelmed/DICOMSR.book.pdf). The same is true of CONTAINER, COMPOSITE and WAVEFORM, and SCOORD (and presumably SCOORD3D) content items. Page 71 notes that there are “a few exceptions” to the requirement that content items have concept names, and this appears to be one of them. Note that there are a few other examples of IMAGE content items not having concept names within the defined templates, for example there are several in [TID 1400](http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1400) . This also makes intuitive sense to me that the meaning of the image within the library is obvious and does not need to be further explained with a concept name.

Now, unfortunately the highdicom API currently *does* require a concept name when constructing an ImageContentItem object, so my reading of the situation is that this constraint will have to be relaxed, e.g. by allowing the user to explicitly pass `name=None` to the ImageContentItem constructor.

Secondly, you asked about the relationship type of the first row of TIB 1601 being missing. This is a common pattern used in the SR template tables and simply means that the relationship is defined where the template is "invoked", if that is the right word. In this case, that means row 4 of table [TID 1600](http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1600) in which it is specified as CONTAINS.

Thirdly, I believe that you have misunderstood the nesting structure of TID 1600 in your current implementation, and this may be what is causing the issues you are seeing with the validator. I believe that this is how this is supposed to work:
-	An imaging library contains 1-n Image Library Groups
-	An Image Library Group contains 1-n Image Library Entries and/or 1-n Image Library Entry Descriptors
-	An Image Library Entry contains 1-n Image Library Entry Descriptors

Note that the descriptors can appear in two different places: at the level of a group, in which case they describe all the images in the group, and/or at the level of an entry, in which case they describe a single image.

Given that your current implementation loops over the datasets individually, I guess that you are aiming for the second of these options, i.e. all images are described separately. However, you are appending the library_item_entry to the group item’s content sequence, which suggests the former. To achieve what I think you are trying to do, I think what you need to do is append the image content item representing the Image Library Entry to the group item’s content sequence, then append the descriptors to the image content item’s content sequence. I.e:


```python
class ImageLibrary(Template):

    """`TID 1600 <http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1600>`_
     Image Library"""  # noqa: E501

    def __init__(
        self,
        datasets: Sequence[Dataset]
    ) -> None:
        """
        Parameters
        ----------
        groups: Sequence[Sequence[highdicom.sr.ImageLibraryEntryDescriptros]]
            Entry descriptors for each image library group
        """
        super().__init__()

        library_item = ContainerContentItem(
            name=CodedConcept(
                value='111028',
                meaning='Image Library',
                scheme_designator='DCM'
            ),
            relationship_type=RelationshipTypeValues.CONTAINS
        )
        library_item.ContentSequence = ContentSequence()
        if datasets is not None:
            # We will use just one group
            group_item = ContainerContentItem(
                name=CodedConcept(
                    value='126200',
                    meaning='Image Library Group',
                    scheme_designator='DCM'
                ),
                relationship_type=RelationshipTypeValues.CONTAINS
            )
            group_item.ContentSequence = ContentSequence()
            for dataset in datasets:
                # Create TID 1601 Image Library Entry
                # represents a single image
                entry = ImageContentItem(
                    name=None,  # UH OH won't work for now!
                    referenced_sop_instance_uid=dataset.SOPInstanceUID,
                    referenced_sop_class_uid=dataset.SOPClassUID
                )

                # Add descriptors nested under the entry
                descriptors = ImageLibraryEntryDescriptors(dataset)
                entry.ContentSequence = ContentSequence()
                entry.ContentSequence.extend(descriptors)

                # Add the entry to the group
                group_item.ContentSequence.append(entry)

            # Add the group to the library
            library_item.ContentSequence.append(group_item)

        self.append(library_item)

```

Note that I have made no attempt to check that this works at all! I also assume a couple of things here that you should check:
1. You only want a single group within the library
2. You don't want to make any attempt to group attributes together if they are shared by all datasets.

As a final thought, we may want to consider creating a new python class for TID 1601 Image Library Entry. This would be somewhat more consistent with the rest of the API, and therefore clearer (even if a bit less space efficient), and also may help if we add parsing functionality to the image library templates in the future, following the pattern we have employed so far in #69 .

hackermd:
> Now, unfortunately the highdicom API currently does require a concept name when constructing an ImageContentItem object, so my reading of the situation is that this constraint will have to be relaxed, e.g. by allowing the user to explicitly pass name=None to the ImageContentItem constructor.

Oh boy...I would not change the `name` parameter of content items and keep it required. We can use a generic name for edge cases where a name is not required by the template.

The standard explicitly allows that (see [PS3.3 C.17.3 SR Document Content Module](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.17.3.html#sect_C.17.3.1)):

> Though many Templates in PS3.16 do not require that the Purpose of Reference be conveyed in the Concept Name, a generic Concept Name, such as (260753009, SCT, "Source"), may be used, since anonymous (unnamed) Content Items may be undesirable for some implementations (e.g., for which the name of a name-value pair is required).

hackermd:
> As a final thought, we may want to consider creating a new python class for TID 1601 Image Library Entry. This would be somewhat more consistent with the rest of the API, and therefore clearer (even if a bit less space efficient), and also may help if we add parsing functionality to the image library templates in the future

Agreed that implementing a class would be clearer. However, we would only be using the class under the hood and would not require users to construct such objects. Therefore, it would not consider it a necessity at this point.

src/highdicom/sr/templates.py
```diff
@@ -3424,14 +3683,21 @@ def __init__(
...
+                        value='121112',
+                        meaning='Source of Measurement',
+                        scheme_designator='DCM'
+                    ),
+                    referenced_sop_instance_uid=dataset.SOPInstanceUID,
+                    referenced_sop_class_uid=dataset.SOPClassUID
+                )
+
+                """`TID 1601 <http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1601>`_
+                     Image Library Entry"""  # noqa: E501
```
hackermd, line 3700: The docstring seems to be misplaced.

src/highdicom/sr/templates.py
```diff
@@ -3388,14 +3613,46 @@ def get_image_measurments(
...
+        group_item = ContainerContentItem(
+            name=CodedConcept(
+                value='126200',
+                meaning='Image Library Group',
+                scheme_designator='DCM'
+            ),
+            relationship_type=RelationshipTypeValues.CONTAINS
+        )
+
+        # $$$
```
hackermd, line 3643: ???

src/highdicom/sr/templates.py
```diff
@@ -3424,14 +3683,21 @@ def __init__(
...
-                if not isinstance(descriptor_items,
-                                  ImageLibraryEntryDescriptors):
-                    raise TypeError(
-                        'Image library group items must have type '
-                        '"ImageLibraryEntry".'
-                    )
+                group_item.ContentSequence = ContentSequence()
+                group_item.ContentSequence.extend(library_item_entry)
+
+                image = ImageContentItem(
```
hackermd, line 3689: I would append the name with `_item` for consistency

```suggestion
                image_item = ImageContentItem(
```

src/highdicom/sr/utils.py
```diff
@@ -174,3 +175,35 @@ def get_coded_value(item: Dataset) -> CodedConcept:
...
     )
+
+
+def get_coded_modality(item: Dataset) -> str:
+    """
+    Gets the coded value of the modality from the dataset's SOPClassUID. The
+    SOPClassUIDs are defined here:
+    http://dicom.nema.org/dicom/2013/output/chtml/part04/sect_B.5.html
+    and the coded values are described here:
+    http://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_CID_29.html
```
hackermd, line 186: Could you please format this part of the docstring according to reStructuredText and the NumPy docstring convention?

src/highdicom/sr/utils.py
```diff
@@ -174,3 +175,35 @@ def get_coded_value(item: Dataset) -> CodedConcept:
...
+
+    Parameters
+    ----------
+    item: pydicom.dataset.Dataset
+        Content Item
+
+    Returns
+    -------
+    codes.cid29 Acquisition Modality
+        str
```
hackermd, line 196:
```suggestion
    pydicom.sr.coding.Code
        Coded Acquisition Modality
```

src/highdicom/sr/utils.py
```diff
@@ -174,3 +175,35 @@ def get_coded_value(item: Dataset) -> CodedConcept:
         meaning=value.CodeMeaning,
         scheme_version=value.get('CodingSchemeVersion', None)
     )
+
+
+def get_coded_modality(item: Dataset) -> str:
```
hackermd, line 180:
```suggestion
def get_coded_modality(item: Dataset) -> Code:
```

hackermd, line 180: I am wondering whether the full dataset should be passed or just SOP Class UID value. Thoughts?

hackermd (changes requested):

src/highdicom/sr/utils.py
```diff
@@ -177,13 +177,13 @@ def get_coded_value(item: Dataset) -> CodedConcept:
     )
 
 
-def get_coded_modality(item: Dataset) -> str:
+def get_coded_modality(sop_class_uid: str) -> Code:
```
hackermd, line 180: It looks like this is limited to datasets representing images. That should be mentioned in the docstring.

src/highdicom/sr/utils.py
```diff
@@ -192,18 +192,152 @@ def get_coded_modality(item: Dataset) -> str:
...
     }
-    if item.SOPClassUID in sopclass_to_modalty_map.keys():
-        return sopclass_to_modalty_map[item.SOPClassUID]
+    if sop_class_uid in sopclass_to_modalty_map.keys():
+        return sopclass_to_modalty_map[sop_class_uid]
     else:
         return None
+
+
+def is_dicom_image(sop_class_uid: str) -> bool:
```
hackermd, line 269: As mentioned in a previous comment, this should be determined based on attributes that describe the pixel data rather than based on the SOP Class UID.

```suggestion
def is_image(dataset: Dataset) -> bool:
```

src/highdicom/sr/utils.py
```diff
@@ -174,3 +175,169 @@ def get_coded_value(item: Dataset) -> CodedConcept:
         meaning=value.CodeMeaning,
         scheme_version=value.get('CodingSchemeVersion', None)
     )
+
+
+def get_coded_modality(sop_class_uid: str) -> Code:
```
hackermd, line 180: I have previously suggested to only pass the SOP Class UID rather than the full dataset (see https://github.com/MGHComputationalPathology/highdicom/pull/83#discussion_r652170213). That may be not such a good idea after all. There may be situation where we need more attributes than the SOP Class UID to determine the modality. For example, the modality of a Parametric Map depends on the modality of the source images.

hackermd: However, I think you intend this function to be used primarily to determine the modality of referenced images (i.e., using [Common Instance Reference Module Attributes](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.12.2.html#table_C.12-8) in which case this wouldn't work either. Maybe it would be better to keep this function private in `highdicom.sr.templates` for now..

src/highdicom/sr/templates.py
```diff
@@ -3388,23 +3623,54 @@ def get_image_measurments(
         return sequences
 
 
-class ImageLibrary(Template):
+class ImageLibraryEntry(Template):
 
+    """`TID 1601 <http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1601>`_
+     Image Library Entry"""  # noqa: E501
```
hackermd, line 3629:
```suggestion
    """`TID 1601 <http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1601>`_
    Image Library Entry
    """  # noqa: E501
```

hackermd (changes requested):

src/highdicom/sr/utils.py
seandoyle: Done. Thanks.

src/highdicom/sr/utils.py
seandoyle: Done.

src/highdicom/sr/utils.py
```diff
@@ -192,18 +192,152 @@ def get_coded_modality(item: Dataset) -> str:
...
         '1.2.840.10008.5.1.4.1.1.4': codes.cid29.MagneticResonance,
-        '1.2.840.10008.5.1.4.1.1.77.1.2': codes.cid29.SlideMicroscopy,
-        '1.2.840.10008.5.1.4.1.1.77.1.6': codes.cid29.SlideMicroscopy
+        '1.2.840.10008.5.1.4.1.1.4.1': codes.cid29.MagneticResonance,
+        '1.2.840.10008.5.1.4.1.1.4.2': codes.cid29.MagneticResonance,
+        '1.2.840.10008.5.1.4.1.1.4.3': codes.cid29.MagneticResonance,
+        '1.2.840.10008.5.1.4.1.1.4.4': codes.cid29.MagneticResonance,
+        '1.2.840.10008.5.1.4.1.1.6.1': codes.cid29.Ultrasound,
+        '1.2.840.10008.5.1.4.1.1.6.2': codes.cid29.Ultrasound,
+        #  Skipping Secondary Capture - not an acquisition modality.
```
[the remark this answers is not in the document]
seandoyle: But OT isn't in [CID 29](http://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_CID_29.html).

src/highdicom/sr/templates.py
```diff
@@ -2751,10 +2752,10 @@ def __init__(
 
         """  # noqa: E501
         super().__init__()
-        modality = get_coded_modality(dataset)
+        modality = get_coded_modality(dataset.SOPClassUID)
         if not modality:
             raise ValueError(f'Dataset has an unsupported SOPClassUID {dataset.SOPClassUID}')  # noqa: E501
-        if not self._is_dicom_image(dataset):
+        if not is_dicom_image(dataset.SOPClassUID):
```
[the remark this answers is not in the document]
seandoyle: Yes - that's much more robust.

src/highdicom/sr/utils.py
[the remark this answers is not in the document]
hackermd: The context group is "extensible". Whenever possible, we should rely on values from the value sets (i.e., context groups), but sometimes it is necessary to use another value. I am also fine with not having an entry for secondary capture..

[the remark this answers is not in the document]
seandoyle: Agreed. Would you want an OT or a SC? The first is more generic but the latter is a more accurate mapping from the Secondary Capture SOP Class.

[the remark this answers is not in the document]
hackermd: I don't think there is a `SC` modality, is there?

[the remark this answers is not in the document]
seandoyle: I was sure that there was - I've had to write code to handle them in previous work. But it's not in the standard! So will use OT!

src/highdicom/sr/utils.py
CPBridge: @seandoyle @hackermd you may want to look at [this method](https://github.com/MGHComputationalPathology/highdicom/blob/8fa057ba15e3dc21270c946dfc25de5c0429cd79/src/highdicom/sr/content.py#L27), which seems very similar in intent

hackermd: Indeed! Thanks for pointing this out @CPBridge.

However, I would suggest changing the existing function ([here](https://github.com/MGHComputationalPathology/highdicom/blob/8fa057ba15e3dc21270c946dfc25de5c0429cd79/src/highdicom/sr/content.py#L50-L54)) and use metadata attributes such as Rows, Columns, etc. rather than the Pixel Data attributes.
In case of large image datasets (whole slide images in pathology) we don't necessary want require users to pass the full data set, but just the metadata.

src/highdicom/sr/templates.py
```diff
@@ -482,6 +486,103 @@ def _contains_image_items(
     return False
 
 
+HighDicomCodes = {
+    "OT": Code(value='1000', scheme_designator="HIGHDICOM", meaning="Modality type OT"),   # noqa: E501
+}
```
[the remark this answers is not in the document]
seandoyle: Done.

src/highdicom/sr/templates.py
```diff
@@ -482,6 +486,103 @@ def _contains_image_items(
...
+        '1.2.840.10008.5.1.4.1.1.4.2': codes.cid29.MagneticResonance,
+        '1.2.840.10008.5.1.4.1.1.4.3': codes.cid29.MagneticResonance,
+        '1.2.840.10008.5.1.4.1.1.4.4': codes.cid29.MagneticResonance,
+        '1.2.840.10008.5.1.4.1.1.6.1': codes.cid29.Ultrasound,
+        '1.2.840.10008.5.1.4.1.1.6.2': codes.cid29.Ultrasound,
+        '1.2.840.10008.5.1.4.1.1.7': HighDicomCodes['OT'],
+        '1.2.840.10008.5.1.4.1.1.7.1': HighDicomCodes['OT'],
+        '1.2.840.10008.5.1.4.1.1.7.2': HighDicomCodes['OT'],
+        '1.2.840.10008.5.1.4.1.1.7.3': HighDicomCodes['OT'],
+        '1.2.840.10008.5.1.4.1.1.7.4': HighDicomCodes['OT'],
```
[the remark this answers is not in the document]
seandoyle: @hackermd I believe that they should. In the future there might be a sequence of models applied to the original acquisition data and some of the later stages in the sequence might use SEG, PM, &etc images from the earlier stages. Also - if a radiologist has a UI that permits updating of an SR they may wish to refer to images that were not in the original acquisition.

src/highdicom/sr/templates.py
```diff
@@ -3519,6 +3623,253 @@ def __init__(
                 item.RelationshipType = relationship_type.value
                 self.append(item)
 
+    def _generate_projection_radiography_descriptors(self, dataset: Dataset) \
+            -> Sequence[ContentItem]:
```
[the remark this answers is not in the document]
seandoyle: Done.

src/highdicom/sr/templates.py
```diff
@@ -3519,6 +3632,297 @@ def __init__(
...
+                    value='mm',
+                    meaning='millimeter',
+                    scheme_designator='UCUM'
+                )
+            )]
+
+        # Not yet implemented from TID 1603 - unclear how to derive data:
+        # - ImageView, ImageViewModifiers
+        # - PositionerPrimaryAngle, PositionerSecondaryAngle
+
```
hackermd, line 3708: Since these are all optional (user defined), it's fine being opinionated and leaving them out.

```suggestion
```

src/highdicom/sr/templates.py
```diff
@@ -3519,6 +3632,297 @@ def __init__(
...
+        return descriptors
+
+    def _generate_cross_sectional_descriptors(
+        self,
+        dataset: Dataset
+    ) -> Sequence[ContentItem]:
+
```
hackermd, line 3715:
```suggestion
```

src/highdicom/sr/templates.py
```diff
@@ -3519,6 +3632,297 @@ def __init__(
                 item.RelationshipType = relationship_type.value
                 self.append(item)
 
+    def _generate_projection_radiography_descriptors(
+        self,
+        dataset: Dataset
+    ) -> Sequence[ContentItem]:
+
+        """
```
hackermd, line 3640:
```suggestion
        """Generate descriptors for projection radiography modalities.
```

src/highdicom/sr/templates.py
```diff
@@ -3519,6 +3632,297 @@ def __init__(
...
+        Parameters
+        ----------
+        dataset: pydicom.Dataset
+            A pydicom Dataset of a cross-sectional image.
+
+        Returns
+        -------
+        Sequence[highdicom.sr.ContentItem]
+            SR Content Items describing the image.
+        """  # noqa: E501
```
hackermd, line 3729:
```suggestion

        """  # noqa: E501
```

src/highdicom/sr/templates.py
```diff
@@ -4415,27 +4809,41 @@ def __init__(
             ),
             relationship_type=RelationshipTypeValues.CONTAINS
         )
-        content = ContentSequence()
-        if groups is not None:
-            for descriptor_items in groups:
-                group_item = ContainerContentItem(
+        library_item.ContentSequence = ContentSequence()
+        if datasets is not None:
+            # We will use just one group
```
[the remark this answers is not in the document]
hackermd: > In your last comment - did you mean Frame of Reference UID instead of Transfer Syntax UID?

Yes, sorry I meant Frame of Reference UID.

> In this case - should we create groups by modality/frame of reference where these are available - and otherwise fall back on only modality to create the groups?

Good point! Yes, that makes sense.

hackermd (changes requested):
Only a few minor style issues. Otherwise LGTM.

src/highdicom/sr/templates.py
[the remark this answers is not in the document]
fedorov: Another use case for this is grouping images by specific series. This is the example we put together to accompany the prostate reporting templates: http://dicom.nema.org/medical/dicom/current/output/chtml/part17/sect_WWWW.2.html.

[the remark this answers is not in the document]
seandoyle: The code now tests images to see if they contain a FrameOfReferenceUID; if they do not only the Modality is used for creating groups.  For a future PR we might pass a function that generated different partitions or provide an option to specify an ImageLibrary as an argument.

src/highdicom/sr/utils.py
hackermd: @seandoyle I would either suggest moving the function into the root package (into `src/highdicom/utils.py`), since it may be useful beyond SR (unless this would lead to circular dependencies).

src/highdicom/sr/templates.py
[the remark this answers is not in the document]
hackermd: @seandoyle ok, let's add Segmentation (modality SEG) and Parametric Map (modality OT?).

src/highdicom/sr/templates.py
```diff
@@ -3546,8 +3897,8 @@ def __init__(
         language_of_content_item_and_descendants: Optional[
             LanguageOfContentItemAndDescendants
         ] = None,
-        image_library_groups: Optional[
-            Sequence[ImageLibraryEntryDescriptors]
+        image_library_datasets: Optional[
+            Sequence[Dataset]
```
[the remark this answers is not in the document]
hackermd: I agree with @CPBridge. The less the user has to know about this the better. I would just create the library in an opinionated way. We could still expose more functionality in the future if this should become necessary.

src/highdicom/sr/utils.py
```diff
@@ -167,3 +167,29 @@ def get_coded_value(item: Dataset) -> CodedConcept:
...
+                                      'Columns',
+                                      'SamplesPerPixel',
+                                      'PhotometricInterpretation',
+                                      'BitsAllocated',
+                                      'BitsStored',
+                                      'HighBit',
+                                      'PixelRepresentation')):
+        return True
+    else:
+        return False
```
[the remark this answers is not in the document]
hackermd: I would probably leave the other TIDs out of scope for now. We can later still add more to the library if this should become necessary.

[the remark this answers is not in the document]
hackermd: > I did remove BitsStored, Highbit, and PixelRepresentation - but importing from the root highdicom package sets up a circular import. I've left it where it was until we can discuss.

Bummer! In this case, it may be better to keep it as a private function (`_is_image()`) directly in `templates.py` for the time being.

src/highdicom/sr/utils.py
hackermd: Turned out this would lead to a circular dependency (https://github.com/MGHComputationalPathology/highdicom/pull/83#discussion_r754608527).

src/highdicom/sr/templates.py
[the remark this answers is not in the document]
seandoyle: Parametric Maps have the modality that matches their acquisition modality or 'OT' if the map is derived from multiple modalities ([Table C.8.31-1](https://dicom.nema.org/medical/dicom/2016c/output/chtml/part03/sect_C.8.32.html#table_PS3.3_C.8.32-1)). But the modality field here is supposed to be constrained to the elements in [CID 29](https://dicom.nema.org/medical/dicom/2020b/output/chtml/part16/sect_CID_29.html) which are much more expressive strings than the standard DICOM modality field. Could we use [CID 32](https://dicom.nema.org/medical/dicom/2020b/output/chtml/part16/sect_CID_32.html) instead? These seem much more closely paired (29 contains Acquisition Modalities, 32 contains Non-Acquisition Modalities).

src/highdicom/sr/utils.py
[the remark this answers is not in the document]
seandoyle: Done.

src/highdicom/sr/templates.py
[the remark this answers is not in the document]
seandoyle: I'm a little unclear what needs to be changed here. We agree that the ImageLibrary should be created in the constructor of the `MeasurementReport` - which is how this is working now.

src/highdicom/sr/templates.py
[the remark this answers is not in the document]
hackermd: Good point @seandoyle! I think the Value Set Constraint of the [(121139, DCM, "Modality") row](https://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#para_6d978b84-80de-4cfe-b4c9-acfc24d771b0) should be [CID 33 Modality](https://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_CID_33.html), which includes both CID 29 "Acquisition Modality" and CID 32 "Non-Acquisition Modality". There is no reason to constrain the value to acquisition modalities only. To the contrary, TID 1500 "Measurement Report" often references Segmentation instances and including entries in the library for those references would be valuable.

src/highdicom/sr/templates.py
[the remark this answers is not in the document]
hackermd: Resolving the comment, since it is already implemented as suggested.

src/highdicom/sr/templates.py
seandoyle: Done.

src/highdicom/sr/templates.py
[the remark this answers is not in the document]
seandoyle: I've added in all the CID32 items except for these - couldn't find SOP Classes that matched.
DOC - Document
HC - Hard Copy
M3D - Model for 3D Manufacturing
PLAN - Plan (all plans I saw were RT Plans - which is a separate CID32 category)
STAIN - Automated Slice Stainer
TEXTUREMAP - Texture Map

seandoyle (commented):
I think that this completes the open items for this PR. All of the CID32 items which had matching SOP Class UIDs in [part 4 ](https://dicom.nema.org/medical/dicom/current/output/chtml/part04/sect_B.5.html)were added; others not.

src/highdicom/sr/templates.py
```diff
@@ -482,6 +486,168 @@ def _contains_image_items(
...
+    -------
+    bool
+        ``True`` if `dataset` is an image, ``False`` otherwise
+
+    """
+    if all(key in dataset for key in ('Rows',
+                                      'Columns',
+                                      'SamplesPerPixel',
+                                      'PhotometricInterpretation',
+                                      'BitsAllocated')):
```
CPBridge, line 645: Preferred formatting

```suggestion
    if all(key in dataset for key in (
        'Rows',
        'Columns',
        'SamplesPerPixel',
        'PhotometricInterpretation',
        'BitsAllocated'
    )):
```

src/highdicom/sr/templates.py
```diff
@@ -3479,33 +3645,32 @@ def from_sequence(
...
-            Frame of Reference UID
-        pixel_data_rows: int
-            Number of rows in pixel data frames
-        pixel_data_columns: int
-            Number of rows in pixel data frames
-        additional_descriptors: Union[Sequence[highdicom.sr.ContentItem], None], optional
-            Additional SR Content Items that should be included
+        image: pydicom.Dataset
+            Metadata of a referenced image instance
+        additional_descriptors: Sequence[highdicom.sr.ContentItem],
```
CPBridge, line 3662: This is how we have been doing parameter types elsewhere
```suggestion
        additional_descriptors: Union[Sequence[highdicom.sr.ContentItem], None], optional
```

src/highdicom/sr/templates.py
```diff
@@ -3516,23 +3681,24 @@ def __init__(
...
-            relationship_type=RelationshipTypeValues.HAS_ACQ_CONTEXT
-        )
-        self.append(frame_of_reference_uid_item)
+        if 'FrameOfReferenceUID' in image:
+            frame_of_reference_uid_item = UIDRefContentItem(
+                name=CodedConcept(
+                    value='112227',
+                    meaning='Frame of Reference UID',
+                    scheme_designator='DCM'
+                ),
```
CPBridge, line 3690: For many of these, wouldn't it be preferable to use the codes in `pydicom`? E.g.

```suggestion
                name=codes.FrameOfReferenceUID,
```

src/highdicom/sr/templates.py
```diff
@@ -3516,23 +3681,24 @@ def __init__(
...
+                value=image.FrameOfReferenceUID,
+                relationship_type=RelationshipTypeValues.HAS_ACQ_CONTEXT
+            )
+            self.append(frame_of_reference_uid_item)
         pixel_data_rows_item = NumContentItem(
             name=CodedConcept(
                 value='110910',
                 meaning='Pixel Data Rows',
                 scheme_designator='DCM'
             ),
```
CPBridge, line 3700:
```suggestion
            name=codes.DCM.PixelDataRows,
```

src/highdicom/sr/templates.py
```diff
@@ -3516,23 +3681,24 @@ def __init__(
...
-            value=pixel_data_rows,
+            value=image.Rows,
             relationship_type=RelationshipTypeValues.HAS_ACQ_CONTEXT,
             unit=CodedConcept(
                 value='{pixels}',
```
CPBridge, line 3704: This seems wrong...

src/highdicom/sr/templates.py
```diff
@@ -4441,21 +4904,55 @@ def get_image_measurement_groups(
...
+        dataset: Dataset,
+    ) -> None:
+        """
+        Parameters
+        ----------
+        dataset: Dataset
+            Image to include in ImageLibrary
+
+        """  # noqa
+
```
CPBridge, line 4923: Should be no blank line
```suggestion
```

CPBridge (commented):
Had another look over. We should wait for @hackermd 's approval but looks basically good to go with some minor suggestions. Most significant being simplifying a large chunk of the code by replacing verbose CodedConcepts with the codes already in pydicom to save space, increase readability and maintainability, and reduce redundancy and likelihood of mistakes.

fedorov:
What is the status of this PR?

Do I understand it correctly that with the current functionality available in the library it is not possible to include IMAGE content items in the Image library container ([TID 1601](https://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1601)).

seandoyle:
> 

It is nearly done -  I need to respond to a request here. I've let this go too long. I can take a look at it tonight.

hackermd:
> Do I understand it correctly that with the current functionality available in the library it is not possible to include IMAGE content items in the Image library container ([TID 1601](https://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1601)).

@fedorov the library library container is currently left empty. As with other content, one can add it manually after construction of the object. 

The changes in this PR will automate the process and automatically include referenced objects to the library.

src/highdicom/sr/templates.py
seandoyle: Agreed. I've changed these on my PR's code - should I change this in the rest of templates.py or make that a separate PR?

src/highdicom/sr/templates.py
seandoyle: Not sure what's incorrect here - the value should be the [number of rows in the pixel data of the image.](https://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_D.html#DCM_110910). The relationship with the parent should be 'Has Acquisition Context'.  What am I missing?

seandoyle (commented):
Still in progress - want to check to make sure that all comments have been addressed.

tests/test_sr.py
```diff
@@ -1872,7 +1871,7 @@ def test_construction_with_frame_reference(self):
         )
         assert (
```
seandoyle, line 1872: Change made so that unit tests pass - was there a change in how frame numbers were managed?

src/highdicom/sr/templates.py
CPBridge: It's the code used to describe the unit that I'm unsure about, i.e. `'{pixels}'`. Initially I thought you were trying to do a python f-string but looking into UCUM a bit more, I see that the curly braces denote some sort of annotation.

Technically speaking of course, this has no unit since it's simply a count. But if we think this is the right way to do it, we can go ahead with this

tests/test_sr.py
CPBridge: Indeed, see #162 and its resolution in #163

CPBridge: Basically it was a breaking change to the way pydicom deals with multivalue elements that have a single value, introduced in pydicom 2.3.0, it broke a bunch of stuff. This fix is fine

src/highdicom/sr/templates.py
```diff
@@ -3732,6 +3894,225 @@ def __init__(
...
+            TextContentItem(
+                name=codes.DCM.PatientOrientationColumn,
+                value=patient_orientation[1],
+                relationship_type=RelationshipTypeValues.HAS_ACQ_CONTEXT,
+            ),
+            NumContentItem(
+                name=codes.DCM.HorizontalPixelSpacing,
+                value=pixel_spacing[1],
+                relationship_type=RelationshipTypeValues.HAS_ACQ_CONTEXT,
+                unit=CodedConcept(
```
CPBridge, line 3934: note that the defined UCUM codes are in pydicom too, so a lot of this could be replaced with `codes.UCUM.mm`

src/highdicom/sr/templates.py
```diff
@@ -489,6 +494,169 @@ def _contains_image_items(
...
+    if all(key in dataset for key in (
+        'Rows',
+        'Columns',
+        'SamplesPerPixel',
+        'PhotometricInterpretation',
+        'BitsAllocated'
+    )):
+        return True
+    else:
+        return False
```
CPBridge, line 657: Note that there is now a function `highdicom._module_utils.does_iod_have_pixel_data` whose intent I think is very similar to this (determining whether a dataset represents an image when the pixel data may not actually be in the dataset). Could we use this here instead to reduce redundancy?

CPBridge (changes requested):
I made some really minor suggestions. Also there is a new file `highdicom_requirements.txt` that seems to have been added accidentally (I can't see a purpose for it). Can we remove that file?

src/highdicom/sr/templates.py
seandoyle: Yes - the highdicom_requirements.txt is deleted. I've been playing around with the requirements to try to get all of the tests to run on my M1 Mac. There is one test that still fails (TestEncodeFrame.test_jpegls_monochrome) - there is some issue with jpeg-ls component.  But this doesn't happen on Intel Linux machines.

I've replaced the constructors for the UCUM elements with values from pydicom for the most part. The ones with the annotations for {pixels} have been replaced `codes.UCUM.NoUnits`. The only ones with annotations that are left are for the orientation vectors where the values are constrained to `{-1:1}`.

src/highdicom/sr/templates.py
seandoyle: Done. There was a similar request from Markus in the past but it lead to a recursive in Python.

seandoyle: @CPBridge, @hackermd  - I've made this change but [note](https://github.com/herrmannlab/highdicom/pull/83#discussion_r656516824) this comment from Markus. He thought that a user might pass the Dataset without the PixelData for performance reasons. This was the reason that we were testing for non-PixelData members.

src/highdicom/sr/templates.py
seandoyle: I think it is the right way because otherwise the DicomSRValidator.sh tool reports it as an error.

src/highdicom/sr/templates.py
seandoyle: The odd {pixels} UCUM annotations were restored because the DicomSRValidator.sh validator requires them.

CPBridge: ok I just wanted to check that there was a reason behind this because it seems a little strange to me

src/highdicom/sr/templates.py
CPBridge: Agreed that that is a consideration. When I wrote the `does_iod_have_pixel_data` function this was a goal too. That function doesn't check whether the *dataset* contains pixel data elements, it instead checks whether the *IOD* contains pixel data so it doesn't matter whether the pixel data has been dropped for performance reasons.

CPBridge (approved):
In my view this is good to go. Thanks so much for the contribution @seandoyle !

src/highdicom/sr/templates.py
```diff
@@ -3721,6 +3847,16 @@ def __init__(
             )
         )
         self.append(pixel_data_cols_item)
+
+        if self._is_cross_sectional(image):
```
hackermd, line 3851: Here, we may want to add additional descriptors for SM images (e.g., Total Pixel Matrix Rows/Columns). However, this is not yet part of the standard and we can address this later.

src/highdicom/sr/templates.py
```diff
@@ -3759,8 +4078,8 @@ def __init__(
         language_of_content_item_and_descendants: Optional[
             LanguageOfContentItemAndDescendants
         ] = None,
-        image_library_groups: Optional[
-            Sequence[ImageLibraryEntryDescriptors]
+        referenced_images: Optional[
+            Sequence[Dataset]
```
hackermd, line 4082: Since this will be a breaking change, we should add a note to the corresponding section of the documentation.

src/highdicom/sr/templates.py
```diff
@@ -3644,60 +3786,48 @@ def from_sequence(
 
 class ImageLibraryEntryDescriptors(Template):
 
-    """`TID 1602 <http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1602>`_
-     Image Library Entry Descriptors"""  # noqa: E501
+    """:dcm:`TID 1602 <part16/chapter_A.html#sect_TID_1602>`
+     Image Library Entry Descriptors"""
```
hackermd, line 3790:
```suggestion
    """:dcm:`TID 1602 Image Library Entry Descriptors <part16/chapter_A.html#sect_TID_1602>`
    """  # noqa: E501
```

src/highdicom/sr/templates.py
```diff
@@ -3732,6 +3868,189 @@ def __init__(
                 item.RelationshipType = relationship_type.value
                 self.append(item)
 
+    def _generate_projection_radiography_descriptors(
+        self,
+        dataset: Dataset
+    ) -> Sequence[ContentItem]:
+
```
hackermd, line 3875:
```suggestion
```

src/highdicom/sr/templates.py
```diff
@@ -4625,51 +4936,104 @@ def get_image_measurement_groups(
...
+    def __init__(
+        self,
+        dataset: Dataset,
+    ) -> None:
+        """
+        Parameters
+        ----------
+        dataset: Dataset
+            Image to include in ImageLibrary
```
hackermd, line 4952:
```suggestion
        dataset: pydicom.dataset.Dataset
            Image to include in image library
```

src/highdicom/sr/templates.py
```diff
@@ -4625,51 +4936,104 @@ def get_image_measurement_groups(
...
+
+        """  # noqa
```
hackermd, line 4954:
```suggestion
        """
```

src/highdicom/sr/templates.py
```diff
@@ -4625,51 +4936,104 @@ def get_image_measurement_groups(
...
-        groups: Optional[Sequence[ImageLibraryEntryDescriptors]] = None
+        datasets: Sequence[Dataset]
     ) -> None:
         """
         Parameters
         ----------
-        groups: Union[Sequence[Sequence[highdicom.sr.ImageLibraryEntryDescriptors]], None], optional
-            Entry descriptors for each image library group
+        datasets: Sequence[pydicom.dataset.Dataset]
+            Datasets to include in ImageLibrary
```
hackermd, line 4980:
```suggestion
        datasets: Sequence[pydicom.dataset.Dataset]
            Datasets to include in image library
```

hackermd: Should we clarify that these datasets must be images?

hackermd (changes requested):
Thank you @seandoyle! Looks good overall. I only found a few minor things.

src/highdicom/sr/templates.py
seandoyle: Maybe this should be addressed with changes needed for EnhancedMR?

hackermd (approved):

PR #84, opened 2021-06-16
Tests for setting content sequence

CPBridge:
Add an integration test between ContentItem and ContentSequence that setting the ContentSequence of a ContentItem should trigger a check that all items in the new sequence have relationship types.


hackermd (approved):
Nice!

Issue #86, opened 2021-07-02
Iter segments does not function properly when there is a single frame in a segmentation image


CPBridge:
This is because pydicom.Dataset.pixel_array returns a 2D rather than 3D array in this case. As a result, iter_segments ends up iterating through the rows of the single frame

PR #87, opened 2021-07-04
Fix iter_segments bug with single frame segs. Add test

CPBridge:
Simple fix for #86 . Add test file with a single segmentation frame, and use it for a regression test for #86


hackermd (approved):
Thanks @CPBridge for fixing this

PR #89, opened 2021-07-07
Implementation of Parametric Map IOD

hackermd:


hackermd:
@cgorman I just noticed that single quotes got replace by double quotes all over the place. Could you kindly undo that?

src/highdicom/map/content.py
```diff
@@ -0,0 +1,120 @@
+from typing import Optional, Sequence, Tuple, Union
+
+from highdicom.sr.coding import CodedConcept
+from pydicom.dataset import Dataset
+from pydicom.sr.coding import Code
+
+
+class RealWorldValueMapping(Dataset):
```
hackermd, line 8: @dclunie could you kindly take a look at the implementation of the [Real World Value Mapping Item Macro](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.16.2.html#sect_C.7.6.16.2.11). In particular, I would appreciate if you could assess we interpreted *Real World Value First/Last Value Mapped* correctly.

In general, I wonder whether that functional group macro should be required, given that a relationship to real world values may not exist for many use cases.

The standard states

> The Items in the Real World Value Mapping Sequence (0040,9096) **may be used** to translate stored values into real world values **when there is such a relationship**

Making the usage `C` instead of `M` would simplify things quite a bit in my opinion.

dclunie: These are required and that is not going to change. This has been discussed extensively in the past.

Worst case, for dimensionless properties, an identity transformation over the full range of possible values is specified.

hackermd: Thanks for clarification @dclunie

src/highdicom/seg/sop.py
```diff
@@ -893,6 +850,62 @@ def add_segments(
         if len(self.PixelData) % 2 == 1:
             self.PixelData += b'0'
 
+    def _format_pixel_array(
```
cgorman, line 853: Is this supposed to return `np.ndarray`?

hackermd: Yes, it's supposed to. Fixed with 3badeb3c2910721dd2dc6dbe1fc7cf35bded1787

src/highdicom/map/sop.py
```diff
@@ -0,0 +1,700 @@
...
+
+            # Plane Position (Patient/Slide)
+            if self._coordinate_system == CoordinateSystemNames.SLIDE:
+                pffg_item.PlanePositionSlideSequence = plane_positions[i]
+            else:
+                pffg_item.PlanePositionSequence = plane_positions[i]
+
+            # Frame Content
+            frame_content_item = Dataset()
+            # FIXME
```
cgorman, line 549: What needs to be fixed?

hackermd: Thanks @cgorman for pointing that out.

@CPBridge we should discuss the implementation of Dimension Index Values. It's currently wrong in Segmentation and we'll have to fix it here, too.

CPBridge: yeah, I still haven't started this. Hoping to get to it sometime in the next week or so. I'll try and make the fix applicable here too if I can

hackermd: Ok. Thanks @CPBridge. I'll remove the FIXME comment and create an issue instead to keep track of it.

src/highdicom/map/sop.py
```diff
@@ -0,0 +1,767 @@
...
+            all_frames.append(self._encode_pixels(plane))
+            pixel_data_bytes = encapsulate(all_frames)
+        else:
+            pixel_data_bytes += self._encode_pixels(plane)
+
+        # FIXME(cg): This fails with the latest release of PyDICOM because the
+        # when using Double Float Pixel Data. The bug is fixed in master but
+        # has not been released yet.
+        # See https://github.com/pydicom/pydicom/pull/1413 for more details
+        setattr(self, pixel_data_attr, pixel_data_bytes)
```
cgorman, line 654: Just so we can track it in the review, this fails on the latest release of PyDICOM when using Double Float Pixel Data. The bug is fixed in their master branch, though.

https://github.com/pydicom/pydicom/pull/1413

src/highdicom/map/sop.py
```diff
@@ -0,0 +1,789 @@
...
+            if plane.dtype == np.uint8:
+                return plane.astype(np.uint16).flatten().tobytes()
+            elif plane.dtype.kind == 'i':
+                plane = plane.astype(np.int16) + 2 ** 16 / 2
+                return plane.astype(np.uint16).flatten().tobytes()
+            else:
+                return plane.flatten().tobytes()
+    
+    @property
+    def pixel_array(self) -> np.ndarray:
```
cgorman, line 767: @hackermd here's what I came up with as a workaround for the PixelRepresentation bug in pydicom, let me know what you think.

.gitignore
```diff
@@ -108,3 +108,6 @@ venv.bak/
 
 #IntelliJ
 .idea
+
+# Visual Studio Code
+.vscode
```
CPBridge, line 113: add newline at end of file

setup.py
```diff
@@ -38,7 +38,7 @@
     package_dir={'': 'src'},
     python_requires='>=3.6',
     install_requires=[
-        'pydicom>=1.4,!=2.1.0,!=2.1.1',
+        'pydicom>2.1',
```
CPBridge, line 41: not sure this is still correct, following discussions?

src/highdicom/map/content.py
```diff
@@ -0,0 +1,127 @@
...
+            and is used to determine whether values are stored as integers or
+            floating-point values. Therefore, use ``(0.0, 1.0)`` instead of
+            ``(0, 1)`` to specify a range of floating-point values.
+        slope: Union[int, float, None], optional
+            Slope of the linear mapping function applied to values in
+            `value_range`.
+        intercept: Union[int, float, None], optional
+            Intercept of the linear mapping function applied to values in
+            `value_range`.
+        lut_data: Union[Sequence[int], Sequence[float] None], optional
```
CPBridge, line 49:
```suggestion
        lut_data: Union[Sequence[int], Sequence[float], None], optional
```
missing comma

src/highdicom/map/sop.py
```diff
@@ -0,0 +1,789 @@
...
+        source_images: Sequence[pydicom.Dataset]
+            Sequence of one or more `pydicom.Dataset`s that this parametric map
+            is derived from
+        pixel_array: numpy.ndarray
+            Array of parametric map pixel data of unsigned integer or
+            floating-point data type representing one or more frames of the
+            parametric map pixel data. The values are supposed to represent a
+            single "feature", i.e., be the result of one set of image
+            transformations such that the same `real_world_value_mappings`
+            apply.
```
CPBridge, line 78: I think more explanation is probably needed here about the shape/dimensions, what order they come in and how they need to relate the the source images (or plane positions)

src/highdicom/map/sop.py
```diff
@@ -0,0 +1,789 @@
...
+            difficult to describe (e.g., in case of the feature maps of a deep
+            convolutional neural network model).
+        window_center: Union[int, float, None], optional
+            Window center for rescaling stored values for display purposes by
+            applying a linear transformation function. For example, in case of
+            floating-point values in the range ``[0.0, 1.0]``, the window
+            center would be ``0.5``, in case of floating-point values in the
+            range ``[-1.0, 1.0]`` the window center would be ``0.0``, in case
+            of unsigned integer values in the range ``[0, 255]`` the window
+            center would be ``128``.
```
CPBridge, line 113: My understanding is that this is just a suggestion that viewers can ignore and users of viewers can alter anyway. I would change the language from "would be" to "could be" or "may be". There may be good reasons to choose a window center somewhere other than the center of the full range, if for example most of interesting detail exists within a certain range of values. Similarly for window width below

src/highdicom/map/sop.py
```diff
@@ -0,0 +1,789 @@
...
+            Serial number (or other identifier) of the device (software)
+            that creates the instance
+        contains_recognizable_visual_features: bool
+            Whether the image contains recognizable visible features of the
+            patient
+        real_world_value_mappings: Sequence[highdicom.map.RealWorldValueMapping]
+            Descriptions of how stored values map to real-world values. The
+            concept of real-world values is a bit fuzzy and the mapping may be
+            difficult to describe (e.g., in case of the feature maps of a deep
+            convolutional neural network model).
```
CPBridge, line 105: Need an explanation about why there may be more than 1, and how this relates to the pixel_array. Also may want to make this a bit more formal in tone?

src/highdicom/map/sop.py
```diff
@@ -0,0 +1,789 @@
...
+        # Parametric Map Frame Type
+        frame_type_item = Dataset()
+        frame_type_item.FrameType = self.ImageType
+        shared_fg_item.ParametricMapFrameTypeSequence = [frame_type_item]
+
+        # Real World Value Mapping Sequence
+        # If the input was a single RWVM or a sequence of size 1 then we will
+        # assign it to the Shared FG Seq. Otherwise it will be per-frame and
+        # will be checked later on
+        if len(real_world_value_mappings) == 1:
```
CPBridge, line 436: Suggest a runtime check that the `real_world_value_mappings` parameter has the correct type

src/highdicom/map/sop.py
```diff
@@ -0,0 +1,789 @@
...
+            plane_orientation = source_plane_orientation
+        self._plane_orientation = plane_orientation
+        self._source_plane_orientation = source_plane_orientation
+
+        # TODO: Double check correctness of DimensionIndexSequence, see
+        # discussion on Segmentation
+        # Also TODO: This includes some segmentation-specific stuff. Fix that
+        self.DimensionIndexSequence = DimensionIndexSequence(
+            coordinate_system=self._coordinate_system
+        )
```
CPBridge, line 382: Just a comment to confirm that this is indeed wrong as it contains the `SegmentNumber` hardcoded as a dimension, which does not apply in this case

src/highdicom/map/sop.py
```diff
@@ -0,0 +1,789 @@
...
+        -------
+        bytes
+            `plane` encoded as a `bytes` object
+
+        Raises
+        ------
+        ValueError
+            If the SOP instance uses an encapsulated transfer syntax and
+            `plane` is not exactly 2 dimensional.
+        """
```
CPBridge, line 741:
```suggestion
            `plane` is not exactly 2 dimensional.

        """
```

src/highdicom/map/sop.py
```diff
@@ -0,0 +1,789 @@
...
+        use the numpy pixel_data handler, however this goes against the DICOM
+        standard. To get around this issue we set the Pixel Representation
+        attribute if it is not set, then get the array, then remove the
+        attribute.
+
+        Returns
+        -------
+        numpy.ndarray
+            The pixel data encoded as a `numpy.ndarray`.
+        """
```
CPBridge, line 780:
```suggestion
            The pixel data encoded as a `numpy.ndarray`.

        """
```

src/highdicom/map/sop.py
```diff
@@ -0,0 +1,789 @@
...
+            )
+        else:
+            if plane.dtype == np.uint8:
+                return plane.astype(np.uint16).flatten().tobytes()
+            elif plane.dtype.kind == 'i':
+                plane = plane.astype(np.int16) + 2 ** 16 / 2
+                return plane.astype(np.uint16).flatten().tobytes()
+            else:
+                return plane.flatten().tobytes()
+
```
CPBridge, line 765: dead whitespace
```suggestion

```

CPBridge (changes requested):
Had a look through and made some comments. Overall looks nice but I think the docstrings could do with a bit more detail. E.g. I'm very confused by why/how there are multiple real world value mappings per object, and how this relates to the different frames in the image and pixel_array. As has been previously noted, there are some issues with the existing implementation of the dimension indices. I will try to fix these in a PR

src/highdicom/map/sop.py
```diff
@@ -0,0 +1,789 @@
...
+    
+    @property
+    def pixel_array(self) -> np.ndarray:
+        """This is a workaround for a bug in Pydicom.
+
+        In PyDicom you are required to set PixelRepresentation in order to
+        use the numpy pixel_data handler, however this goes against the DICOM
+        standard. To get around this issue we set the Pixel Representation
+        attribute if it is not set, then get the array, then remove the
+        attribute.
```
CPBridge, line 774: I appreciate the detailed comment, but as written this docstring for this public property will go into the sphinx generated user documentation. The docstring should be written for users and hide this sort of detail, and comments like this should probably be code comments at the top of the method.

src/highdicom/map/sop.py
hackermd: Yes, I saw that. That was quite an elegant solution. However, the problem is that the Pixel Data element in the `pydicom.Dataset` could not be decoded when it would be read by `pydicom.dcmread()`. Including the attribute into the data set is also not ideal, though. I will report this to pydicom..

src/highdicom/map/sop.py
CPBridge: This is now outdated

src/highdicom/map/sop.py
CPBridge: This is now outdated

src/highdicom/map/sop.py
```diff
@@ -0,0 +1,706 @@
...
+        # If the same set of mappings applies to all frames, the information
+        # is stored in the Shared Functional Groups Sequence. Otherwise, it
+        # is stored for each frame separately in the Per Frame Functional
+        # Groups Sequence.
+        if pixel_array.ndim == 2:
+            pixel_array = pixel_array[np.newaxis, ..., np.newaxis]
+            sffg_item.RealWorldValueMappingSequence = real_world_value_mappings
+        elif pixel_array.ndim == 3:
+            pixel_array = pixel_array[..., np.newaxis]
+            sffg_item.RealWorldValueMappingSequence = real_world_value_mappings
```
CPBridge, line 373: Need a check here on the size of the mappings list. What if the user has passed a Sequence[Sequence[RealWorldValueMapping]]?

src/highdicom/map/sop.py
hackermd: @CPBridge I only included the spatial dimensions into the Dimension Index Sequence, since I didn't find an attribute that could be indexed for a particular set of real world value mappings (image transformation). I initially considered using the `RealWorldValueMappingSequence.LUTLabel`, but that has an incompatible type. Dimension Index Values has VR UL (Unsigned Short). Furthermore, the real world value mapping may be shared by all frames and consequently end up in the Shared Functional Groups Sequence rather the Per Frame Functional Groups Sequence.

@dclunie are you aware of a required attribute that could be included in the index?

src/highdicom/map/sop.py
hackermd: Resolving this comment, since the issue will be fixed with the upcoming 2.2 release.

src/highdicom/map/sop.py
hackermd: Here is the pydicom issue: https://github.com/pydicom/pydicom/issues/1457

.gitignore
hackermd: Addressed via a9e561634873616e2cae62ee6d8667fdd1ef8431

setup.py
hackermd: Addressed via 0580ea90d03ec8985f9714f14b1b12d80fb10e00

src/highdicom/map/content.py
hackermd: Addressed via 20ec9fc7fe33b12a3b6827e5e31e921a5fa9e461

src/highdicom/map/sop.py
hackermd: Addressed via 351475c61e7f58446485b552347ed8d54c6f2479

src/highdicom/map/sop.py
hackermd: Addressed via 71f8940b0a303fb5d3695c9e3068ab05f644d053

src/highdicom/map/sop.py
hackermd: Closing this comment, since it's outdated.

src/highdicom/map/sop.py
hackermd: Resolving this conversation, since it's outdated.

src/highdicom/map/sop.py
hackermd: Addressed via 71f8940b0a303fb5d3695c9e3068ab05f644d053

src/highdicom/map/sop.py
hackermd: Addressed via 71f8940b0a303fb5d3695c9e3068ab05f644d053

.gitignore
hackermd: And subsequently fixed via a558940a99bb0559b3c6f56e8c36c6d91b112985

setup.py
hackermd: I initially tried setting `>=2.2`, but that currently doesn't work yet. We'll have to update once the 2.2 version is released.

src/highdicom/map/sop.py
hackermd: Should get merged into pydicom soon: https://github.com/pydicom/pydicom/pull/1458

src/highdicom/map/sop.py
```diff
@@ -0,0 +1,732 @@
...
+            function that maps stored values to unit-less real-world values.
+            Each channel encoded in `pixel_array` shall be described with one
+            or more real-world value mappings.
+            If `pixel_array` is a 2D or 3D array and only one channel exists
+            at each spatial image position), then one or more real-world value
+            mappings shall be provided in a flat sequence.
+            If `pixel_array` is a 4D array and multiple channels exist at each
+            spatial image position, then one or more mappings shall be provided
+            for each channel in a nested sequence of length ``m``, where ``m``
+            shall match the channel dimension of the `pixel_array``.
```
CPBridge, line 133:
```suggestion
            Descriptions of how stored values map to real-world values.
            Each channel encoded in `pixel_array` shall be described with one
            or more real-world value mappings. Multiple mappings might be
            used for different representations such as log versus linear scales
            or for different representations in different units.
            If `pixel_array` is a 2D or 3D array and only one channel exists
            at each spatial image position), then one or more real-world value
            mappings shall be provided in a flat sequence.
            If `pixel_array` is a 4D array and multiple channels exist at each
            spatial image position, then one or more mappings shall be provided
            for each channel in a nested sequence of length ``m``, where ``m``
            shall match the channel dimension of the `pixel_array``.

            In some situations the mapping may be difficult to describe (e.g., in
            case of a transformation performed by a deep convolutional neural
            network). The real-world value mapping may simply describe an
            identity function that maps stored values to unit-less real-world
            values.
```

Swapping around the order here makes this clearer I think. I added a point of clarification about why multiple mappings may be used. I also reworded the last part to have a more formal tone.

CPBridge (approved):
I added a suggested change to the docstring, but otherwise looks good to go

PR #90, opened 2021-07-08
Add :dcm: sphinx link extension to save space in docstrings

CPBridge:
I noticed this trick in pydicom's sphinx config and thought we should adopt it too in order to cut down those long ugly docstrings. Now you can use, for example

```
:dcm:`TID 8002 <part16/chapter_C.html#sect_TID_8002>`
```
and sphinx will automatically expand the URL for you.


hackermd (approved):
Nice! I wasn't aware of that Sphinx feature. Comes very handy.

PR #93, opened 2021-07-26
Highdicom 0.9.0

CPBridge:
Development branch combining work from #69 and #74 (SR and SEG decoding) into a single PR


src/highdicom/sr/templates.py
```diff
@@ -1640,6 +2766,42 @@ def __init__(
             qualitative_evaluations=qualitative_evaluations,
             geometric_purpose=geometric_purpose
         )
+        self[0].ContentTemplateSequence[0].TemplateIdentifier = '1410'
+
+    @property
+    def roi(self) -> Union[ImageRegion, ImageRegion3D, None]:
```
CPBridge, line 2772: Should we have another way to get information about the roi if the roi is encoded "by reference", e.g. a reference to a segment in a segmentation image?

src/highdicom/sr/templates.py
```diff
@@ -1955,19 +3248,454 @@ def __init__(
...
+            context = SubjectContext(
+                subject_class=item.value,
+                subject_class_specific_context=attributes
+            )
+            subject_contexts.append(context)
+        return subject_contexts
+
+    def get_planar_roi_measurement_groups(
+        self,
+        tracking_uid: Optional[str] = None,
```
CPBridge, line 3406: Add tracking_identifier filter in addition to tracking UID?

src/highdicom/sr/templates.py
```diff
@@ -1955,19 +3248,454 @@ def __init__(
...
+        finding_type: Optional[Union[CodedConcept, Code]] = None,
+        finding_site: Optional[Union[CodedConcept, Code]] = None,
```
CPBridge, line 3408: I think filtering by finding site may require bit more thought. In the "encoding" API (and the "finding_sites" property of the decoding API), finding sites are specified using their own `FindingSite` class, which allows for laterality and topographical modifiers. But here we are just checking the name against a coded concept. Should this method accept a `FindingSite` object, and include the modifier and laterality in the definition of a match, so that you can search for the "left lung" for example? Perhaps we should add a `__eq__` method on `FindingSite`?

Additionally I tried search some of my SRs by tracking UID and this method just returned an empty list even when the finding site matched.

src/highdicom/sr/templates.py
```diff
@@ -1955,19 +3248,454 @@ def __init__(
...
+
+        Parameters
+        ----------
+        tracking_uid: str, optional
+            Unique tracking identifier
+        finding_type: Union[highdicom.sr.CodedConcept, pydicom.sr.coding.Code], optional
+            Finding
+        finding_site: Union[highdicom.sr.CodedConcept, pydicom.sr.coding.Code], optional
+            Finding site
+        reference_type: Union[highdicom.sr.GraphicTypeValues, highdicom.sr.GraphicTypeValues3D], optional
```
CPBridge, line 3533: I think this is a mistake
```suggestion
        reference_type: Union[highdicom.sr.CodedConcept, pydicom.sr.coding.Code], optional
```

src/highdicom/sr/templates.py
```diff
@@ -1955,19 +3248,454 @@ def __init__(
...
+        Parameters
+        ----------
+        tracking_uid: str, optional
+            Unique tracking identifier
+        finding_type: Union[highdicom.sr.CodedConcept, pydicom.sr.coding.Code], optional
+            Finding
+        finding_site: Union[highdicom.sr.CodedConcept, pydicom.sr.coding.Code], optional
+            Finding site
+        reference_type: Union[highdicom.sr.CodedConcept, pydicom.sr.coding.Code], optional
+            Type of referenced ROI
```
CPBridge, line 3431: I think this needs more explanation for people to be able to understand what it really means. Probably an explicit list of possible values in the docstring is necessary, and perhaps a check on the input to make sure it makes sense (to give users feedback if they are passing something innappropriate)

src/highdicom/sr/templates.py
```diff
@@ -1955,19 +3248,454 @@ def __init__(
...
+        """Get imaging measurement groups of planar regions of interest.
+
+        Finds (and optionally filters) content items contained in the
+        CONTAINER content item "Measurement group" as specified by TID 1410
+        "Planar ROI Measurements and Qualitative Evaluations".
+
+        Parameters
+        ----------
+        tracking_uid: str, optional
+            Unique tracking identifier
```
CPBridge, line 3425: I tried searching some of my SRs by tracking UID and it returned an empty list when the tracking UID matched.

src/highdicom/sr/templates.py
```diff
@@ -1955,19 +3248,454 @@ def __init__(
...
+        measurement_group_items = self._find_measurement_groups()
+        sequences = []
+        for group_item in measurement_group_items:
+            if group_item.template_id is not None:
+                if group_item.template_id != '1410':
+                    continue
+            else:
+                contains_rois = _contains_planar_rois(group_item)
+                if not(contains_rois):
+                    continue
```
CPBridge, line 3459: Can be simplified?
```suggestion
                if not _contains_planar_rois(group_item):
                    continue
```

src/highdicom/sr/templates.py
```diff
@@ -1955,19 +3248,454 @@ def __init__(
...
+                    matches_graphic_type = _contains_scoord3d_items(
+                        group_item,
+                        name=codes.DCM.ImageRegion,
+                        graphic_type=graphic_type
+                    )
+                matches.append(matches_graphic_type)
+
+            seq = PlanarROIMeasurementsAndQualitativeEvaluations.from_sequence(
+                [group_item]
+            )
```
CPBridge, line 3500: Didn't we decide to eagerly perform type conversion? In which case this is unnecessary and `group_item` can be returned directly?

src/highdicom/sr/templates.py
```diff
@@ -1955,19 +3248,454 @@ def __init__(
...
+        sequences = []
+        measurement_group_items = self._find_measurement_groups()
+        for group_item in measurement_group_items:
+            if group_item.template_id is not None:
+                if group_item.template_id != '1411':
+                    continue
+            else:
+                contains_rois = _contains_volumetric_rois(group_item)
+                if not(contains_rois):
+                    continue
```
CPBridge, line 3562: as above
```suggestion
                if not _contains_volumetric_rois(group_item):
                    continue
```

src/highdicom/sr/templates.py
```diff
@@ -1955,19 +3248,454 @@ def __init__(
...
+            SOP Class UID of the referenced instance, which may be a
+            segmentation image, source image for the region or segmentation, or
+            RT struct, depending on `reference_type`
+
+        Returns
+        -------
+        List[highdicom.sr.PlanarROIMeasurementsAndQualitativeEvaluations]
+            Sequence of content items for each matched measurement group
+
+        """  # noqa: E501
```
CPBridge, line 3448: Should we add checks that a user has provided a sensible combination of inputs? For example it makes no sense to ask for `reference_type` of referenced segment, and also provide a value for `graphic_type` right?

src/highdicom/sr/templates.py
```diff
@@ -1215,6 +2111,40 @@ def __init__(
...
+            value=item.value,
+            unit=item.unit,
+            qualifier=item.qualifier
+        )
+        if 'ContentSequence' in item:
+            measurement.ContentSequence = ContentSequence.from_sequence(
+                item.ContentSequence
+            )
+        else:
+            measurement.ContentSequence = ContentSequence()
```
CPBridge, line 2145: I think having an empty content sequence like this may be invalid. 

According to [the standard](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.17.3.html#table_C.17-6):
> One or more items shall be included in this sequence.

I think it should just be omitted if the original item has no content sequence

CPBridge (commented):
@hackermd I have been playing around with your sr parsing part and left a few comments. A couple of important things:

- Some of the filtering methods are not working for me. For example filtering ROIs by tracking_uid or finding_site is not giving me the answers I expect. I will follow up with a test file to demonstrate this.
- I like the new way of `get_measurements` and `get_qualitative_evaluations` returning the relevant template classes. But I think this would be even better if those classes themselves had properties to access their important attributes, such as name, value, unit. Making sense of the template objects, especially `Measurement`, is a bit unwieldy at the moment.
- In the seg part of this PR, I added a `segread` function (similar to dcmread) that reads in a seg dataset from file and converts to a `hd.seg.Segmentation`in a single line of user code. What do you think about adding a corresponding `srread`, that figures out the correct SR class and does the conversion, after reading from disk.

src/highdicom/seg/sop.py
```diff
@@ -860,8 +746,229 @@ def add_segments(
...
+        Parameters
+        ----------
+        described_segment_numbers: np.ndarray
+            The segment numbers from the segment descriptions, in the order
+            they were passed. 1D array of integers.
+
+        Raises
+        ------
+        ValueError
+            If the described_segment_numbers do not have the required values
```
hackermd:
```suggestion
            If the `described_segment_numbers` do not have the required values
```

src/highdicom/seg/sop.py
```diff
@@ -860,8 +746,229 @@ def add_segments(
...
+        segmentation_type: hd.seg.SegmentationTypeValues
+            The segmentation_type parameter.
+
+        Returns
+        -------
+        pixel_array: np.ndarray
+            Input pixel array with the data type simplified if possible.
+        segments_overlap: hd.seg.SegmentationOverlaps
+            The value for the SegmentationOverlaps attribute, inferred from the
+            pixel array.
```
hackermd:
```suggestion
        pixel_array: numpy.ndarray
            The segmentation pixel array.
        described_segment_numbers: numpy.ndarray
            The segment numbers from the segment descriptions, in the order
            they were passed. 1D array of integers.
        segmentation_type: highdicom.seg.SegmentationTypeValues
            The segmentation_type parameter.

        Returns
        -------
        pixel_array: numpy.ndarray
            Input pixel array with the data type simplified if possible.
        segments_overlap: highdicom.seg.SegmentationOverlaps
            The value for the SegmentationOverlaps attribute, inferred from the
            pixel array.
```

src/highdicom/seg/sop.py
```diff
@@ -860,8 +746,229 @@ def add_segments(
...
+                'and monotonically increasing by 1.'
+            )
+        if described_segment_numbers[0] != 1:
+            raise ValueError(
+                'Segment descriptions should be numbered starting '
+                f'from 1. Found {described_segment_numbers[0]}. '
+            )
+
+    @staticmethod
+    def _check_pixel_array(
```
hackermd: This method does more than just checking the array. It also casts it to the correct data type if necessary. It may be beneficial for the name to reflect that.

```suggestion
    def _check_and_cast_pixel_array(
```

src/highdicom/seg/sop.py
```diff
@@ -860,8 +746,229 @@ def add_segments(
...
+
+        Returns
+        -------
+        pixel_array: np.ndarray
+            Pixel array with empty frames removed
+        plane_positions: Sequence[PlanePositionSequence]
+            Plane positions with entries corresponding to empty frames removed.
+        source_image_indices: List[int]
+            List giving for each frame in the output pixel array the index of
+            the corresponding frame in the original pixel array
```
hackermd:
```suggestion
        pixel_array: np.ndarray
            Segmentation pixel array
        plane_positions: Sequence[PlanePositionSequence]
            Plane positions for each of the frames

        Returns
        -------
        pixel_array: numpy.ndarray
            Pixel array with empty frames removed
        plane_positions: Sequence[highdicom.PlanePositionSequence]
            Plane positions with entries corresponding to empty frames removed.
        source_image_indices: List[int]
            List giving for each frame in the output pixel array the index of
            the corresponding frame in the original pixel array
```

hackermd (changes requested):
Found a couple of minor things mostly related to docstrings.

src/highdicom/seg/sop.py
CPBridge: resolved in fd1384d

src/highdicom/seg/sop.py
CPBridge: resolved in fd1384d

src/highdicom/seg/sop.py
CPBridge: resolved in fd1384d

src/highdicom/seg/sop.py
CPBridge: resolved in fd1384d

CPBridge:
@hackermd Note that I'm getting a bunch of mypy errors from the SR stuff

CPBridge:
@hackermd commit c268cda appears to have introduced a major regression. Many tests are now failing on comparisons between coded concepts

hackermd:
> @hackermd commit c268cda appears to have introduced a major regression. Many tests are now failing on comparisons between coded concepts

Sorry about that. Fixed with aec828a8ded1d0e3ee2a50ec5483800f8e31b34c

CPBridge:
@hackermd @cgorman  I just had an important realisation, thanks to my syntax highlighting...

`map` is a Python built-in. Therefore having a sub-package called `map` could be a really bad idea. 
Suggest we change to `pmap` before merging to master. Thoughts?

e.g. 
```python
from highdicom import map  #collision

map([1, 2, 3], lambda x: x + 1)  # uh oh
```

hackermd:
> `map` is a Python built-in. Therefore having a sub-package called `map` could be a really bad idea.
> Suggest we change to `pmap` before merging to master. Thoughts?

Yes, if imported that way, it can cause trouble. We chose `map` to allow for implementation of other "maps" under that namespace. However, it's unclear whether we will ever implement them and risking overriding a built-in symbol is not nice. Linters will also highlight it, which may be confusing. I would be fine with renaming the package to `pmaps` or `pm`.

cgorman:
> > `map` is a Python built-in. Therefore having a sub-package called `map` could be a really bad idea.
> > Suggest we change to `pmap` before merging to master. Thoughts?
> 
> Yes, if imported that way, it can cause trouble. We chose `map` to allow for implementation of other "maps" under that namespace. However, it's unclear whether we will ever implement them and risking overriding a built-in symbol is not nice. Linters will also highlight it, which may be confusing. I would be fine with renaming the package to `pmaps` or `pm`.

Ah yeah, I feel like you told me about this one time and I forgot... I vote for `pm` for consistency

hackermd:
@CPBridge renamed package to `pm` with ac711e32fe4c11ed242ea18a1d41841ca09fc2c5

src/highdicom/sr/templates.py
hackermd: What about the following?

* `def image_region(self) -> Union[ImageRegion, ImageRegion3D, None]`
* `def referenced_segment(self) -> Union[ReferencedSegment, ReferencedSegmentationFrame, None]`

src/highdicom/sr/templates.py
hackermd: At this point, I am reluctant to add more filters. Unless you really need this, I would (at least for now) focus on getting the filters right that we already implemented.

src/highdicom/sr/templates.py
hackermd: The `group_item` is just an individual `ContentItem`, but the `PlanarROIMeasurementsAndQualitativeEvaluations` is a `Template`, which is a `ContentSequence` containing one or more `ContentItem` objects.

src/highdicom/sr/templates.py
hackermd: Good catch! Fixed via 79aa9fa1d0b1e0cd48c371affb1bc2a9798041dd

src/highdicom/sr/templates.py
hackermd: Is this still the case or has this meanwhile been fixed via b3316a47b3a185133e331e6593b97447e2bd8725?

src/highdicom/sr/templates.py
hackermd: I think the problem with the finding site match has been fixed via b3316a47b3a185133e331e6593b97447e2bd8725.

I like the idea of passing a `FindingSite` object and thereby also considering the laterality modifier. Adding `__eq__` would be a convenient way to achieve that. Now that I think about it, we should probably override `__eq__` on most if not all objects in the `content` module(s).

src/highdicom/sr/templates.py
CPBridge: This is resolved

src/highdicom/sr/templates.py
CPBridge: The problem with the matching has been resolved.

src/highdicom/sr/templates.py
CPBridge: I think this could be very commonly used for ML projects, even if using the UID would arguably be better. The reason to add it now would be that if we wait until later, it will have to go to the end of the function signature for backwards compatibility, when it probably belongs next to tracking_uid

src/highdicom/sr/templates.py
CPBridge: Ah right...

src/highdicom/sr/templates.py
CPBridge: I'm not sure it makes sense to just ignore the parameter if they are used in an "and" fashion, because it would need to match both parameters to be returned, which is impossible. Which one would be ignored? Presumably under the current behaviour it would just return an empty list every time this combination of parameters is passed regardless of the contents of the SR, which is ok but an error message telling users that they've asked for something impossible seems potentially helpful? I'm ok with leaving as it is

hackermd (approved):

PR #96, opened 2021-08-06
Fixing "optionals" in docstings

### LINKED ISSUE #78: Documentation of None type for optional parameters
Currently, the `None` type is omitted from type annotations in the docstrings.

As discussed in https://github.com/MGHComputationalPathology/highdicom/pull/75#discussion_r643935439, we should consistently use `Union[..., None], optional` to clearly indicate that `None` is a valid option for these parameters.

```python
from typing import Optional


def foo(bar: int, optional_arg: Optional[str] = None) -> None:
    """

    Parameters
    -----------
    bar: int
        A required parameter
    optional_arg: Union[str, None], optional
        An optional parameter
    """
```

CPBridge:
Originally intended to fix #78, this PR now does a couple of general "spring clean" things that I noticed on the way.

- Implements the docstring changes in #78, specifically including the `None` in the docstrings (within a Union) for parameters that are both optional (have defaults) and Optional (can be `None`).
- Fixes a few other typos and errors in docstrings
- Introduces a new enum for the PatientSex attribute, accepts it (or str as before) wherever patient sex is passed and uses it to check the value of any str passed as patient sex. Previously, there were no checks on values passed as patient sex, so people could pass anything at all...
- Added docstrings to the individual attributes of enums where the meaning is not obvious from the name. Some of these were previously very cryptic, especially `PatientOrientationValuesBiped` and `PatientOrientationValuesQuadruped`, which use just a single letter but had no documentation previously.
- Removes the `GraphicTypesValues.ELLIPSOID`, which I believe should never have been there in the first place, and all references to it. This is only valid for GraphicTypeValues3D. This is a backwards incompatible change, but hopefully no-one was using this invalid value before...

Closes #78


src/highdicom/enum.py
```diff
@@ -29,9 +29,23 @@ class DimensionOrganizationTypeValues(Enum):
     TILED_SPARSE = 'TILED_SPARSE'
 
 
+class PatientSexValues(Enum):
+
+    """Enumerated values for Patient's Sex attribute."""
+
+    MALE = 'M'
+    FEMALE = 'F'
+    OTHER = 'O'
```
hackermd, line 38: For consistency with the other enums, I would suggesting using the abbreviations here, too (`M`, `F`, `O`).

src/highdicom/sc/sop.py
```diff
@@ -125,43 +126,43 @@ def __init__(
...
+        patient_id: Union[str, None], optional
            ID of the patient (medical record number)
-        patient_name: Optional[Union[str, PersonName]], optional
+        patient_name: Union[str, PersonName, None], optional
            Name of the patient
-        patient_birth_date: str, optional
+        patient_birth_date: Union[str, None], optional
            Patient's birth date
-        patient_sex: str, optional
+        patient_sex: Union[str, highdicom.enum.PatientSexValues, None], optional
```
hackermd, line 135: We may not want to include the "private" `enum` namespace in the docstring.

```suggestion
        patient_sex: Union[str, highdicom.PatientSexValues, None], optional
```

src/highdicom/sc/sop.py
```diff
@@ -125,43 +126,43 @@ def __init__(
...
+        accession_number: Union[str, None], optional
            Accession number of the study
-        study_date: Union[str, datetime.date], optional
+        study_date: Union[str, datetime.date, None], optional
            Date of study creation
-        study_time: Union[str, datetime.time], optional
+        study_time: Union[str, datetime.time, None], optional
            Time of study creation
-        referring_physician_name: Optional[Union[str, PersonName]], optional
+        referring_physician_name: Union[str, PersonName, None], optional
```
hackermd, line 145:
```suggestion
        referring_physician_name: Union[str, pydicom.valuerep.PersonName, None], optional
```

hackermd: I just realized that we forgot to expose the content of the `enum` module at the package level. I would suggest exposing the enum classes similar to those in the sub-packages (`seg`, `sr`). What do you think?

src/highdicom/seg/enum.py
```diff
@@ -34,6 +34,8 @@ class SpatialLocationsPreservedValues(Enum):
     YES = 'YES'
     NO = 'NO'
     REORIENTED_ONLY = 'REORIENTED_ONLY'
+    """A projection radiograph that has been flipped, and/or rotated by a
```
hackermd, line 37: I am wondering to which extent we should document individual fields or rather provide a reference to the relevant section of the standard at the class level. What do you think?

src/highdicom/seg/sop.py
```diff
@@ -199,15 +199,15 @@ def __init__(
...
             Type of fractional segmentation that indicates how pixel data
             should be interpreted
         max_fractional_value: int, optional
             Maximum value that indicates probability or occupancy of 1 that
             a pixel represents a given segment
-        content_description: str, optional
+        content_description: Union[str, None], optional
             Description of the segmentation
-        content_creator_name: Optional[Union[str, PersonName]], optional
+        content_creator_name: Union[str, PersonName, None], optional
```
hackermd, line 210:
```suggestion
        content_creator_name: Union[str, pydicom.valuerep.PersonName, None], optional
```

src/highdicom/seg/sop.py
```diff
@@ -216,16 +216,16 @@ def __init__(
             FRACTIONAL segmentation type:
             RLE Lossless (``"1.2.840.10008.1.2.5"``) and
             JPEG 2000 Lossless (``"1.2.840.10008.1.2.4.90"``).
-        pixel_measures: PixelMeasures, optional
+        pixel_measures: Union[PixelMeasures, None], optional
```
hackermd, line 219:
```suggestion
        pixel_measures: Union[highdicom.PixelMeasures, None], optional
```

src/highdicom/seg/enum.py
hackermd: These are all terms defined by the standard and I am concerned that we are duplicating a lot of standard text, which is continuously updated.

src/highdicom/sr/templates.py
```diff
@@ -3066,7 +3066,7 @@ def __init__(
             Number of rows in pixel data frames
         pixel_data_columns: int
             Number of rows in pixel data frames
-        additional_descriptors: Sequence[highdicom.sr.value_types.ContentItem], optional
+        additional_descriptors: Union[Sequence[highdicom.sr.value_types.ContentItem], None], optional
```
hackermd, line 3069: Should we also expose `ContentItem` at the package level? Would it instead make sense to use a `Union` and potentially even restrict the set of value types?

src/highdicom/sr/templates.py
```diff
@@ -3582,21 +3582,21 @@ def get_volumetric_roi_measurement_groups(
...
+        tracking_uid: Union[str, None], optional
             Unique tracking identifier
-        finding_type: Union[highdicom.sr.CodedConcept, pydicom.sr.coding.Code], optional
+        finding_type: Union[highdicom.sr.CodedConcept, pydicom.sr.coding.Code, None], optional
             Finding
-        finding_site: Union[highdicom.sr.CodedConcept, pydicom.sr.coding.Code], optional
+        finding_site: Union[highdicom.sr.CodedConcept, pydicom.sr.coding.Code, None], optional
             Finding site
-        reference_type: Union[highdicom.sr.GraphicTypeValues, highdicom.sr.GraphicTypeValues3D], optional
+        reference_type: Union[highdicom.sr.GraphicTypeValues, highdicom.sr.GraphicTypeValues3D, None], optional
```
hackermd, line 3591: This was actually wrong

```suggestion
        reference_type: Union[highdicom.sr.CodedConcept, pydicom.sr.coding.Code, None], optional
```

src/highdicom/sr/templates.py
```diff
@@ -3582,21 +3582,21 @@ def get_volumetric_roi_measurement_groups(
...
             Type of referenced ROI
-        graphic_type: highdicom.sr.GraphicTypeValues3D, optional
+        graphic_type: Union[highdicom.sr.GraphicTypeValues3D, None], optional
```
hackermd, line 3593: This was also wrong

```suggestion
        graphic_type: Union[highdicom.sr.GraphicTypeValues, highdicom.sr.GraphicTypeValues3D, None], optional
```

hackermd (changes requested):
Thanks @CPBridge for this heroic effort!

I left a few suggestions to fixed a couple of things that I stumbled over when reviewing. In addition, I made a comment regarding the documentation of individual enumerators.

src/highdicom/enum.py
CPBridge: Hmm. I was going with the precedent of the enum immediately above using `THREE_DIMENSIONAL = '3D'`, but now I think about it I assume that was because beginning a variable with a number is disallowed, and not for clarity. Personally I think using clearer enums would lead to clearer code and fewer mistakes but I guess the ship has sailed now. 

Resolved in 3525c76

src/highdicom/sc/sop.py
CPBridge: As noted below, this is because the enum sub-package is not exposed at the root of the package

src/highdicom/sc/sop.py
CPBridge: Resolved in 9a6f15f

src/highdicom/sc/sop.py
CPBridge: Resolved in 9a6f15f

src/highdicom/sc/sop.py
CPBridge: Exposed enum at the root as suggested in 9a6f15f

src/highdicom/sr/templates.py
CPBridge: It's very weird that this was not previously exposed before... Think it must have been an oversight. I have now exposed it at the level of the `sr` sub-package, in line with its subclasses, and updated the docstrings which all previously used the full path. (fc2fb3f)

Regarding limiting the value types, quite possibly. But I think this is best left outside the scope of this PR. I think there needs to be further work on the ImageLibrary in general.

src/highdicom/seg/enum.py
CPBridge: I would argue that documenting the individual fields makes the documentation much more accessible and useful (check out how it appears in the rendered pages compared to before). Further although the standard changes often, these enums are very unlikely to change, and if they do change, they will change in a backwards-compatible way so nothing we put here will be invalidated. If new  fields are added, which may well happen, we will need to update the code anyway...

For me it's a no-brainer for the enums with very simple meanings that are completely obscured by the abbreviations used (e.g. 'F' for 'FOOT' in the `PatientOrientationValues`). With some of the more complicated meanings (e.g. value types and relationship types), I can see an argument for simplifying the docstrings and pointing to the standard. However in these cases I have already provided a link to the standard. Are there particular enums where you think we should remove the docstrings?

CPBridge:
@hackermd Thanks for looking through this carefully. I have addressed your comments above.

src/highdicom/sr/enum.py
```diff
@@ -4,77 +4,199 @@
...
+    """Reference to UIDs of Image Composite SOP Instances."""
+
     NUM = 'NUM'
+    """Numeric value and associated Unit of Measurement."""
+
     PNAME = 'PNAME'
+    """Name of person."""
+
     SCOORD = 'SCOORD'
+    """Listing of spatial coordinates."""
```
hackermd, line 39:
```suggestion
    """Listing of spatial coordinates defined in 2D pixel matrix."""
```

src/highdicom/sr/enum.py
```diff
@@ -4,77 +4,199 @@
...
+
     SCOORD3D = 'SCOORD3D'
+    """Listing of spatial 3D coordinates."""
```
hackermd, line 42:
```suggestion
    """Listing of spatial coordinates defined in 3D frame of reference."""
```

src/highdicom/sr/enum.py
```diff
@@ -4,77 +4,199 @@
...
+    """Enumerated values for attribute Graphic Type.
+
+    See :dcm:`C.18.6.1.1 <part03/sect_C.18.6.html#sect_C.18.6.1.1>`.
+
+    """
 
     CIRCLE = 'CIRCLE'
+    """A circle defined by two (column,row) pairs. The first point is the
+    central pixel. The second point is a pixel on the perimeter of the
+    circle."""
```
hackermd, line 71:
```suggestion
    """A circle defined by two (column,row) pairs.
    
    The first pair is the central point and
    the second pair is a point on the perimeter of the circle.
    """
```

hackermd: Let's try to get towards [PEP 257](https://www.python.org/dev/peps/pep-0257/) compliance and use a single-line summary at the top of a longer docstring.

src/highdicom/sr/enum.py
```diff
@@ -4,77 +4,199 @@
...
+
     ELLIPSE = 'ELLIPSE'
-    ELLIPSOID = 'ELLIPSOID'
+    """An ellipse defined by four pixel (column,row) pairs, the first two
+    points specifying the endpoints of the major axis and the second two points
+    specifying the endpoints of the minor axis of an ellipse."""
```
hackermd, line 76:
```suggestion
    """An ellipse defined by four pixel (column,row) pairs.
    
    The first two pairs specify the endpoints of the major axis and
    the second two pairs specify the endpoints of the minor axis.
    """
```

src/highdicom/sr/enum.py
```diff
@@ -4,77 +4,199 @@
...
     MULTIPOINT = 'MULTIPOINT'
+    """Multiple pixels each denoted by an (column,row) pair."""
+
     POINT = 'POINT'
+    """A single pixel denoted by a single (column,row) pair."""
+
     POLYLINE = 'POLYLINE'
+    """A series of connected line segments with ordered vertices denoted by
+    (column,row) pairs; if the first and last vertices are the same it is a
+    closed polygon."""
```
hackermd, line 87:
```suggestion
    """Connected line segments with vertices denoted by (column,row) pairs.
    
    If the first and last pairs are the same it is a closed polygon.
    """
```

src/highdicom/sr/enum.py
```diff
@@ -4,77 +4,199 @@
...
+    """Enumerated values for attribute Graphic Type 3D.
+
+    See :dcm:`C.18.9.1.2 <part03/sect_C.18.9.html#sect_C.18.9.1.2>`.
+
+    """
 
     ELLIPSE = 'ELLIPSE'
+    """An ellipse defined by four (x,y,z) triplets, the first two triplets
+    specifying the endpoints of the major axis and the second two triplets
+    specifying the endpoints of the minor axis."""
```
hackermd, line 101:
```suggestion
    """An ellipse defined by four (x,y,z) triplets.
    
    The first two triplets specify the endpoints of the major axis and
    the second two triplets specify the endpoints of the minor axis.
    """
```

src/highdicom/sr/enum.py
```diff
@@ -4,77 +4,199 @@
...
+
     ELLIPSOID = 'ELLIPSOID'
+    """A three-dimensional geometric surface whose plane sections are either
+    ellipses or circles and contains three intersecting orthogonal axes, "a",
+    "b", and "c"; the ellipsoid is defined by six (x,y,z) triplets, the first
+    and second triplets specifying the endpoints of axis "a", the third and
+    fourth triplets specifying the endpoints of axis "b", and the fifth and
+    sixth triplets specifying the endpoints of axis "c"."""
```
hackermd, line 109:
```suggestion
    """A three-dimensional geometric surface defined by six (x,y,z) triplets.
    
    The plane sections of the surface are either ellipses or circles and
    the surface contains three intersecting orthogonal axes:
    "a", "b", and "c".
    The first and second triplets specify the endpoints of axis "a",
    the third and fourth triplets specify the endpoints of axis "b", and
    the fifth and sixth triplets specify the endpoints of axis "c".
    """
```

src/highdicom/sr/enum.py
```diff
@@ -4,77 +4,199 @@
...
+
     MULTIPOINT = 'MULTIPOINT'
+    """Multiple locations each denoted by an (x,y,z) triplet; the points need
+    not be coplanar."""
```
hackermd, line 113:
```suggestion
    """Multiple points each denoted by an (x,y,z) triplet.
    
    The points need not be coplanar.
    """
```

src/highdicom/sr/enum.py
```diff
@@ -4,77 +4,199 @@
...
+
     POINT = 'POINT'
+    """A single location denoted by a single (x,y,z) triplet."""
```
hackermd, line 116:
```suggestion
    """An individual point denoted by a single (x,y,z) triplet."""
```

src/highdicom/sr/enum.py
```diff
@@ -4,77 +4,199 @@
...
+
     POLYLINE = 'POLYLINE'
+    """A series of connected line segments with ordered vertices denoted by
+    (x,y,z) triplets; the points need not be coplanar."""
```
hackermd, line 120:
```suggestion
    """Connected line segments with vertices denoted by (x,y,z) triplets.
    
    The points need not be coplanar.
    """
```

src/highdicom/sr/enum.py
```diff
@@ -4,77 +4,199 @@
...
+
     POLYGON = 'POLYGON'
+    """A series of connected line segments with ordered vertices denoted by
+    (x,y,z) triplets, where the first and last vertices shall be the same
+    forming a polygon; the points shall be coplanar."""
```
hackermd, line 125:
```suggestion
    """Connected line segments with vertices denoted by (x,y,z) triplets.
    
    The first and last triplets shall be the same forming a closed polygon.
    The points shall be coplanar.
    """
```

src/highdicom/sr/enum.py
```diff
@@ -4,77 +4,199 @@
...
-    """Enumerated values for attribute Temporal Range Type."""
+    """Enumerated values for attribute Temporal Range Type.
+
+    See :dcm:`C.18.7.1.1 <part03/sect_C.18.7.html#sect_C.18.7.1.1>`.
+
+    """
 
     BEGIN = 'BEGIN'
+    """A range beginning at one temporal point, and extending beyond the end of
+    the acquired data."""
```
hackermd, line 138:
```suggestion
    """A range that begins at the identified temporal point.
    
    It extends beyond the end of the acquired data.
    """
```

src/highdicom/sr/enum.py
```diff
@@ -4,77 +4,199 @@
...
+
     END = 'END'
+    """A range beginning before the start of the acquired data, and extending
+    to (and including) the identified temporal point."""
```
hackermd, line 142:
```suggestion
    """A range that ends at the identified temporal point.
    
    It begins before the start of the acquired data and
    extends to (and includes) the identified temporal point. 
    """
```

src/highdicom/sr/enum.py
```diff
@@ -4,77 +4,199 @@
...
+    """Description of properties of the Source Content Item."""
+
     INFERRED_FROM = 'INFERRED FROM'
+    """Source Content Item conveys a measurement or other inference made from
+    the Target Content Items. Denotes the supporting evidence for a measurement
+    or judgment."""
+
     SELECTED_FROM = 'SELECTED FROM'
+    """Source Content Item conveys spatial or temporal coordinates selected
+    from the Target Content Item(s)."""
```
hackermd, line 191: Shouldn't we talk about `parent` and `child` instead of `source` and `target`? In my opinion, this would be easier to understand. Thoughts?

hackermd: Could you please also format the docstrings such that they start with a single-line summary.

src/highdicom/sr/enum.py
```diff
@@ -4,77 +4,199 @@
...
 
 class PixelOriginInterpretationValues(Enum):
 
     """Enumerated values for attribute Pixel Origin Interpretation."""
 
     FRAME = 'FRAME'
+    """Relative to the individual frame."""
+
     VOLUME = 'VOLUME'
+    """Relative to the Total Image Matrix (for tiled images)."""
```
hackermd, line 202: The term `"VOLUME"` is very confusing in this context. Attempt to clarify:

```suggestion
    """Relative to the Total Pixel Matrix of the VOLUME image."""
```

hackermd (changes requested):
Added a couple of suggestions to reformat docstrings for improved compliance with [PEP 257](https://www.python.org/dev/peps/pep-0257/).

src/highdicom/sr/enum.py
CPBridge: Resolved in 52aa0b0

CPBridge:
@hackermd I have addressed your comments

src/highdicom/sr/enum.py
CPBridge: (p.s. source and target are the terms used in the standard... but I agree they are confusing and switched to parent and child)

PR #97, opened 2021-08-07
Alternative SCImage Constructor

CPBridge:
The current constructor for the SCImage requires several pieces of patient or study related metadata. Often in practice, this information is simply copied from another dataset in the study in user code, but this process is annoying and verbose. I have now done this in several places personally.

This PR introduces a new alternative constructor for SCImage called `from_study_dataset` (open to other suggestions for the name), that accepts another dataset from the study and copies the patient and study related metadata from there. This should make user code more concise and less error prone for this very common situation. Also added tests.

### DIFF
--- a/src/highdicom/sc/sop.py
+++ b/src/highdicom/sc/sop.py
@@ -367,3 +367,158 @@ class SCImage(SOPClass):
             self.PixelData = encapsulate([encoded_frame])
         else:
             self.PixelData = encoded_frame
+
+    @classmethod
+    def from_study_dataset(
+        cls,
+        study_dataset: Dataset,
+        pixel_array: np.ndarray,
+        photometric_interpretation: Union[
+            str,
+            PhotometricInterpretationValues
+        ],
+        bits_allocated: int,
+        coordinate_system: Union[str, CoordinateSystemNames],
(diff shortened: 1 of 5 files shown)


src/highdicom/sc/sop.py
```diff
@@ -367,3 +367,158 @@ def __init__(
             self.PixelData = encapsulate([encoded_frame])
         else:
             self.PixelData = encoded_frame
+
+    @classmethod
+    def from_study_dataset(
```
hackermd, line 372: I find the name `study_dataset` confusing, since a data set encodes an instance. Can we call it `reference_dataset` or `ref_dataset` instead?

hackermd (commented):
Nice! That's handy. Just one comment regarding the name of the method and its parameter.

src/highdicom/sc/sop.py
CPBridge: Totally agree. I knew `study_dataset` wasn't great but was drawing a blank on something better and was hoping you could suggest one. Changed the method and its parameter to `ref_dataset` as suggested in 5cce71d.

hackermd (approved):

PR #102, opened 2021-09-08
Provide image orientation and position for segmentation of slide microscopy image

hackermd:


src/highdicom/seg/sop.py
```diff
@@ -535,6 +535,33 @@ def __init__(
...
+            else:
+                plane_position_values = np.array([
+                    (
+                        float(pos[0].XOffsetInSlideCoordinateSystem),
+                        float(pos[0].YOffsetInSlideCoordinateSystem),
+                        float(pos[0].ColumnPositionInTotalImagePixelMatrix),
+                        float(pos[0].RowPositionInTotalImagePixelMatrix),
+                    )
+                    for pos in plane_positions
+                ])
```
CPBridge, line 553: I don't think this can be correct. Firstly it's redundant, since `plane_position_value` is calculated a few lines down. Secondly, why do you not you not use? `DimensionIndexSequence.get_index_values()`. Thirdly, what if the first frame is omitted due to being empty, would that not make the value calculated here wrong?

CPBridge (changes requested):

src/highdicom/seg/sop.py
hackermd: > Firstly it's redundant, since plane_position_value is calculated a few lines down.

We need to compute it here before empty frames will be removed, but you are right that we shouldn't compute it twice. I will instead compute it only once before frames are being removed and subsequently use `source_image_indices` to select the remaining subset.

> Secondly, why do you not you not use DimensionIndexSequence.get_index_values()?

You are right. I should re-use that.

> Thirdly, what if the first frame is omitted due to being empty, would that not make the value calculated here wrong?

No, that's the whole point of why I am doing this. The total pixel matrix includes the empty frames that will be omitted from the data set and we need to compute the offset of the first frame before any frames will be removed. Otherwise, the image position may be wrong.

hackermd: Addressed via e7d28243432c1ae47d12a36a8fdfdbdd84ab65ab

hackermd: cc @dclunie for awareness

CPBridge: Ok I think this makes sense

CPBridge:
I would like to go through this more carefully because this stuff is really complicated.  However in the meantime @hackermd, I notice that there is a test failing?

hackermd:
> I would like to go through this more carefully because this stuff is really complicated.

Very complicated indeed. I also had a separate discussion with @dclunie on how to improve the standard (not directly affecting this PR).

> However in the meantime @hackermd, I notice that there is a test failing?

I think these are only the mypy tests, which are currently supposed to fail due to the type hint issues with pydicom. Or did I miss a pytest test?

CPBridge:
I thought there was a pytest test failing but it appears I imagined it, sorry

CPBridge (approved):
Sorry for the delay. Went through it carefully and looks good to me!

(Also just pushed a commit fixing a missing newline at end of file)

PR #103, opened 2021-09-09
Fix a typo in one of the keyword used

malaterre:
s/FrameVolumeBasedCalculationTechnique/VolumeBasedCalculationTechnique/

ref:
http://dicom.nema.org/medical/dicom/current/output/chtml/part06/chapter_6.html#para_676396af-e8dd-43d2-8cee-1a72701c8b6e


hackermd (approved):
Good catch! Thank you for fixing the bug.

PR #104, opened 2021-09-10
Further SR parsing functionality for referenced segments

CPBridge:
Currently there is the `.roi` property to access the referenced region of a measurement group when it is given "by value".

However, there is no convenient way to do this for regions referenced "by reference", i.e. those which refer to segments in segmentation images.

This PR adds several properties to access information about referenced segments in an intuitive way.

Note that I decided not to return objects of type `ReferencedSegment` or `ReferencedSegmentationFrame` directly as this got a bit messy. We would have to find the referenced segment/segmentation frame content item and the relevant source image/series item(s) and piece back together into the highdicom `ReferencedSegment` or `ReferencedSegmentationFrame`. Instead there are separate properties to access information about the referenced segment and the source images from which it is derived. Let me know what you think about this.


src/highdicom/sr/templates.py
```diff
@@ -2918,13 +2925,65 @@ def roi(self) -> Union[ImageRegion, ImageRegion3D, None]:
...
+            logger.warning(
+                'found more than one "Source Image For Segmentation" content '
+                'item in "Planar ROI Measurements and Qualitative Evaluations" '
+                'template'
+            )
+        elif len(matches) == 0:
+            return None
+
+        return SourceImageForSegmentation.from_dataset(matches[0])
+
```
hackermd, line 2986: I think it would be cleaner to return `Union[hd.seg.ReferencedSegment, hd.seg.ReferencedSegmentationFrame]` so that the segmentation and corresponding source images are referenced together and match the constructor arguments.

src/highdicom/sr/templates.py
```diff
@@ -2918,13 +2925,65 @@ def roi(self) -> Union[ImageRegion, ImageRegion3D, None]:
...
+        if len(matches) > 1:
+            logger.warning(
+                'found more than one "Referenced Segmentation Frame" content '
+                'item in "Planar ROI Measurements and Qualitative Evaluations" '
+                'template'
+            )
+        elif len(matches) == 0:
+            return None
+
+        return ImageContentItem.from_dataset(matches[0])
```
hackermd, line 2959: Should they already be of type `ImageContentItem`?

hackermd:
> Note that I decided not to return objects of type ReferencedSegment or ReferencedSegmentationFrame directly as this got a bit messy. We would have to find the referenced segment/segmentation frame content item and the relevant source image/series item(s) and piece back together into the highdicom ReferencedSegment or ReferencedSegmentationFrame. Instead there are separate properties to access information about the referenced segment and the source images from which it is derived. Let me know what you think about this.

I would prefer piecing it together even if it gets a bit messy internally.

src/highdicom/sr/templates.py
```diff
@@ -247,19 +255,68 @@ def _get_roi_reference_type(
...
     for item in group_item.ContentSequence:
-        for concept_name in allowed_reference_types:
-            found_concept = CodedConcept.from_dataset(
-                item.ConceptNameCodeSequence[0]
-            )
-            if found_concept == concept_name:
-                return concept_name
-    return None
+        rel_type = RelationshipTypeValues(item.RelationshipType)
+        if rel_type != RelationshipTypeValues.CONTAINS:
```
hackermd, line 285: Could we now directly use `item.relationship_type`?

```suggestion
        if item.relationship_type != RelationshipTypeValues.CONTAINS:
```

src/highdicom/sr/templates.py
```diff
@@ -247,19 +255,68 @@ def _get_roi_reference_type(
...
+            # All ROI reference content items have relationship type CONTAINS
+            continue
+
+        found_concept = CodedConcept.from_dataset(
+            item.ConceptNameCodeSequence[0]
+        )
+        if found_concept in allowed_reference_types:
+            expected_value_types = ref_type_value_type_map[found_concept]
+            found_value_type = ValueTypeValues(item.ValueType)
+            if found_value_type in expected_value_types:
```
hackermd, line 295:
```suggestion
            if item.value_type in expected_value_types:
```

src/highdicom/sr/templates.py
```diff
@@ -385,11 +442,61 @@ def _contains_uidref_items(
...
+    return False
+
+
 def _contains_scoord_items(
     parent_item: ContentItem,
     name: Union[Code, CodedConcept],
     graphic_type: Optional[GraphicTypeValues] = None,
-    relationship_type: Optional[RelationshipTypeValues] = None
+    relationship_type: Optional[RelationshipTypeValues] = None,
+    source_image_sop_instance_uid=None
```
hackermd, line 499: Type hint is missing

```suggestion
    source_image_sop_instance_uid: Optional[str] = None
```

src/highdicom/sr/templates.py
```diff
@@ -385,11 +442,61 @@ def _contains_uidref_items(
...
 
 
+def _contains_image_items(
+    parent_item: ContentItem,
+    name: Union[Code, CodedConcept],
+    referenced_sop_class_uid: Union[str, None] = None,
+    referenced_sop_instance_uid: Union[str, None] = None,
+    relationship_type: Optional[RelationshipTypeValues] = None
+) -> bool:
+    """Checks whether an item contains a specific item with value type IMAGE.
```
hackermd, line 452: I've been trying to switch to using this form as specified by [PEP 0257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings):

```suggestion
    """Check whether an item contains a specific item with value type IMAGE.
```

src/highdicom/sr/templates.py
```diff
@@ -419,10 +528,17 @@ def _contains_scoord_items(
...
+                continue
+        if source_image_sop_instance_uid is not None:
+            if not _contains_image_items(
+                item,
+                name=codes.DCM.OriginalSource,
+                referenced_sop_instance_uid=source_image_sop_instance_uid,
+                relationship_type=RelationshipTypeValues.SELECTED_FROM
+            ):
+                continue
+        return True
```
hackermd, line 541: Is this the desired behaviour?

src/highdicom/sr/templates.py
```diff
@@ -3889,6 +4004,16 @@ def get_planar_roi_measurement_groups(
...
         # FIXME: reference type and referenced sop instance uid
 
+        if graphic_type is not None:
+            if not isinstance(
+                graphic_type,
+                (GraphicTypeValues, GraphicTypeValues3D)
+            ):
+                raise TypeError(
+                    'graphic_type must be of type GraphicTypeValues or '
+                    'GraphicTypeValues3D, or None.'
```
hackermd, line 4014: For improved consistency with the style of other error messages:

```suggestion
                    'Argument "graphic_type" must be of type GraphicTypeValues, '
                    'GraphicTypeValues3D or None.'
```

src/highdicom/sr/templates.py
```diff
@@ -3889,6 +4004,16 @@ def get_planar_roi_measurement_groups(
         """  # noqa: E501
         # FIXME: reference type and referenced sop instance uid
```
hackermd, line 4005: Can this comment be removed?

src/highdicom/sr/templates.py
```diff
@@ -4018,6 +4208,16 @@ def get_volumetric_roi_measurement_groups(
...
```
hackermd, line 4209: Can this comment be removed?

hackermd (changes requested):
Nice work! This all looks very reasonable, but this stuff is quite complex. We should probably add a bunch of unit tests to cover all (most) conditional blocks to prevent future regressions. We may end up refactoring the implementation quite a bit moving forward to improve query performance or add additional query parameters.

CPBridge:
Sorry @hackermd, have been making steady progress on this as my time allows but did not intend this to be reviewed again just yet. Major outstanding issue is additional unit tests as you say, we definitely need some thorough tests for this functionality or it is guaranteed to break in the future. Also there are some further invalid combinations of reference_type and graphic_type that it would be good to warn users are impossible. I'll let you know when it's done

hackermd:
@CPBridge you could convert the PR into a draft for the time being.

src/highdicom/sr/templates.py
CPBridge: After refactoring, this now returns a `Union[ReferencedSegmentationFrame, None]`. Note that a `PlanarROIMeasurementsAndQualitativeEvaluations` cannot contain a "Referenced Segment" (they can only be used in the volumetric measurements group)

src/highdicom/sr/templates.py
CPBridge: Removed in 4fc6386

src/highdicom/sr/templates.py
CPBridge: Removed in 4fc6386

src/highdicom/sr/templates.py
CPBridge: Yes, I was copying some of the existing code, which does not assume that content items have been converted to their highdicom types. I guess it was written before the "eager" type conversion was implemented and could now be simplified.

src/highdicom/sr/templates.py
CPBridge: Resolved in bd0f5d3

src/highdicom/sr/templates.py
CPBridge: Resolved in bd0f5d3

src/highdicom/sr/templates.py
CPBridge: Resolved in bd0f5d3

src/highdicom/sr/templates.py
CPBridge: This method is actually no longer needed and was removed in bd0f5d3

CPBridge:
This is now complete @hackermd, including with new properties for the `VolumeSurface` class and a pretty comprehensive set of tests.

One question I have is what a good property name would be for what is currently called `VolumeSurface.graphic_data_items`, which returns the list of `Scoord3DContentItems` that actually make up the surface. Would be open to other names but this was the best I came up with at the time.

hackermd:
> One question I have is what a good property name would be for what is currently called VolumeSurface.graphic_data_items, which returns the list of Scoord3DContentItems that actually make up the surface. Would be open to other names but this was the best I came up with at the time.

I'd just call it `graphic_data` to match the name of the constructor parameter.

CPBridge:
I initially called it `graphic_data_items` since it returned the content items, not just the bare numpy arrays like the constructor parameter. But I don't think there's really a need for that so have changed to a `graphic_data` property that returns a list of arrays, just like the constructor (1f40735)

src/highdicom/sr/content.py
```diff
@@ -853,6 +855,181 @@ def __init__(
...
+        """List[highdicom.sr.Scoord3DContentItem]:
+            Graphic data elements that comprise the volume surface.
+        """
+        return cast(
+            List[Scoord3DContentItem],
+            self._lut[codes.DCM.VolumeSurface]
+        )
+
+    @property
+    def graphic_data(self) -> List[np.ndarray]:
```
hackermd, line 1026: I am wondering whether this should be `Union[numpy.ndarray, List[numpy.ndarray]]` to a) match the constructor argument and b) provide a single set of coordinates in case of `ELLIPSOID`.

What do you think?

hackermd (approved):

src/highdicom/sr/content.py
CPBridge: Generally I try to avoid having dynamically-determined returned types. But I think here maybe it makes sense and does match the input signature for the constructor. Implemented in 9ec6836

hackermd (approved):

PR #105, opened 2021-09-13
Properly set Pixel Presentation

malaterre:
Pixel Presentation != Pixel Representation

For reference:

* http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.16.2.html#table_C.8-131


fedorov:
I believe this has been addressed in #34.

CPBridge:
Thanks @malaterre !

> I believe this has been addressed in #34.

I would not be opposed to merging this as a quick fix nonetheless, in the meantime

CPBridge (approved):

PR #107, opened 2021-09-14
Content items of special templates lack relationship

hackermd:
Content items of acquisition and protocol templates don't have a Relationship field. We consider these items to be similar to CONTAINER items at the root of a SR document content tree and instantiate `ContentSequence` objects with `is_root` to allow content items without the `RelationshipType` attribute.


CPBridge (approved):

CPBridge:
Looks good to me!

PR #112, opened 2021-09-21
Make ImageRegion3Ds invalid for volumetric measurements group

CPBridge:
As was looking through the definition of [TID 1411](http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1411) "Volumetric ROI Measurements and Qualitative Evaluations" I noticed that in fact it is not part of the template to allow Image Region items to use SCOORD3D. In TID 1410 (Planar...), this is allowed. However in TID 1411 the only way to specify points in the 3D frame of reference is via "Volume Surface" items and not "Image Region" items.

Unfortunately, highdicom currently allows the user to pass a sequence of `ImageRegion3D` objects (which have value type SCOORD3D) items to the constructor of `VolumetricROIMeasurementsAndQualitative Evaluations`, which would give rise to an invalid instantiation of the template. This PR removes this option in the type signatures and docstrings, and enforces it via runtime checks.

However there is a further, uglier issue: the template actually uses "Volume Surface" items with multiplicity 1-n. This is the only way to include multiple planes of interest using SCOORD3D (since Image Regions using SCOORD3D are not allowed). However, only a single `VolumeSurface` is supported by the constructor. I have not yet attempted to fix this part of the issue. We could change the `referenced_volume_surface` from `Optional[VolumeSurface]` to `Optional[Sequence[VolumeSurface]]` but this would be backwards incompatible. Do you think we should allow both, i.e. (`Optional[Union[VolumeSurface, Sequence[VolumeSurface]]]`) for a few releases? @hackermd

### DIFF
--- a/src/highdicom/sr/templates.py
+++ b/src/highdicom/sr/templates.py
@@ -2967,7 +2967,7 @@ class VolumetricROIMeasurementsAndQualitativeEvaluations(
         self,
         tracking_identifier: TrackingIdentifier,
         referenced_regions: Optional[
-            Union[Sequence[ImageRegion], Sequence[ImageRegion3D]]
+            Union[Sequence[ImageRegion]]
         ] = None,
         referenced_volume_surface: Optional[VolumeSurface] = None,
         referenced_segment: Optional[
@@ -2992,7 +2992,7 @@ class VolumetricROIMeasurementsAndQualitativeEvaluations(
         ----------
         tracking_identifier: highdicom.sr.TrackingIdentifier
             identifier for tracking measurements
-        referenced_regions: Union[Sequence[highdicom.sr.ImageRegion], Sequence[highdicom.sr.ImageRegion3D], None], optional
+        referenced_regions: Union[Sequence[highdicom.sr.ImageRegion], None], optional
             regions of interest in source image(s)
         referenced_volume_surface: Union[highdicom.sr.VolumeSurface, None], optional
             volume of interest in source image(s)
@@ -3031,6 +3031,16 @@ class VolumetricROIMeasurementsAndQualitativeEvaluations(
         together with the corresponding source image(s) or series.
 
         """  # noqa: E501
+        if referenced_regions is not None and any(
+            isinstance(r, ImageRegion3D) for r in referenced_regions
+        ):
+            raise TypeError(
+                'Including items of type ImageRegion3D in "referenced_regions" '
+                'is invalid within a volumetric ROI measurement group is '
+                'invalid. To specify the referenced region in 3D frame of '
+                'reference coordinates, use the "referenced_volume_surface" '
+                'argument instead.'
+            )
         super().__init__(
             measurements=measurements,
             tracking_identifier=tracking_identifier,


CPBridge:
Three further issues I have noticed:

- The "Source Image For Segmentation" items that accompany a volume surface are at the top level of the measurements group according to the template, however highdicom currently puts them in the Content Sequence of the Volume Surface item.
- It should also be permissible to use "Source *Series* for Segmentation" instead of "Source Image for Segmentation" when using a Volume Surface, but this is not currently allowed by highdicom.
- Volume Surfaces can have graphic type `ELLIPSOID`, `POINT`, `POLYGON` or `ELLIPSE` (depending on how many items are present). Currently only `ELLIPSOID` is allowed by highdicom.

It seems that the way to fix all these properly is to re-base `VolumeSurface` on `ContentSequence` so that it can consist of multiple content items with "Volume Surface" and items of type "Source Image For Measurement" at the same level. This would match the pattern used by the `ReferencedSegment` class. Do you agree @hackermd?

hackermd:
> The "Source Image For Segmentation" items that accompany a volume surface are at the top level of the measurements group according to the template, however highdicom currently puts them in the Content Sequence of the Volume Surface item.

Good catch! We should fix this.

> It should also be permissible to use "Source Series for Segmentation" instead of "Source Image for Segmentation" when using a Volume Surface, but this is not currently allowed by highdicom.

Yes, we should support that, too.

> Volume Surfaces can have graphic type ELLIPSOID, POINT, POLYGON or ELLIPSE (depending on how many items are present). Currently only ELLIPSOID is allowed by highdicom.

I restricted it to the use of `ELLIPSOID` on purpose. There is a lot of confusion about the difference between "volumetric" and the SCOORD3D value type. SCOORD3D only specifies a geometric structure in a 3D coordinate system, which may or may not be volumetric. For the "Volume Surface" content item of the "Volumetric ROI Measurements and Qualitative Evaluations" template, I would argue that the geometry should indeed enclose a volume. However, a POINT, POLYGON, or ELLIPSE do not describe a volume.

> It seems that the way to fix all these properly is to re-base VolumeSurface on ContentSequence so that it can consist of multiple content items with "Volume Surface" and items of type "Source Image For Measurement" at the same level. This would match the pattern used by the ReferencedSegment class.

Agreed, that sounds like a good idea.

CPBridge:
> I would argue that the geometry should indeed enclose a volume. However, a POINT, POLYGON, or ELLIPSE do not describe a volume.

I would agree that this would be logical, but... that's not always the best criterion to apply.

If we disallow ImageRegion3D (which are invalid here as explained in my first comment above), and restrict VolumeSurfaces to be only `ELLIPSOID` it means that the only way that highdicom supports using 3D frame of reference coordinates (i.e. SCOORD3D) in the "Volumetric ROI Measurements and Qualitative Evaluations" is via Volume Surafces with graphic type `ELLIPSOID`. There would be no way to store arbitrary shapes or points in SCOORD3D as ROIs.

I feel that highdicom should provide a way to store coordinates of more general ROIs in SCOORD3D, and I suspect you would agree. I think that it is certainly a little confusing that this would be under a content item named "Volume Surface", but that still seems better than nothing.

Note also that POLYGONs and ELLIPSEs can only be used when there are multiple such items. My interpretation of this is that the volume is constructed "plane-by-plane" and then "joined" up to create a volume surface. A single ELLIPSE or POLYGON cannot describe a 3D surface and as such is disallowed by the standard. I think this is a little clumsy but logically consistent.

That leaves POINTs, which I can't really explain the logic of except that it is the infinitesimal limit of a volume surface. But, again, disallowing volume surfaces of type POINT would mean that there would be no way to store POINTs as SCOORD3D within a "Volumetric ROI Measurements and Qualitative Evaluations", and I feel that there should be a way (it will be important for highlighting anatomic landmarks for example). Alternatively we could disallow this and force users to use "Planar ROI Measurements and Qualitative Evaluations" to store points in SCOORD3D (as ImageRegion3Ds). However storing frame of reference coordinates in 3D within a planar measurements group feels incoherent in a different way.

In summary I would be tempted to allow the use of ELLIPSE and POLYGON graphic data as Volume Surfaces if and only if there are multiple such content items in the measurements group. I would also lean towards allowing POINT graphic data if and only if there is a single such content to cover the full range of the standard.

CPBridge:
@hackermd  I have now implemented this, and managed to do it in a backwards compatible way (provided users are not relying on `VolumeSurface` being a `ContentItem` subclass). I also added the ability to use POINTs, ELLIPSEs and POLYGONs following my reasoning above, but we can discuss further before merging.

I was also wrong about my second point above, `SourceSeriesForSegmentation` was always implemented but I managed to confuse myself somehow.

hackermd:
> If we disallow ImageRegion3D (which are invalid here as explained in my first comment above), and restrict VolumeSurfaces to be only ELLIPSOID it means that the only way that highdicom supports using 3D frame of reference coordinates (i.e. SCOORD3D) in the "Volumetric ROI Measurements and Qualitative Evaluations" is via Volume Surafces with graphic type ELLIPSOID. There would be no way to store arbitrary shapes or points in SCOORD3D as ROIs.

> I feel that highdicom should provide a way to store coordinates of more general ROIs in SCOORD3D, and I suspect you would agree. I think that it is certainly a little confusing that this would be under a content item named "Volume Surface", but that still seems better than nothing.

Those would not be "volumetric" and should instead be included in "Planar ROI Measurements and Evaluations". The distinction between SCOORD versus SCOORD3D and planar versus volumetric is really confusing. Ideally, the terms "2D" and "3D" should be avoided for that reason. Unfortunately, we just introduced them for [Annotation Coordinate Type](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.37.html#para_7efde114-af13-4d52-a03a-839c46c817af).

> That leaves POINTs, which I can't really explain the logic of except that it is the infinitesimal limit of a volume surface. But, again, disallowing volume surfaces of type POINT would mean that there would be no way to store POINTs as SCOORD3D within a "Volumetric ROI Measurements and Qualitative Evaluations", and I feel that there should be a way (it will be important for highlighting anatomic landmarks for example). Alternatively we could disallow this and force users to use "Planar ROI Measurements and Qualitative Evaluations" to store points in SCOORD3D (as ImageRegion3Ds). However storing frame of reference coordinates in 3D within a planar measurements group feels incoherent in a different way.

I disagree. There is nothing wrong about 3D coordinates defined in the frame of reference within a planar measurements group. The "3D" just refers to the coordinate system in which the coordinates are defined and is independent from the structure of the ROI. For example, a POLYGON can only be defined using 3D spatial coordinates, but the point must all be co-planar. In contrast, "planar" and "volumetric" refer to the type of measurements and evaluations, which depend on the structure of the ROI (planar POLYGON and volumetric ELLIPSOID).

Strictly speaking, POINTs should neither be in "Planar ROI Measurements and Evaluations" nor in "Volumetric ROI Measurements and Evaluations", since a point neither defines an area nor a volume and any non-zero "planar" or "volumetric" measurement of the point would be invalid. However, one could argue that the POINT is just the graphical annotation of the measured image region and the measurements do not have to be derived from the graphic defined by SCOORD or SCOORD3D.

What the standard leaves undefined is whether it would be valid to take a physical measurement (e.g., Square millimeter) of a ROI defined by 2D spatial coordinates in the Total Pixel Matrix in pixel unit. @fedorov @dclunie what are your thoughts on this?

> Note also that POLYGONs and ELLIPSEs can only be used when there are multiple such items. My interpretation of this is that the volume is constructed "plane-by-plane" and then "joined" up to create a volume surface. A single ELLIPSE or POLYGON cannot describe a 3D surface and as such is disallowed by the standard. I think this is a little clumsy but logically consistent.

Good point! Makes sense.

CPBridge:
OK in that case, the above is what is already implemented in this PR

src/highdicom/sr/content.py
```diff
@@ -750,9 +750,30 @@ def __init__(
...
+            If ``graphic_type`` is "ELLIPSE" or "POLYGON", the volume surface
+            will consist of two or more planar regions that together define the
+            surface. Therefore a list of two or more 2D numpy arrays should be
+            passed for ``graphic_data``.
+
+            Each 2D numpy array should have dimension N x 3 where each row of
+            the array represents a coordinate in the 3D frame of reference. The
+            number, N, and meaning of the coordinates depends upon the value of
+            "graphic_type". See :class:`highdicom.sr.GraphicTypeValues3D` for
+            details.
```
hackermd, line 775: Minor changes for consistency with other docstrings

```suggestion
            If `graphic_type` is ``"ELLIPSOID"`` or ``"POINT"``, the volume
            surface will consist of a single element that defines the entire
            surface. Therefore, a single 2D NumPy array should be passed
            as a list of length 1. For reasons of backwards compatibility,
            it is also possible to pass a NumPy array directly.

            If `graphic_type` is ``"ELLIPSE"`` or ``"POLYGON"``, the volume
            surface will consist of two or more planar regions that together
            define the surface. Therefore a list of two or more 2D NumPy
            arrays should be passed.

            Each 2D NumPy array should have dimension N x 3 where each row of
            the array represents a coordinate in the 3D Frame of Reference. The
            number, N, and meaning of the coordinates depends upon the value of
            `graphic_type`. See :class:`highdicom.sr.GraphicTypeValues3D` for
            details.
```

src/highdicom/sr/templates.py
```diff
@@ -2967,7 +2966,7 @@ def __init__(
         self,
         tracking_identifier: TrackingIdentifier,
         referenced_regions: Optional[
-            Union[Sequence[ImageRegion], Sequence[ImageRegion3D]]
+            Union[Sequence[ImageRegion]]
         ] = None,
         referenced_volume_surface: Optional[VolumeSurface] = None,
```
hackermd, line 2971: Conceptually, I would argue that SCOORD3D content items with POLYGON and ELLIPSE graphic type should rather be passed via `referenced_regions` ("Image Region") rather than `referenced_volume_surface` ("Volume Surface"), since the individual items do not actually represent volumes. This would also be in line with TID 1410 (Row 7b).
However, that's one of the many inconsistencies in TID 1500 that we'll probably just have to adopt. Thoughts?

hackermd (changes requested):
Thanks @CPBridge. Just one super-minor suggestion for a docstring and one question regarding SCOORD/SCOORD3D.

src/highdicom/sr/templates.py
CPBridge: Conceptually perhaps, but I think that deviation from the standard would likely cause more confusion than it would eliminate. Unless we dramatically changed things around, the user would still have to create a `VolumeSurface` object (to have the concept name correct) and then pass this to `referenced_regions` rather than `referenced_volume_surface` depending on the graphic type. We can't use the `ImageRegion3D` class because this has the wrong concept name. It would also cause an odd asymmetry with the decoding behaviour, and be very confusing for the (few) people that are familiar with the templates. I would vote to keep it as it is

hackermd (approved):

PR #113, opened 2021-09-22
Missing import of generate_uid from pydicom.uid

erikogabrielsson:


hackermd (approved):
Good catch! We should start including those code snippets into the tests.

We could consider using `hd.UID` instead.

PR #114, opened 2021-09-27
Implementation of Microscopy Bulk Simple Annotations

hackermd:


CPBridge:
@hackermd I think we probably need to resolve these:
- https://github.com/MGHComputationalPathology/highdicom/pull/114#discussion_r728587939
- https://github.com/MGHComputationalPathology/highdicom/pull/114#discussion_r726198137

And I think this could be implemented fast and is well worth it in my opinion to catch any weird bugs we may have missed:
- https://github.com/MGHComputationalPathology/highdicom/pull/114#discussion_r725541127

Then we are good to go

PR #115, opened 2021-10-05
Annotation init files

CPBridge:
It seems like the `__init__` files were not correctly populated so things aren't importing properly for me if I use the "recommended" pattern...

We should probably switch to using the "recommended" import pattern, e.g.

```python
import highdicom as hd

hd.ann.AnnotationGroup(...)
```

consistently throughout the tests to avoid things like this slipping through.

Have fixed the inits for you and target this pull request at the feature branch


hackermd (approved):
Thanks!

PR #116, opened 2021-10-06
Fix sorting of segmentation image frames

hackermd:


CPBridge:
I think the issue (#117) was closed without need for further changes to the library. Should we just merge this as it is now? @hackermd

PR #118, opened 2021-10-06
Fix type check of referenced segments

hackermd:
The constructor of `highdicom.sr.PlanarROIMeasurementsAndQualitativeEvaluations` only allows `referenced_segment` to be of type `highdicom.sr.ReferencedSegmentationFrame`. It should also allow `highdicom.sr.ReferencedSegmentation`, because in case of tiled slide microscopy images, a planar ROI may cover more than one frame.

### DIFF
--- a/src/highdicom/sr/content.py
+++ b/src/highdicom/sr/content.py
@@ -1468,8 +1468,9 @@ class ReferencedSegmentationFrame(ContentSequence):
                         )
                     src = drv_image.SourceImageSequence[0]
                     source_image = SourceImageForSegmentation(
-                        referenced_sop_class_uid=src.ReferencedSOPClassUID,
-                        referenced_sop_instance_uid=src.ReferencedSOPInstanceUID
+                        src.ReferencedSOPClassUID,
+                        src.ReferencedSOPInstanceUID,
+                        frame_numbers
                     )
                     found_source_image = True
                     break


CPBridge (changes requested):
(see comments above for specific changes)

fedorov:
I am not familiar with how segmentations work for SM, but wouldn't this clause from TID 1410 Row 7 resolve the issue:

> For references to tiled Segmentation Images, one or more values shall be specified in Referenced Frame Number (0008,1160), unless all frames in the referenced Segmentation Image are selected and there is only a single Segment, in which case Referenced Frame Number (0008,1160) will be absent. The referenced tiles shall all be in the same plane.

CPBridge:
> What shall we do? Allow `frame_numbers` to also be of type `Sequence[int]`?

Looks like that's the solution. Also we may want to support allowing it to be absent per @fedorov's comment , though I'm concerned the result of that would be that people would omit the frame number without understanding what that signifies

hackermd:
> Looks like that's the solution. Also we may want to support allowing it to be absent per @fedorov's comment , though I'm concerned the result of that would be that people would omit the frame number without understanding what that signifies.

Agreed! Let's keep it required. I'd like to avoid those nasty conditionals whenever possible.

CPBridge:
Maybe we should require the user to explicitly state that their intent is to reference the whole segment with a further boolean parameter (`whole_segment`?) rather than just omit the frame numbers, to avoid this issue.

hackermd:
> Maybe we should require the user to explicitly state that their intent is to reference the whole segment with a further boolean parameter (whole_segment?) rather than just omit the frame numbers, to avoid this issue.

Or just use "Referenced Segment" instead :)

I think that is an edge case that should be rather rare. I'd suggest keeping it as is for now.

PR #130, opened 2021-10-25
Allow users to specify Content Label and ensure the value is formatted correctly

hackermd:
This PR adds a `content_label` attribute to the `Segmentation`, `ParametricMap` and `MicroscopyBulkSimplyAnnotation` SOP classes to allow users to set the value of the Content Label attribute.

Since the meaning of that attribute is not immediately obvious (thanks @dclunie for clarification), this PR further adds enumerated values for specific SOP classes. However, users do not have to choose a value from the enum, but can also supply custom values. To ensure the values are formatted correctly, a `check_code_string()` function is added that asserts values are formatted according to the constraints of the CS VR.


CPBridge (approved):
Looks good

PR #132, opened 2021-11-04
Implementation of Key Object Selection Document

hackermd:


cgorman (approved):
Looks good as far as i can tell and the tests pass (except for some of the ann package but that's out of scope i figure)

hackermd:
> Looks good as far as i can tell and the tests pass (except for some of the ann package but that's out of scope i figure)

Are you sure you've been using the latest pydicom version (required for the ann package)?

cgorman:
> > Looks good as far as i can tell and the tests pass (except for some of the ann package but that's out of scope i figure)
> 
> Are you sure you've been using the latest pydicom version (required for the ann package)?

Ah yes that would do it, yeah I forgot to update my deps before running it. 👍

PR #134, opened 2021-11-09
Fix `from_dataset` constructor in `KeyObjectSelectionDocument`

cgorman:
Fixed the class method `KeyObjectSelection.from_dataset()` and added a test for it.

I had to change a bit around so you may want to pay close attention to what I did to ensure it makes sense.

### DIFF
--- a/src/highdicom/ko/content.py
+++ b/src/highdicom/ko/content.py
@@ -172,7 +172,7 @@ class KeyObjectSelection(ContentSequence):
                 'Item #1 of sequence is not an appropriate SR Content Item '
                 'because it does not have Template Identifier "2010".'
             )
-        instance = ContentSequence.from_sequence(sequence, is_root=True)
+        instance = ContentSequence.from_sequence(sequence, is_root=is_root)
         instance.__class__ = KeyObjectSelection
         return cast(KeyObjectSelection, instance)
 
--- a/src/highdicom/ko/content.py
+++ b/src/highdicom/ko/content.py
@@ -172,7 +172,7 @@ class KeyObjectSelection(ContentSequence):
                 'Item #1 of sequence is not an appropriate SR Content Item '
                 'because it does not have Template Identifier "2010".'
             )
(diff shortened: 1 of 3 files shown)


src/highdicom/ko/content.py
```diff
@@ -172,7 +172,7 @@ def from_sequence(
                 'Item #1 of sequence is not an appropriate SR Content Item '
                 'because it does not have Template Identifier "2010".'
             )
-        instance = ContentSequence.from_sequence(sequence, is_root=True)
+        instance = ContentSequence.from_sequence(sequence, is_root=is_root)
```
hackermd, line 175: I would keep this hardcoded. We just keep the `is_root` parameter to satisfy the [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle)

tests/test_ko.py
```diff
@@ -201,3 +201,37 @@ def test_construction(self):
...
+            sop_instance_uid=UID(),
+            instance_number=1,
+            manufacturer='MGH Computational Pathology',
+            institution_name='Massachusetts General Hospital',
+            institutional_department_name='Pathology'
+        )
+        assert isinstance(document, KeyObjectSelectionDocument)
+        assert isinstance(document.content, KeyObjectSelection)
+
+        test_document = KeyObjectSelectionDocument.from_dataset(document)
```
hackermd, line 220: That is maybe a bit too simplistic. I would suggest to serialize the `document` instance to a buffer and read it back out using the pydicom `dcmwrite()` and `dcmread()` methods, respectively (see `test_seg.py` for examples).

hackermd (changes requested):

src/highdicom/ko/content.py
cgorman: Fixed in `0b52144`

tests/test_ko.py
cgorman: Fixed in `0b52144`

hackermd (approved):

PR #137, opened 2021-11-18
Ensure segment pixel data are properly scaled

hackermd:
When segmentation type is FRACTIONAL, the pixel data may need to be scaled from [0, 1] to [0, 255].

### DIFF
--- a/src/highdicom/seg/sop.py
+++ b/src/highdicom/seg/sop.py
@@ -641,6 +641,12 @@ class Segmentation(SOPClass):
                 else:
                     planes = pixel_array
                 planes = planes.astype(np.uint8)
+                # It may happen that an boolean array is passed that should be
+                # interpreted as fractional segmentation type. In this case, we
+                # also need to stretch pixel valeus to 8-bit unsigned integer
+                # range by multiplying with the maximum fractional value.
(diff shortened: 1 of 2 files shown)


src/highdicom/seg/sop.py
```diff
@@ -641,6 +641,12 @@ def __init__(
                 else:
                     planes = pixel_array
                 planes = planes.astype(np.uint8)
+                # It may happen that an boolean array is passed that should be
+                # interpreted as fractional segmentation type. In this case, we
+                # also need to stretch pixel valeus to 8-bit unsigned integer
+                # range by multiplying with the maximum fractional value.
```
CPBridge, line 647:
```suggestion
                # It may happen that a boolean array is passed that should be
                # interpreted as fractional segmentation type. In this case, we
                # also need to stretch pixel valeus to 8-bit unsigned integer
                # range by multiplying with the maximum fractional value.
```

CPBridge (approved):
Hmm... nice catch

CPBridge:
Probably we should add a quick regression test here too

hackermd:
> Probably we should add a quick regression test here too

Good idea. Actually found an issue with the existing unit test. Addressed via fbeeb40bf1f26aa8e8add210a3b736e48e474e37

PR #138, opened 2021-11-18
Add finding category to sr content

hackermd:
This PR adds the Finding Category concept, which will be included into the standard via CP 2176.

### DIFF
--- a/src/highdicom/sr/templates.py
+++ b/src/highdicom/sr/templates.py
@@ -2381,7 +2381,8 @@ class MeasurementsAndQualitativeEvaluations(Template):
         measurements: Sequence[Measurement] = None,
         qualitative_evaluations: Optional[
             Sequence[QualitativeEvaluation]
-        ] = None
+        ] = None,
+        finding_category: Optional[Union[CodedConcept, Code]] = None
     ):
         """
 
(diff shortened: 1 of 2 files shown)


CPBridge:
Can't fault the implementation, but why is this only in TID1501 and not TID1410 and TID1411 (Planar/Volumetric ROI Measurements and Qualitative Evaluations)?

hackermd:
> Can't fault the implementation, but why is this only in TID1501 and not TID1410 and TID1411 (Planar/Volumetric ROI Measurements and Qualitative Evaluations)?

You are referring to the CP? Good point! I agree, it should be included in all three of them. @dclunie could you kindly update the CP accordingly?

hackermd:
@CPBridge just to confirm we are on the same page. In the implementation, the property is only implemented on `MeasurementsAndQualitativeEvaluations` (corresponding to TID1501), but the other templates inherit from that class.

hackermd:
@dclunie I just double-checked the CP and it's included in all three templates. Sorry for the confusion.

CPBridge:
Ah ok, sorry. I was getting confused about the inheritance hierarchy. Makes sense

CPBridge (approved):

Issue #135, opened 2021-11-12
Segmentation TotalPixelMatrixRows/Columns are being saved as `float` 


cgorman:
When constructing a Segmentation object I hit [this block](https://github.com/MGHComputationalPathology/highdicom/blob/d16ce415b5ae40afac8f395a0d8a86ac851898c6/src/highdicom/seg/sop.py#L550-L581) which results in a Segmentation object with non-integer TotalPixelMatrixRows and TotalPixelMatrixColumns.

`plane_position_values[last_frame_index]` is `array([[ 1.        ,  1.        , 18.719027  , 53.52480485,  0.        ]])` and the row and column indices are 2 and 3 respectively. The resulting instance has `TotalPixelMatrixRows: 152.719027` and `TotalPixelMatrixColumns: 375.0`.

cgorman:
@hackermd We can probably look at this on Monday, I just didn't want to lose track of it before the weekend

hackermd:
Oh...it seems we cannot directly use the index returned by [DimensionIndexSequence.get_index_position()](https://github.com/MGHComputationalPathology/highdicom/blob/d16ce415b5ae40afac8f395a0d8a86ac851898c6/src/highdicom/seg/content.py#L513), because the index values returned by [DimensionIndexSequence.get_index_values()](https://github.com/MGHComputationalPathology/highdicom/blob/d16ce415b5ae40afac8f395a0d8a86ac851898c6/src/highdicom/seg/content.py#L545) omit the first index of the Segment Number (see [here](https://github.com/MGHComputationalPathology/highdicom/blob/d16ce415b5ae40afac8f395a0d8a86ac851898c6/src/highdicom/seg/content.py#L578)).

We could either fix the index [here](https://github.com/MGHComputationalPathology/highdicom/blob/d16ce415b5ae40afac8f395a0d8a86ac851898c6/src/highdicom/seg/sop.py#L554) and [here](https://github.com/MGHComputationalPathology/highdicom/blob/d16ce415b5ae40afac8f395a0d8a86ac851898c6/src/highdicom/seg/sop.py#L558) or change the return value of `DimensionIndexSequence.get_index_values()`.

@CPBridge what are your thoughts?

CPBridge:
[...]

Issue #122, opened 2021-10-15
Allow for Real World Value Mapping as a Shared Functional Group for Parametric Map IOD


cgorman:
I think if the user passes in either a single RWVM instance in the parametric map constructor, or passes a sequence with a single RWVM, we should add it to the Shared Functional Groups Sequence rather than forcing them into the Per-frame sequence.

@hackermd  I can take care of this if you would like, let me know.

PR #140, opened 2021-11-28
Fix Parametric Map compliance issues

hackermd:
I realized that the Parametric Map IOD requires the Modality LUT Transformation step in the grayscale rendering pipeline to be an identity transformation (i.e., it requires Identity Pixel Value Transformation rather than Pixel Value Transformation): https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_A.75.5.html

That also means only unsigned integer values can be stored. They can still be mapped to a different range of "real world" values though using Real World Value Mapping.

Also fixed several other compliance issues.


src/highdicom/pm/enum.py
```diff
@@ -0,0 +1,45 @@
+"""Enumerate values specific to Parametric Map IODs."""
+from enum import Enum
+
+
+class ImageFlavorValues(Enum):
```
cgorman, line 5: Should we consider including the [CT- and MR-specific defined terms from C.8.15.2.1.1.3 and C.8.13.1.1.1.3](https://dicom.innolitics.com/ciods/parametric-map/parametric-map-image/00080008)? (also for the derived pixel contrast values)

cgorman (commented):
Looks good, I just had a comment on the image value enums but that's it.

src/highdicom/pm/enum.py
hackermd: Addressed via 7216af5c10b05131f42f9ec1a5462f0ede2ebe1f

hackermd:
@cgorman I added a unit test for Parametric Map that will test the dimension index lookup in case spatial coordinates are not preserved. That part of the code was previously not covered.

I also went ahead and fixed #135.

hackermd:
@cgorman also realized that #122 was actually not yet implemented. Fixed that and added a unit test.

PR #142, opened 2021-12-10
Add support for JPEG-LS encoding/decoding

hackermd:
This PR will make it possible to encode and decode images using the JPEG-LS codec (which is much much fast than JPEG2000 for both encoding and decoding).

It would introduce additional dependencies:

1. **pillow-jpls**: enables reading/writing JPEG-LS images using the Pillow API
2. **pylibjpeg**: enables reading of JPEG, JPEG-LS, and JPEG-2000 compressed frames via pydicom

### DIFF
--- a/src/highdicom/legacy/sop.py
+++ b/src/highdicom/legacy/sop.py
@@ -6,8 +6,16 @@ from typing import Any, Dict, List, Optional, Sequence
 
 from pydicom.datadict import tag_for_keyword
 from pydicom.dataset import Dataset
+from pydicom.encaps import encapsulate
+from pydicom.uid import (
+    ImplicitVRLittleEndian,
+    ExplicitVRLittleEndian,
+    JPEG2000Lossless,
+    JPEGLSLossless,
+)
 
 from highdicom.base import SOPClass
+from highdicom.frame import encode_frame
 from highdicom._iods import IOD_MODULE_MAP, SOP_CLASS_UID_IOD_KEY_MAP
 from highdicom._modules import MODULE_ATTRIBUTE_MAP
 
@@ -26,9 +34,9 @@ LEGACY_ENHANCED_SOP_CLASS_UID_MAP = {
 
 
 def _convert_legacy_to_enhanced(
-        sf_datasets: Sequence[Dataset],
-        mf_dataset: Optional[Dataset] = None
-    ) -> Dataset:
+    sf_datasets: Sequence[Dataset],
+    mf_dataset: Optional[Dataset] = None
+) -> Dataset:
     """Converts one or more MR, CT or PET Image instances into one
     Legacy Converted Enhanced MR/CT/PET Image instance by copying information
     from `sf_datasets` into `mf_dataset`.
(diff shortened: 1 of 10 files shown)


hackermd:
We could avoid the **pylibjpeg** dependency by implementing JPEG-LS decoding in pydicom using Pillow (with the extension).

CPBridge (commented):
Looks good! I am not too concerned about the new dependency.

However, currently while this ability to encode JPEG-LS frames will be available, it will not be used by any of the implemented SOP classes. Seg/pm/sc all have a list of supported transfer syntaxes in their implementations (and docstrings). Should we enable support for JPEG-LS in these classes in this PR too?

hackermd:
> Looks good! I am not too concerned about the new dependency.
> 
> However, currently while this ability to encode JPEG-LS frames will be available, it will not be used by any of the implemented SOP classes. Seg/pm/sc all have a list of supported transfer syntaxes in their implementations (and docstrings). Should we enable support for JPEG-LS in these classes in this PR too?

Addressed via 6f58aedb3f4fd81f6990956b4b8242c2d5b6c8f2 and 99a9da5ed2d7e5bcd2240f629a4caea93ac7a5be

src/highdicom/legacy/sop.py
```diff
@@ -454,7 +489,7 @@ def __init__(
             instance_number=instance_number,
             manufacturer=ref_ds.Manufacturer,
             modality=ref_ds.Modality,
-            transfer_syntax_uid=None,  # FIXME: frame encoding
+            transfer_syntax_uid=transfer_syntax_uid,
```
CPBridge, line 492: Just a though, should the default behaviour be to copy the transfer syntax from the input images, and not re-encode the pixels?

CPBridge (approved):
One question to perhaps think about, but otherwise this is good to go

src/highdicom/legacy/sop.py
hackermd: I generally agree.

However, this may be a good opportunity to avoid some of the exotic lossless JEPG transfer syntaxes that are not widely supported and thereby facilitate downstream decoding. Thoughts?

CPBridge: Are you suggesting that if no transfer syntax is explicitly specified, we selectively re-encode some lossless codecs into a more commonly-supported format?

hackermd: Yes, or just not compress at all by default (which has been the behavior so far)

hackermd: If any of the source images was lossy compressed, we also shouldn't re-compress the enhanced image with a lossy transfer syntax

CPBridge: Right, the main thing to avoid would be to decompress lossy data and then recompress losslessly (or store with no compression, as is currently the default) thus giving the impression of lossless compression when in fact information has been lost. That's why I'm suggesting simply leaving the transfer syntax and compression alone by default in those situations unless the user specifically requests otherwise. However I don't see a harm in silently translating between different lossless formats, although it is a bit magical

hackermd: The user can currently not set a lossy transfer syntax. If the pixel data was lossy compressed in the first place, the attribute `LossyImageCompression` should already reflect that (and should stay unchanged even if pixel data is now encoded using a lossless transfer syntax).

The behaviour so far has been to decode the pixel data and re-encode it uncompressed. We can (and probably should) change that behavior upon refactoring of the legacy conversion package as part of https://github.com/MGHComputationalPathology/highdicom/pull/34

PR #143, opened 2021-12-12
Example updates to avoid doctest errors

shishirpy:
Removes errors from some examples.


CPBridge:
Hi @shishirpy, thanks for jumping in here to fix these issues, much appreciated! Unfortunately some of the examples depend on fictitious data, so we'll need to slightly rework some of the sections to make sure they demonstrate what they are supposed to. I'll follow up sometime in the next couple of weeks

shishirpy:
@CPBridge I tried a number of images from the test files, could not get any valid values other than empty lists. If you could suggest a file to use, I can fix these issues too.

Regarding the doctests, I think one can add the following to .travis.yml file - `pytest --doctest-modules`

src/highdicom/io.py
```diff
@@ -225,12 +225,14 @@ class ImageFileReader(object):
...
-    >>> with ImageFileReader('/path/to/file.dcm') as image:
+    >>> with ImageFileReader('data/test_files/ct_image.dcm') as image:
     ...     print(image.metadata)
     ...     for i in range(image.number_of_frames):
     ...         frame = image.read_frame(i)
     ...         print(frame.shape)
-
+    Dataset.file_meta -------------------------------
+    (0002, 0000) File Meta Information Group Length  UL: 192
+    ...
```
CPBridge, line 235: The motivation behind the ImageFileReader is to read individual frames from multiframe DICOM files, so using a single frame DICOM file like this is not a good example. I suggest we switch to an example multiframe CT from the pydicom test data, which has two frames to iterate over:

```python
>>> from pydicom.data import get_testdata_file
>>> test_filepath = get_testdata_file('eCT_Supplemental.dcm')
>>>
>>> with ImageFileReader(test_filepath) as image:
...     print(image.metadata.SOPInstanceUID)
...     for i in range(image.number_of_frames):
...         frame = image.read_frame(i)
...         print(frame.shape)
1.3.6.1.4.1.5962.1.1.10.3.1.1166562673.14401
(512, 512)
(512, 512)
```

src/highdicom/seg/content.py
```diff
@@ -529,6 +529,11 @@ def get_index_position(self, pointer: str) -> int:
         >>> dimension_index = DimensionIndexSequence("SLIDE")
         >>> i = dimension_index.get_index_position("ReferencedSegmentNumber")
         >>> segment_numbers = dimension_index[i]
+        >>> segment_numbers
+        (0020, 9164) Dimension Organization UID          ...
+        (0020, 9165) Dimension Index Pointer             AT: (0062, 000b)
+        (0020, 9167) Functional Group Pointer            AT: (0062, 000a)
+        (0020, 9421) Dimension Description Label         LO: 'Segment Number'
```
CPBridge, line 536: You've made me realise that the name "segment_numbers" doesn't describe what this variable actually is. To make this clearer we should change the variable name:

```suggestion
        >>> dimension_description = dimension_index[i]
        >>> dimension_description
        (0020, 9164) Dimension Organization UID          ...
        (0020, 9165) Dimension Index Pointer             AT: (0062, 000b)
        (0020, 9167) Functional Group Pointer            AT: (0062, 000a)
        (0020, 9421) Dimension Description Label         LO: 'Segment Number'
```

src/highdicom/seg/sop.py
```diff
@@ -1478,28 +1478,33 @@ def get_segment_numbers(
...
         institution-specific tracking ID:
 
         >>> segment_numbers = seg.get_segment_numbers(
-        >>>     tracking_id='Tumor #1234567'
-        >>> )
-        [13]
+        ...     tracking_id='Tumor #1234567'
+        ... )
+        >>> segment_numbers
+        []
```
CPBridge, line 1499: The segmentation file you chose has only a single segment, which means it not the best example for this snippet. Switching to the `'data/test_files/seg_image_sm_control.dcm'` will work better. In either case, searching for `Tumor` segments doesn't make sense because there are none. I suggest everything could look like this:

```python
        >>> from pydicom.sr.codedict import codes
        >>> from highdicom.seg import SegmentAlgorithmTypeValues, Segmentation
        >>> from pydicom import dcmread
        >>> ds = dcmread('data/test_files/seg_image_sm_control.dcm')
        >>> seg = Segmentation.from_dataset(ds)
        >>> segment_numbers = seg.get_segment_numbers(
        ...     segmented_property_type=codes.SCT.ConnectiveTissue,
        ...     algorithm_type=SegmentAlgorithmTypeValues.AUTOMATIC
        ... )
        >>> segment_numbers
        [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]

        Get segment numbers of all segments identified by a given
        institution-specific tracking ID:

        >>> segment_numbers = seg.get_segment_numbers(
        ...     tracking_id='Segment #4'
        ... )
        >>> segment_numbers
        [4]

        Get segment numbers of all segments identified a globally unique
        tracking UID:

        >>> uid = '1.2.826.0.1.3680043.8.498.42540123542017542395135803252098380233'
        >>> segment_numbers = seg.get_segment_numbers(tracking_uid=uid)
        >>> segment_numbers
        [13]

```

CPBridge (changes requested):
Hi @shishirpy,

Sorry it took me so long to get around to looking at this properly. As I said before, we just need to tweak some of the examples a bit so that they convey the behaviour we are trying to demonstrate. I have made a few comments showing my suggestions for exactly how to do this. I f you could just incorporate these into your branch, we should be good to merge.

There is one outstanding failing docstring that I'll look at more closely once we merge this in.

Thanks again for your help!

CPBridge:
Closing this is favour of #165, which includes contents of this PR on a new branch.

PR #146, opened 2022-01-30
Account for non-tiled images with slide-based coordinate system

hackermd:
So far, we have assumed that all images that are positioned relative to the slide coordinate system are multi-frame and tiled, i.e., have `TotalPixelMatrixRows`, `TotalPixelMatrixColumns`, and `TotalPixelMatrixOriginSequence` attributes. However, it turned out that is not true. There are images (e.g., VL Slide-Coordinates Microscopic Image instances) that are positioned in the slide coordinate system but are single-frame and non-tiled. This PR introduces changes that make it possible to encode Segmentation images that are derived from such images.


src/highdicom/seg/content.py
```diff
@@ -500,13 +501,39 @@ def get_plane_positions_of_series(
...
-            PlanePositionSequence(
-                coordinate_system=CoordinateSystemNames.PATIENT,
-                image_position=img.ImagePositionPatient
-            )
-            for img in images
-        ]
+        if self._coordinate_system == CoordinateSystemNames.SLIDE:
+            plane_positions = []
+            for img in images:
+                # Unfortunatel, the image position is not specified relative to
```
CPBridge, line 507:
```suggestion
                # Unfortunately, the image position is not specified relative to
```

src/highdicom/seg/content.py
```diff
@@ -500,13 +501,39 @@ def get_plane_positions_of_series(
...
+                # the top left corner but to the center of the image.
+                # Therefore, we need to compute the offset and subtract it.
+                center_item = img.ImageCenterPointCoordinatesSequence[0]
+                x_center = center_item.XOffsetInSlideCoordinateSystem
+                y_center = center_item.YOffsetInSlideCoordinateSystem
+                z_center = center_item.ZOffsetInSlideCoordinateSystem
+                offset_coordinate = map_pixel_into_coordinate_system(
+                    coordinate=((img.Columns / 2, img.Rows / 2)),
```
CPBridge, line 515: Not certain about this in this specific context... but isn't the centre of an image actually at coordinate `((img.Columns - 1) / 2, (img.Rows - 1) / 2)`?

src/highdicom/seg/sop.py
```diff
@@ -287,6 +288,8 @@ def __init__(
                 'Only one source image should be provided in case images '
                 'are multi-frame images.'
             )
+        is_relative_to_slide = hasattr(src_img, 'ImageOrientationSlide')
```
CPBridge, line 291: I think this is basically just a duplicate of the logic in line 334. We probably don't want two local variables capturing the same information floating around, it's bound to cause future issues.

```python
        if (hasattr(src_img, 'ImageOrientationSlide') or
                hasattr(src_img, 'ImageCenterPointCoordinatesSequence')):
            self._coordinate_system = CoordinateSystemNames.SLIDE
        else:
            self._coordinate_system = CoordinateSystemNames.PATIENT
```

src/highdicom/seg/sop.py
```diff
@@ -443,20 +446,21 @@ def __init__(
...
             )
+        else:
+            if is_multiframe:
+                src_sfg = src_img.SharedFunctionalGroupsSequence[0]
+                source_plane_orientation = src_sfg.PlaneOrientationSequence
+            else:
+                source_plane_orientation = PlaneOrientationSequence(
+                    coordinate_system=self._coordinate_system,
+                    image_orientation=src_img.ImageOrientationPatient
+                )
```
CPBridge, line 462: Could this branching be simplified a bit by having a variable holding `image_orientation` something like this?

```python
if is_multiframe:
    src_sfg = src_img.SharedFunctionalGroupsSequence[0]
    source_plane_orientation = src_sfg.PlaneOrientationSequence
else:
    image_orientation = src_img.ImageOrientationSlide if is_relative_to_slide else src_img.ImageOrientationPatient
    source_plane_orientation = PlaneOrientationSequence(
        coordinate_system=self._coordinate_system,
        image_orientation=image_orientation
    )
```

src/highdicom/seg/sop.py
```diff
@@ -541,15 +545,19 @@ def __init__(
...
-                self.TotalPixelMatrixColumns = \
-                    source_images[0].TotalPixelMatrixColumns
+            )
+            if are_spatial_locations_preserved and is_tiled:
+                self.TotalPixelMatrixOriginSequence = deepcopy(
+                    src_img.TotalPixelMatrixOriginSequence
+                )
+                self.TotalPixelMatrixRows = src_img.TotalPixelMatrixRows
+                self.TotalPixelMatrixColumns = src_img.TotalPixelMatrixColumns
+            elif are_spatial_locations_preserved and ~is_tiled:
```
CPBridge, line 557: Unless I'm missing something...

```suggestion
            elif are_spatial_locations_preserved and not is_tiled:
```

src/highdicom/seg/sop.py
```diff
@@ -565,23 +573,42 @@ def __init__(
...
+                        plane_position_values[last_frame_index, row_index] +
+                        self.Rows
+                    )
+                    self.TotalPixelMatrixColumns = int(
+                        plane_position_values[last_frame_index, col_index] +
+                        self.Columns
+                    )
+                else:
+                    center_coordinate = map_pixel_into_coordinate_system(
+                        coordinate=((self.Columns / 2, self.Rows / 2)),
```
CPBridge, line 601: See comment above about whether this is correct

src/highdicom/seg/sop.py
```diff
@@ -565,23 +573,42 @@ def __init__(
...
+                y_origin = plane_position_values[first_frame_index, y_index]
+                z_index = plane_position_names.index(
+                    'ZOffsetInSlideCoordinateSystem'
                 )
+                z_origin = plane_position_values[first_frame_index, z_index]
+
+                if is_tiled:
+                    origin_item = Dataset()
+                    origin_item.XOffsetInSlideCoordinateSystem = x_origin
+                    origin_item.YOffsetInSlideCoordinateSystem = y_origin
```
CPBridge, line 589: These are DS attributes. Until we have a better solution, make sure to convert to string explicitly with `pydicom.valuerep.format_number_as_ds` to ensure there is no overflow.

src/highdicom/seg/sop.py
```diff
@@ -565,23 +573,42 @@ def __init__(
...
+                        coordinate=((self.Columns / 2, self.Rows / 2)),
+                        image_position=(x_origin, y_origin, z_origin),
+                        image_orientation=plane_orientation,
+                        pixel_spacing=pixel_measures[0].PixelSpacing
+                    )
+                    x_center, y_center, z_center = center_coordinate
+                    center_item = Dataset()
+                    origin_item.XOffsetInSlideCoordinateSystem = x_center
+                    origin_item.YOffsetInSlideCoordinateSystem = y_center
+                    origin_item.ZOffsetInSlideCoordinateSystem = z_center
```
CPBridge, line 610: These are DS attributes. Until we have a better solution, make sure to convert to string explicitly with `pydicom.valuerep.format_number_as_ds` to ensure there is no overflow.

CPBridge (changes requested):
Oh boy... A few minor things to address

CPBridge:
Also, given how complex this part of the code is getting, I think a test would be pretty important. Do you have a suitable test image we can include?

src/highdicom/seg/content.py
hackermd: I think the first pixel of the image has coordinate (1, 1) in the image coordinate system (Column, Row). Therefore, it should not be necessary to subtract one, should it?

src/highdicom/seg/sop.py
hackermd: True! Addressed via 20787eb877effa0e04e165e5c70283c36eb6f345

src/highdicom/seg/sop.py
hackermd: Good catch! Addressed via 18b2e3ef9b24aefb20271ef98f8e3bbb655fe588

src/highdicom/seg/sop.py
hackermd: Addressed via 18b2e3ef9b24aefb20271ef98f8e3bbb655fe588

hackermd:
> Also, given how complex this part of the code is getting, I think a test would be pretty important. Do you have a suitable test image we can include?

I agree! I am currently waiting on feedback from @dclunie on a standard correction proposal. These aspects are not directly related to the changes introduced via the PR, but I would like to make sure the test image is conformant with the standard.

src/highdicom/seg/sop.py
CPBridge: I wondered this, but if I'm not mistaken the current implementation doesn't account for that either?

src/highdicom/seg/content.py
CPBridge: It would appear not. If that were true, then (1, 1) would map to the ImagePositionPatient, which is the frame of reference coordinate of the top left pixel (by definition). But it doesn't:

```python
import highdicom as hd


hd.spatial.map_pixel_into_coordinate_system(
    coordinate=(1, 1),
    image_position=[0.0, 0.0, 0.0],
    image_orientation=[1.0, 0.0, 0.0, 0.0, 1.0, 0.0],
    pixel_spacing=[1.0, 1.0]
)
# outputs (1.0, 1.0, 0.0) not (0.0, 0.0, 0.0)
```

hackermd: Hmm...

The standard states the following

In [Part 3 Section C.8.12.6 Plane Position (Slide) Macro](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.12.6.html#table_C.8.12.6.1-1):

> The column/row position of the top left hand pixel of the frame in the Total Pixel Matrix. The column/row position of the top left pixel of the Total Pixel Matrix is 1.

In [Part 3 Section C.8.12.4.1.4 Total Pixel Matrix Origin Sequence and Image Orientation (Slide)](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.12.4.html#sect_C.8.12.4.1.4)

> Total Pixel Matrix Origin Sequence (0048,0008) specifies the location of the top leftmost pixel of the pixel matrix, and Image Orientation (Slide) (0048,0102) specifies the direction cosines of the first row and the first column of the pixel matrix, both with respect to the Slide Coordinate System Frame of Reference.

In [Part 3 Section C.7.6.2.1.1 Image Position and Image Orientation](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.2.html#sect_C.7.6.2.1.1)

> Image Position (0020,0032) specifies the x, y, and z coordinates of the upper left hand corner of the image; it is the center of the first voxel transmitted. Image Orientation (0020,0037) specifies the direction cosines of the first row and the first column with respect to the patient. These Attributes shall be provide as a pair. Row value for the x, y, and z axes respectively followed by the Column value for the x, y, and z axes respectively.

While the origin of the image coordinate system is (0, 0) and coordinates can be specified at sub-pixel resolution,  the (center of ?) top leftmost pixel of the image seems to be located at (1, 1) per definition.

Am I missing something?

src/highdicom/seg/sop.py
hackermd: My understanding is as follows:

In case of a slide coordinate system, the information is always stored in `ImageOrientationSlide` irrespective of whether the source image is multi-frame or single-frame. However, that is one of the aspects I am currently discussing with @dclunie as part of the mentioned correction proposal.

In case of a patient coordinate system, the information is stored either in `PlaneOrientationSequence` if the source image is multi-frame or in `ImageOrientationPatient` if the source image is single-frame.

Does the current implementation not match this?

src/highdicom/seg/content.py
CPBridge: Right, but what's at issue here is not really what the standard states but rather what the `map_pixel_into_coordinate_system` function implements, since this is what you are calling. As shown in my snippet above. It appears that when we implemented this function, we defined the inputs in the numpy/python definition rather than the DICOM definition, such that it gives you the Frame of Reference coordinate for a given pixel specified in numpy/python convention. Thus you want to pass to it the coordinates of the centre pixel defined in numpy/python convention, not DICOM convention.

This raises two further issues:
- Was it a mistake to define the `map_pixel_into_coordinate_system` in this way? I would argue probably not given that we defer to numpy for all pixel manipulation, so it is natural for users to specify in this way.
- Is this sufficiently clearly documented? I would argue it probably isn't...

hackermd: I don't think this has to do with NumPy versus DICOM but rather with the definition of the image coordinate system and its relationship to the frame of reference.

[Equation C.7.6.2.1-1](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.2.html#sect_C.7.6.2.1.1) states:

> *i* Column index to the image plane. The first column is index zero.
> *j* Row index to the image plane. The first row index is zero.

That means that image coordinate `(0.0, 0.0)` should map to slide coordinates `(0.0, 0.0, 0.0)` (for image position `(0.0, 0.0, 0.0)`, image orientation `(1.0, 0.0, 0.0, 0.0, 1.0, 0.0)`, and pixel spacing `(1.0, 1.0)`) in the frame of reference (X, Y, Z).
So I think we implemented the transformation function correctly. I also think our documentation is correct. However, we may want to further clarify what "(Column, Row) position in the Total Pixel Matrix in pixel unit at sub-pixel resolution" means.

My interpretation is that the center of the top leftmost pixel in the pixel matrix is then actually defined as `(0.5, 0.5)` in the image coordinate system. That also makes sense, because the `Column` coordinate and the `Row` coordinate of any point selected within an image at sub-pixel resolution (e.g., during vector graphic annotation) would then actually lie within the `[0, Columns]` range and the `[0, Rows]` range, respectively.

That would mean, however, that the center of the top leftmost pixel would map to `(0.5, 0.5, 0.0)` in the frame of reference, wouldn't it?

It seems DICOM uses one-based indexing to specify the position of a plane relative to the total pixel matrix and zero-based indexing to specify the position of a plane relative to the frame of reference.

Am I understanding this correctly?

CPBridge: Ok you're right. The DICOM/numpy thing was a confusing red herring. It seems both define the coordinates of the top left at (0, 0).

However the crucial difference is whether the (0, 0) represents the centre of the top left pixel, or the top left of the top left pixel, and equivalently whether the ImagePositionPatient/Slide is the frame of reference location of the centre of the top left pixel or the top left of the top left pixel.

It would appear that (0, 0) is the image space coordinate of the *centre* of the top left pixel, based on this:

> Image Position (0020,0032) specifies the x, y, and z coordinates of the upper left hand corner of the image; it is the center of the first voxel transmitted

If we work on the assumption that that is true, then the centre of the image is at image coordinate `((img.Columns - 1) / 2, (img.Rows - 1) / 2)`. E.g. in an image with 3 rows and three columns it is at the centre of the second pixel in each dimension, which has image coordinates `(1, 1)`.

src/highdicom/seg/sop.py
CPBridge: Ah I see, makes sense

src/highdicom/seg/content.py
hackermd: > It would appear that (0, 0) is the image space coordinate of the centre of the top left pixel

Yes, it would appear so and I think you were right about `((img.Columns - 1) / 2, (img.Rows - 1) / 2)` to begin with.

However, the [C.18.6 Spatial Coordinates Macro](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.18.6.html) states:

> An ordered set of (column,row) pairs that denote positions in an image specified with sub-pixel resolution such that the origin at the TLHC of the TLHC pixel is 0.0\0.0, the BRHC of the TLHC pixel is 1.0\1.0, and the BRHC of the BRHC pixel is Columns\Rows. The values must be within the range 0\0 to Columns\Rows.

So it would appear that the position of the image is defined relative to the center of the TLHC pixel, but any points within the image at sub-pixel resolution are defined relative to the top left corner of the TLHC pixel.

Does that mean that we have to add half a pixel to each point when mapping graphic data from SCOORD to SCOORD3D?

I am confused...

CPBridge: Yikes. That would be my reading of it too. I guess that is not technically a contradiction, but it's certainly a very strange design choice... Maybe we should ask around for confirmation? Don't really know how to proceed...

hackermd: @dclunie @fedorov @pieper could you kindly help us shine light into this?

It appears that spatial coordinates (SCOORD) are defined relative to the top left corner of the TLHC pixel, while the image position is defined relative to the center of the TLHC pixel. Is that correct?

pieper: This is how I did this in Slicer.  It'll be good to develop some reference examples to confirm all the code is consistent with the standard.

https://github.com/QIICR/QuantitativeReporting/blob/master/DICOMPlugins/DICOMTID1500Plugin.py#L424-L436

dclunie: Yes, exactly.

Historically, CP 137 clarified that the 3D coordinates in Image Position (Patient) are the center of the voxel; this was ambiguous prior to that.

When Sup 23 (SR) added SCOORD, we had just finished presentation states (Sup 33), during the development of which we realized that we needed to support sub-pixel resolution for image and displayed area relative coordinates, so we decided to specify them as 0-n (where n is the number of rows or columns), amongst many other untidy alternatives, as illustrated in this figure:
[https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.10.5.html#figure_C.10.5-1](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.10.5.html#figure_C.10.5-1)

Handling this is not as simple as "adding 0.5" - also consider that the range of the 2D coordinates is 0 to n, not 0 to (n-1) (e.g., 513, not 512, in the case of a 512x512 matrix frame).

hackermd: Thanks for the clarification @dclunie

hackermd: @dclunie just to make sure I actually got this:

We try to compute the distance between the center pixel of the image and the TLHC pixel of the image in the frame of reference to determine the image position relative to the origin of the slide coordinate system.

To calculate the coordinates of these two points in the frame of reference, we need to use spatial coordinates with sub-pixel resolution:

- Center pixel: `((Columns + 1) / 2, (Rows + 1) / 2)`
- TLHC pixel: `(0.0, 0.0)`

Is that correct?

dclunie: In the general case, you need to project the vector from the TLHC to the image center onto the plane defined by Image Orientation, and then translate the resulting XYZ vector to the center of the TLHC voxel (Image Position) origin.

The distance (in image-relative 2D pixels) from the TLHC to the center pixel is the same whether measured from the TLHC of the TLHC pixel to the TLHC of the center pixel, or the center of the TLHC pixel to the center of the center pixel, is it not?

Consider though, whether there is such a thing as a "center pixel", when the number of rows or columns is even rather than odd.

So I assume you actually mean "center of the image" rather than "center pixel".

Consider the case of a 2x2 image as a fringe case.

The distance projected into 3D needs to be scaled by PixelSpacing values (accounting for non-square pixels).

hackermd: > Consider though, whether there is such a thing as a "center pixel", when the number of rows or columns is even rather than odd.
> So I assume you actually mean "center of the image" rather than "center pixel".
> Consider the case of a 2x2 image as a fringe case.

Yes, I mean center of the image.

> The distance projected into 3D needs to be scaled by PixelSpacing values (accounting for non-square pixels).

That is handled by the transformation function. The question is what image coordinates do we need to provide as input to get the correct slide coordinates.

dclunie: The question is, what coordinates does the mysterious "transformation function" expect?

hackermd: That mysterious transformation function is the one defined by [Equation C.7.6.2.1-1](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.2.html#sect_C.7.6.2.1.1).

I think the source of confusion are these two lines:

> *i* Column index to the image plane. The first column is index zero.
> *j* Row index to the image plane. The first row index is zero.

If my understanding is correct, *i* and *j* are not **indices** to the image plane, but rather spatial coordinates at sub-pixel resolution defined relative to the origin of the total pixel matrix (top left corner of the top left pixel), i.e., floating-point values in the range `[0, Columns]` and `[0, Rows]`.

pieper: I'd say they range from [0, Columns+1)  and [0, Rows+1) where the index is the floor of the value.  E.g. the line i=1 is exactly on the border between the first and second pixel.

dclunie: C.7.6.2.1-1 does not account for sub-pixel resolution, and instead seems to map whole pixel references where the whole TLHC pixel is given the index (0,0), and maps the center of that pixel to the Image Position (Patient) 3D location.

This definitely needs clarification in the standard.

This equation and the accompanying text was added in Sup 73 Spatial Registration Storage SOP Classes, though really has nothing to do with registration per se; I don't think it occurred to us at that time that we needed to consider the implications for interpretation of SCOORD sub-pixel resolution and mapping to 3D.

I think there is a need for scaling by n/(n+1) and translation by -0.5*s before applying C.7.6.2.1-1 to sub-pixel resolution image-relative coordinates, such as used in SCOORD.

I will work on a CP.

dclunie: I am not much good at matrix math, but here is a first attempt at a CP:
[cp_dac607_SubPixelResolutionSCOORDMappingTo3DRCS.pdf](https://github.com/MGHComputationalPathology/highdicom/files/8008168/cp_dac607_SubPixelResolutionSCOORDMappingTo3DRCS.pdf)

hackermd: Thanks @dclunie!

@CPBridge and I also discussed this at greater detail last night and reached the following conclusion:

There are three coordinate systems in DICOM:

1. **Frame of Reference**: Three-dimensional patient or slide coordinate system with axes X, Y, and Z. Values of (x, y, z) coordinates are single-precision floating-point numbers in range [-inf, inf] in millimeter unit. These values correspond to 3D spatial coordinates (SCOORD3D). The (0.0, 0.0) point is located at the origin, which may be explicitly (e.g., SLIDE_CORNER in case of slide coordinate system) or implicitly defined. We refer to the (x, y, z) triplets as *reference coordinates*.
2. **Image**: Two-dimensional image coordinate system with axes Columns and Rows. Values of (column, row) coordinates are single-precision floating-point numbers in range [0, Columns] and [0, Rows], respectively, in pixel unit at sub-pixel resolution. These values correspond to spatial coordinates (SCOORD). The (0.0, 0.0) point is located at the top left corner of the top left pixel of the total pixel matrix. We refer to (column, row) pairs as *image coordinates*.
3. **Total Pixel Matrix**: Two-dimensional coordinate system with axes Columns and Rows. Values of (i, j) coordinates are integers in range [0, Columns - 1] and [0, Rows - 1], respectively, in pixel unit and represent zero-based indices to the matrix. The (0, 0) index is located at the center of the top left pixel of the total pixel matrix. We refer to (i, j) pairs as *image indices*.

Equation C.7.6.2.1-1 defines the transformation of two-dimensional *image indices* (3) into three-dimensional *reference coordinates* (1).
If two-dimensional *image coordinates* (SCOORD) (2) should be transformed into three-dimensional *reference coordinates* (SCOORD3D) (1), then the 0.5 pixel translation between (2) and (3) needs to be taken into account.

Your CP uses similar terminology and clarifies the above issue. However, I think there are a few more things we should further clarify as well:

- Distinction between *image coordinates* and *image indices* and whether either of them are defined outside of the Total Pixel Matrix. Can one have negative *image indices* or negative *image coordinates*? Those may occur, for example, upon image annotation when a user draws beyond the border of the pixel matrix or when *reference coordinates* are mapped onto an image plane but don't map to a location within the borders of the pixel matrix.
- Distinction between zero-based *image indices* that define Image Position and one-based indices that define  Plane Position. Confusingly, the offset of a tile from the top left pixel of the total pixel matrix is defined by one-based (column, row) indices. To translate a plane position into *image coordinates* (SCOORD) one needs to subtract 1.5!

hackermd: @CPBridge could you take a look at  a4ef08a85aa7f8fdc07dec1852e40b9bf0e06e34?

I have not yet added tests for the image coordinates <-> reference coordinates transformations, since I wanted to get your input first.
If I am not mistaken, we should just be able to account for the -0.5 shift prior to or after applying the affine transformation to translate between image coordinates and (sub-pixel resolution) pixel indices.

src/highdicom/spatial.py
```diff
@@ -35,6 +35,395 @@ def create_rotation_matrix(
     ])
 
 
+def _create_affine_transformation_matrix(
+    image_position: Sequence[float],
+    image_orientation: Sequence[float],
+    pixel_spacing: Sequence[float],
+) -> np.ndarray:
+    """Create affine matrix for transformation.
```
CPBridge, line 43: In case we ever have to revisit this again (you may have to fix my spacing here as the github interface makes that fiddly):

```suggestion
    """Create affine matrix for transformation.
    
    The resulting transformation matrix maps the center of a pixel identified by zero-based 
    integer indices into the frame of reference, i.e. an input value of (0, 0) represents the
    center of the top left hand corner pixel.
    
    See :dcm:`Equation C.7.6.2.1-1 <part03/sect_C.7.6.2.html#sect_C.7.6.2.1.1>`.
    
```

src/highdicom/spatial.py
```diff
@@ -35,6 +35,395 @@ def create_rotation_matrix(
...
+
+
+def _create_inv_affine_transformation_matrix(
+    image_position: Sequence[float],
+    image_orientation: Sequence[float],
+    pixel_spacing: Sequence[float],
+    spacing_between_slices: float = 1.0
+) -> np.ndarray:
+    """Create affine matrix for inverse transformation.
+
```
CPBridge, line 116:
```suggestion
    """Create affine matrix for inverse transformation.
 
    The resulting transformation matrix maps a frame of reference coordinate
    to pixel indices, where integer pixel index values represent the center of the
    pixel in the image, i.e. an output value of exactly (0.0, 0.0)
    represents the center of the top left hand corner pixel.
```

src/highdicom/spatial.py
```diff
@@ -35,6 +35,395 @@ def create_rotation_matrix(
...
+                -np.dot(inv_rotation, translation)
+            ]),
+            [0.0, 0.0, 0.0, 1.0]
+        ]
+    )
+
+
+class PixelToReferenceTransformer(object):
+
+    """Class for transforming pixel indices to reference coordinates.
```
CPBridge, line 192: I think it would really help users to add a clarifying `Note` here about the two definitions of the image coordinate system and how these four classes relate to them, what do you think?

CPBridge (approved):
Looks good. The implementation of the coordinate transformation is correct according to my current understanding. I have approved but made some minor documentation suggestions.

src/highdicom/seg/sop.py
hackermd: Now using image coordinates rather than pixel indices

PR #147, opened 2022-01-30
Add method to get all graphic data of an annotation group

hackermd:
Currently, the `highdicom.ann.AnnotationGroup` class only provides the `get_coordinates()` method to get the coordinates of a single graphical annotation. This method decodes the binary graphic data every time it is called.

This PR adds a `get_graphic_data()` method to get the coordinates of all graphical annotations at once. In addition, it adds a mechanism to cache the decoded coordinates data and thereby speed subsequent calls to either `get_graphic_data()` or `get_coordinates()`.


src/highdicom/ann/content.py
```diff
@@ -496,6 +498,73 @@ def primary_anatomic_structures(self) -> List[CodedConcept]:
...
+
+        Returns
+        -------
+        List[numpy.ndarray]
+            Two-dimensional array of floating-point values representing either
+            2D or 3D spatial coordinates for each graphical annotation
+
+        """  # noqa: E501
+        coordinate_type = AnnotationCoordinateTypeValues(coordinate_type)
+        if self._graphic_data is None:
```
CPBridge, line 520: Technically, this caching mechanism doesn't take into account the value of the `coordinate_type` parameter.

Consequently, subsequent calls to the method can pass the wrong value and it will return the old data. I doubt this would happen much in practice, but still seems like a potentially nasty bug waiting to happen:

```python
data = group.get_graphic_data('SCOORD')
...
more_data = group.get_graphic_data('SCOORD3D')  # will return data, which is definitely not the expected behaviour...
```

I suggest you consider caching the coordinate type too somehow between calls.

EDIT this is especially pernicious when you realise that users may call this function indirectly through the `get_coordinates` method on multiple occasions and potentially pass different `coordinate_type`s (not that there's any reason to of course, but it still *could* happen...).

src/highdicom/ann/content.py
```diff
@@ -496,6 +498,73 @@ def primary_anatomic_structures(self) -> List[CodedConcept]:
...
+                coordinates_dtype = np.float32
+            decoded_coordinates_data = np.frombuffer(
+                coordinates_data,
+                coordinates_dtype
+            )
+
+            self._graphic_data = []
+            for i in range(self.number_of_annotations):
+                annotation_number = i + 1
+                coordinate_index = self._get_coordinate_index(
```
CPBridge, line 540: Take it or leave it... but if I'm not mistaken the length of the annotations is fixed in the case of a RECTANGLE, POINT, or ELLIPSE. In these cases the entire loop here could be replaced with a basic array reshape, which would be very significantly faster. If you're worried about efficiency at all this is probably worth implementing two different cases (one for RECTANGLE, POINT, ELLIPSE and the other for POLYGON, POLYLINE).

CPBridge (changes requested):
Nice, like the idea. Couple of comments about implementation

src/highdicom/ann/content.py
CPBridge: In fact, now that I think about it, I think you may be able to use `np.split()` and the point index list to replace the loop even in the polygon/polyline case. You'd just have to figure out what to do in the common z value case.

src/highdicom/ann/content.py
hackermd: Addressed via c61aa724b733463e8e3cad9275a28b2f9fbf288a

src/highdicom/ann/content.py
hackermd: Nice one! Addressed via 0b5096527929f83376167b518c9588b7b1b510f2.

We still need one loop (now implemented via a list comprehension) to reshape the coordinates and include the z-coordinate. But it should be faster.

CPBridge (commented):
Happy to merge as is now but I have suggested what I think is an improvement to remove all loops

src/highdicom/ann/content.py
hackermd: Nice! Addressed via 852dc1d94446db09d2b8025b4b6e14ca637f2197

CPBridge (approved):
Good to go

PR #151, opened 2022-02-18
Fix datetime formatting during legacy conversion

hackermd:


src/highdicom/legacy/sop.py
```diff
@@ -172,8 +172,8 @@ def _convert_legacy_to_enhanced(
         frame_content_item = Dataset()
         if 'AcquisitionDate' in ds and 'AcquisitionTime' in ds:
             frame_content_item.FrameAcquisitionDateTime = '{}{}'.format(
-                ds.AcquisitionDate,
-                ds.AcquisitionTime
+                ds.AcquisitionDate.strftime('%Y%m%d'),
+                ds.AcquisitionTime.strftime('%H%M%S')
             )
```
CPBridge, line 177: Consider the following perhaps slightly neater way:

```python
frame_content_item.FrameAcquisitionDateTime = datetime.datetime.combine(
    ds.AcquisitionDate, ds.AcquisitionTime
).strftime('%Y%m%d%H%M%S')
```

which I believe gives the same result

CPBridge (approved):
Fine as is, but one small suggestion for a slightly neater way

src/highdicom/legacy/sop.py
hackermd: Addressed via 821a65a258b8d9b0d13c952fc58fcadd3ef7b82e

PR #154, opened 2022-02-24
Change person name checks from errors to warnings

CPBridge:
I think I was probably a bit over zealous in raising an error when a malformed Person Name is passed. We are finding here that many DICOM files exist "in the wild" with incorrectly formatted patient names, unfortunately. This is now bringing down model inference pipelines when attempting to copy across the name and crashing everything when writing out results...

I suggest a more pragmatic compromise may be to raise a warning instead of an error, as implemented in this PR. @hackermd thoughts?

### DIFF
--- a/src/highdicom/valuerep.py
+++ b/src/highdicom/valuerep.py
@@ -7,7 +7,7 @@ from pydicom.valuerep import PersonName
 
 
 def check_person_name(person_name: Union[str, PersonName]) -> None:
-    """Check the value representation person name strings.
+    """Check value is valid for the value representation "person name".
 
     The DICOM Person Name (PN) value representation has a specific format with
     multiple components (family name, given name, middle name, prefix, suffix)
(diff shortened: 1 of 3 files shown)


CPBridge:
Tagging @delton137

hackermd:
@CPBridge I ran into problems with this check, too.

Strict checks make sense for construction, but are problematic for parsing. However, we sometimes call constructors via `from_dataset()`, which can also lead to problems.

hackermd (approved):

hackermd (changes requested):
@CPBridge could you please update the tests accordingly?

delton137 (approved):

CPBridge:
I did update the tests, no?

> Strict checks make sense for construction, but are problematic for parsing. 

Right but this is sort of a grey zone on between where we are constructing based on the values found in another file, and there's no way for the code in that function to know whether it was user input or copied from somewhere else.

CPBridge:
Oh i see it's failing in 3.10, hmm

CPBridge:
Fixed the tests

seandoyle:
A further reason why changing the error to a warning is that many anonymization programs (CTP, the old CCDS one, I think a few others) replace the names with strings created with a truncated hash of the original name.

hackermd (approved):

PR #156, opened 2022-02-25
Remove superfluous file meta information attributes

erikogabrielsson:
DICOMPrefix and FilePreamble are not recognized and not needed by pydicom and gives warnings/exceptions depending on setting.

Removed file save to /tmp-path.

### DIFF
--- a/src/highdicom/base.py
+++ b/src/highdicom/base.py
@@ -129,8 +129,6 @@ class SOPClass(Dataset):
         # to a file in PS3.10 format.
         self.preamble = b'\x00' * 128
         self.file_meta = FileMetaDataset()
-        self.file_meta.DICOMPrefix = 'DICM'
-        self.file_meta.FilePreamble = self.preamble
         self.file_meta.TransferSyntaxUID = UID(transfer_syntax_uid)
         self.file_meta.MediaStorageSOPClassUID = UID(sop_class_uid)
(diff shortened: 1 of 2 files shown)


src/highdicom/base.py
```diff
@@ -129,8 +129,6 @@ def __init__(
         # to a file in PS3.10 format.
         self.preamble = b'\x00' * 128
```
hackermd, line 130: We can now also remove this line.

```suggestion
```

hackermd (changes requested):
Thanks @erikogabrielsson! I have just one minor suggestion.

hackermd (approved):

src/highdicom/base.py
```diff
@@ -127,10 +127,7 @@ def __init__(
 
         # Include all File Meta Information required for writing SOP instance
         # to a file in PS3.10 format.
-        self.preamble = b'\x00' * 128
         self.file_meta = FileMetaDataset()
```
hackermd, line 130:
```suggestion
        self.preamble = b'\x00' * 128
        self.file_meta = FileMetaDataset()
```

hackermd: Turns out I was wrong about this. I reintroduced the attribute via ef8f0756ef858025b57d5c15814274eb6155184b

PR #157, opened 2022-02-26
Fix JPEG encoding of frames

hackermd:
With https://github.com/herrmannlab/highdicom/pull/152 we fixed the decoding of JPEG compressed frames. I just realized, however, that the encoding has also been wrong.
We previously assumed that `pydicom.pixel_array` would decode the data correctly and thus tweaked the encoding such that the output of `pydicom.pixel_array` would match the input into the encoding function. Now that we figured out why the color space of `pydicom.pixel_array` is incorrect (unexpected), we can resolve the mystery of why we had to convert the color space during encoding.


src/highdicom/frame.py
```diff
@@ -52,26 +52,37 @@ def encode_frame(
...
         When `transfer_syntax_uid` is not supported or when
         `planar_configuration` is missing in case of a color image frame.
 
-    """
+    Note
+    ----
+    In case of color image frames, the `photometric_interpretation` parameter
+    describes the color space of the **encoded** pixel data and data may be
+    converted from RGB color space into the specified color space upon
+    encoding.  For example, the used JPEG codec convert pixels from RGB into
```
CPBridge, line 79: This does quite sound right
```suggestion
    encoding.  For example, the JPEG codec converts pixels from RGB into
```

src/highdicom/frame.py
```diff
@@ -350,8 +359,28 @@ def decode_frame(
     ValueError
         When transfer syntax is not supported.
 
-    """
+    Note
+    ----
+    In case of color image frames, the `photometric_interpretation` parameter
+    describes the color space of the **encoded** pixel data and data may be
+    converted from the specified color space into RGB color space upon
+    decoding.  For example, JPEG codec generally convert pixels from RGB into
```
CPBridge, line 367:
```suggestion
    decoding.  For example, the JPEG codec generally converts pixels from RGB into
```

src/highdicom/frame.py
```diff
@@ -350,8 +359,28 @@ def decode_frame(
...
+    YBR color space prior to compression to take advantage of the correlation
+    between RGB color bands and improve compression efficiency. In case of an
+    image data set with an encapsulated Pixel Data element containing JPEG
+    compressed image frames, the value of the Photometric Interpretation
+    element specifies the color space in which image frames were compressed.
+    If `photometric_interpretation` specifies a YBR color space, then the
```
CPBridge, line 373: for clarity

```suggestion
    If `photometric_interpretation` specifies a YBR color space, then this
```

CPBridge (approved):
I think I'm getting close to understanding this,,, glad you have!

Made a couple of minor suggestions

PR #158, opened 2022-03-05
Fix computation of plane positions for TILED_FULL

hackermd:
The utility function `highdicom.compute_plane_position_slide_per_frame()` did not handle images with multiple optical paths correctly and returned the wrong number of plane positions. This PR fixes the issue and adds unit tests to avoid regression in the future.


CPBridge (approved):
Looks good to me. Sorry for delay

PR #166, opened 2022-04-12
Refactor Softcopy Presentation State constructors

hackermd:
This PR introduces the following changes based on discussion on #139:

- Use setter functions to add attributes of modules to instances of Softcopy Presentation State classes instead of deriving the classes from a common base class. This will allow implementation of additional classes (e.g., Advanced Blending Softcopy Presentation State), which contain different modules.
- Remove `copy_modality_lut` and `copy_voi_lut` parameters from constructor of `GrayscaleSoftcopyPresentationState` and friends and instead always copy attributes of Modality LUT and VOI LUT modules from referenced source images by default. In almost all cases, this is probably the desired behavior. In case a developers really wants to override that behavior and not apply a Modality LUT or VOI LUT, a custom Modality LUT or VOI LUT can be provided via the `modality_lut` or `softcopy_voi_lut`, respectively, to describe an identity function.
- Fix color management using ICC profiles. Similar to the logic of dealing with Modality LUT and VOI LUT for grayscale images in `GrayscaleSoftcopyPresentationState`, an ICC Profile from one of the referenced color images will be used by default for `ColorSoftcopyPresentationState` if no ICC Profile is provided via the `icc_profile` parameter. For `PseudoColorSoftcopyPresentationState`, a default ICC Profile for the sRGB color space will be included if none is provided by the developer via the `icc_profile` parameter.


src/highdicom/pr/content.py
```diff
@@ -737,3 +757,1295 @@ def __init__(
...
+        self.ReferencedImageSequence = ref_im_seq
+        self.StudyInstanceUID = ref_im.StudyInstanceUID
+        self.SeriesInstanceUID = UID()
+
+        # Modality LUT
+        if modality_lut is not None:
+            _add_modality_lut_attributes(
+                self,
+                modality_lut=modality_lut
+            )
```
CPBridge, line 1876: Check that rescale parameters were not provided in this case (I think they should be mutually exclusive, right?)

src/highdicom/pr/content.py
```diff
@@ -737,3 +757,1295 @@ def __init__(
...
+
+
+def _extract_softcopy_voi_lut_attributes(
+    referenced_images: Sequence[Dataset]
+) -> Dataset:
+    """Get attributes of module Softcopy VOI LUT from referenced images.
+
+    Any Window Center, Window Width, Window Explanation, VOI LUT Function,
+    or VOI LUT Sequence attributes the referenced images are copied to the
+    presentation state dataset.  Missing values will cause no errors, and
```
CPBridge, line 1454:
```suggestion
    new dataset.  Missing values will cause no errors, and
```

src/highdicom/pr/content.py
```diff
@@ -737,3 +757,1295 @@ def __init__(
...
+        if institutional_department_name is not None:
+            dataset.InstitutionalDepartmentName = institutional_department_name
+    dataset.DeviceSerialNumber = device_serial_number
+    dataset.ManufacturerModelName = manufacturer_model_name
+    dataset.SoftwareVersions = software_versions
+
+    # Not technically part of PR IODs, but we include anyway
+    now = datetime.datetime.now()
+    dataset.ContentDate = DA(now.date())
+    dataset.ContentTime = TM(now.time())
```
CPBridge, line 806: Is this the best place for this? Wouldn't `_add_presentation_state_identification_attributes`,  where the `PresentationCreationDate` is added, make more sense?

src/highdicom/pr/content.py
```diff
@@ -737,3 +757,1295 @@ def __init__(
...
+        dataset.ContentCreatorIdentificationCodeSequence = \
+            content_creator_identification
+
+
+def _add_presentation_state_relationship_attributes(
+    dataset: Dataset,
+    referenced_images: Sequence[Dataset]
+) -> None:
+    """Add attributes of module Presentation State Relationship.
+
```
CPBridge, line 890:
```suggestion

Also perform checks that the referenced images are suitable.

```

src/highdicom/pr/content.py
```diff
@@ -737,3 +757,1295 @@ def __init__(
...
+            'Could not find an ICC Profile in any of the referenced images.'
+        )
+    if len(set(icc_profiles)) > 1:
+        raise ValueError(
+            'Found more than one ICC Profile in referenced images.'
+        )
+
+    dataset = Dataset()
+    dataset.ICCProfile = icc_profiles[0]
+    return dataset
```
CPBridge, line 1685: Not sure I really understand the purpose of returning a dataset here, rather than just the bu
ytes?

src/highdicom/pr/sop.py
```diff
@@ -145,63 +157,45 @@ def __init__(
...
-                raise ValueError(
-                    'All referenced images must belong to the same series.'
+        for kw in [
+            'icc_profile',
+        ]:
+            if kw in kwargs:
+                raise TypeError(
+                    'GrayscaleSoftcopyPresentationState() got an unexpected '
+                    f'keyword argument "{kw}".'
                 )
```
CPBridge, line 185: This test is no longer needed if the class does not inherit from the pr base class.

CPBridge: Previously it was possible for people to sneak through some of the keyword args for the base class that are not used for the subclass in the `**kwargs` of the subclass's constructor call so we had to guard against this. Yeah... ugly....

src/highdicom/pr/sop.py
```diff
@@ -1409,93 +770,122 @@ def __init__(
...
+            'rescale_slope',
+            'rescale_type',
+            'modality_lut',
+            'softcopy_voi_luts',
+        ]:
+            if kw in kwargs:
+                raise TypeError(
+                    'ColorSoftcopyPresentationState() got an unexpected '
+                    f'keyword argument "{kw}".'
+                )
```
CPBridge, line 784: Like for the GSPS case I mentioned above, this test is no longer needed and should be removed

src/highdicom/pr/sop.py
```diff
@@ -1572,65 +969,169 @@ def __init__(
...
+
+        if not np.array_equal(
+            blending_input_numbers,
+            np.arange(1, len(blending_input_numbers) + 1, 1)
+        ):
+            raise ValueError(
+                'The values of attribute Blending Input Number of items of '
+                'argument "blending" must be ordinal numbers starting from 1 '
+                'and monotonically increasing by 1.'
+            )
```
CPBridge, line 1054: This seems quite overly complex, what's wrong with this:
```suggestion
        for i, item in enumerate(blending, 1):
            if not isinstance(item, AdvancedBlending):
                raise TypeError(
                    'Items of argument "blending" must have type '
                    'AdvancedBlending.'
                )
            if item.StudyInstanceUID != ref_item.StudyInstanceUID:
                raise ValueError(
                    'All items of argument "blending" must have the same '
                    'Study Instance UID.'
                )
            if int(item.BlendingInputNumber) != i:
                raise ValueError(
                    'The values of attribute Blending Input Number of items of '
                    'argument "blending" must be ordinal numbers starting from 1 '
                    'and monotonically increasing by 1.'
                )
```

CPBridge (changes requested):
Thanks @hackermd, that was quite a rewrite...

A couple of minor things

src/highdicom/pr/content.py
CPBridge: Note that this applies also to the constructors of GSPS and PseudoColorPS

src/highdicom/content.py
```diff
@@ -1341,7 +1347,13 @@ def __init__(
 
 class VOILUT(Dataset):
 
-    """Dataset describing a value-of-interest lookup table."""
+    """Class implementating the VOI LUT Macro.
+
+    Warning
+    -------
+    This class does **not** implement an item of the VOI LUT Sequence.
```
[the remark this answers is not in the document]
hackermd: I think technically speaking, it would still implement the macro. However, the macro attributes are included into the Frame VOI LUT Sequence:
https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.16.2.html#table_C.7.6.16-11b

Even more confusingly, the VOI LUT macro includes the VOI LUT Sequence attribute and the `VOILUT` class has attribute `VOILUTSequence`. Therefore, it is important to make clear that instances of the class are not items of the VOI LUT Sequence attribute.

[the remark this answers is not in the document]
hackermd: Maybe we should rename the class to `VOILUTTransformation` or something along those lines. That would be a deviation from the standard, but it would probably be much less confusing.

src/highdicom/content.py
```diff
@@ -1478,15 +1503,318 @@ def __init__(
                 )
 
 
+class ColorLUT(Dataset):
```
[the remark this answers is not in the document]
hackermd: The *highdicom* interface (protocol) is very similar indeed, but the attributes have very different names (`RedPaletteColorLookupTableDescriptor` instead of `LUTDescriptor`, etc.)

src/highdicom/content.py
```diff
@@ -1478,15 +1503,318 @@ def __init__(
...
+
+        Parameters
+        ----------
+        first_mapped_value: int
+            Pixel value that will be mapped to the first value in the
+            lookup table.
+        number_of_entries: int
+            Total number of entries in the expanded lookup table
+        segmented_lut_data: numpy.ndarray
+            Segmented lookup table data. Must be of type uint8 or uint16.
```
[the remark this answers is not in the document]
hackermd: Fully agreed. I was thinking that we could implement an alternative constructor that would create an instance of the class similar to [matplotlib.colors.LinearSegmentedColormap.from_list()](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.LinearSegmentedColormap.html#matplotlib.colors.LinearSegmentedColormap.from_list). However, the way segemented colormaps are specified in matplotlib differs from the way it's being done in DICOM, and then I thought it may just add further confusion.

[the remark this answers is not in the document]
hackermd: It's hard to check the validity of `segmented_lut_data` without parsing it in its entirety. Currently, this is done lazily when the `lut_data` property is accessed. We could consider doing this upon construction, but it could take some time. What do you think?

src/highdicom/content.py
```diff
@@ -1478,15 +1503,318 @@ def __init__(
...
+        color: str
+    ):
+        """
+
+        Parameters
+        ----------
+        first_mapped_value: int
+            Pixel value that will be mapped to the first value in the
+            lookup table.
+        number_of_entries: int
```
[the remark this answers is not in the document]
hackermd: We may be able to deduce it from `segmented_lut_data`. However, we would then need to expand the data first (loop over all entries of the segmented LUT data). It may also be good to let the user specify the intended number of entries in the expanded LUT data and use this information to subsequently check whether the LUT data has been encoded correctly. Thoughts?

src/highdicom/content.py
```diff
@@ -1478,15 +1503,318 @@ def __init__(
...
+
+        setattr(
+            self,
+            f'Segmented{self._attr_name_prefix}Data',
+            segmented_lut_data.tobytes()
+        )
+        setattr(
+            self,
+            f'{self._attr_name_prefix}Descriptor',
+            [number_of_entries, int(first_mapped_value), bits_per_entry]
```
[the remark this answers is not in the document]
hackermd: Let me know whether you think this parameter should even exist (see comment above).

src/highdicom/content.py
```diff
@@ -1506,56 +1834,82 @@ def __init__(
...
         # Checks on inputs
-        all_luts = [red_lut, green_lut, blue_lut]
-        for lut in all_luts:
-            if not isinstance(lut, LUT):
+        self._color_luts = {
+            'Red': red_lut,
+            'Green': green_lut,
+            'Blue': blue_lut
+        }
+        for lut in self._color_luts.values():
```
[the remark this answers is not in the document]
hackermd: I am not sure. However, it don't see why they couldn't have different types. For example, if you encode a continuous colormap for a single color and the other colors remain constant, you may want to provide a segmented LUT for the colors that don't change and provide a LUT for the color that does change (in particular if the change is not linear).

src/highdicom/pr/content.py
```diff
@@ -737,3 +757,1295 @@ def __init__(
...
+                )
+
+        # Palette Color Lookup Table
+        _add_palette_color_lookup_table_attributes(
+            self,
+            palette_color_lut=palette_color_lut
+        )
+
+
+class BlendingDisplayInput(Dataset):
```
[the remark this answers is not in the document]
hackermd: I generally agree. However, since this sequence attribute exists in the standard, it may be further extended in the future (i.e., additional attributes may be added) and then this logic would break. Therefore, I would prefer keeping it as is, even it leads to slightly more verbose code.

src/highdicom/pr/content.py
hackermd: I think this is related to the issues with the current implementation of `ModalityLUT` versus `VOILUT`. The former implements an item of the Modality LUT Sequence (which is a `LUT` + one attribute), while the latter implements the VOI LUT macro, which includes the VOI LUT Sequence (where each item is just a `LUT`), **plus** additional attributes such as Window Center and Window Width.

Looking at this again, I think we should do the following:
- Implement `ModalityLUT` and `VOILUT` such that both represent items of the corresponding **sequence items** (for `VOILUT` that would mean just subclassing `LUT` without any additional parameters/attributes)
- Implement `ModalityTransformation` and `VOITransformation` which implement the **macros**, which includes additional attributes. In case of Modality, the macro includes Rescale Intercept and Rescale Slope and in case of VOI it includes Window Center and Window Width.

That would finally make sense in my opinion.

src/highdicom/pr/content.py
hackermd: Makes sense

hackermd: Addressed with 3718b14aa4581b7db0043ae787351c01eb207b83

src/highdicom/pr/content.py
hackermd: Yeah..you are probably right. I just thought if these "extract" functions return more than one value, then it may get messy returning a tuple and it may make more sense to bundle the extracted attributes in a dataset. Your call.

src/highdicom/pr/sop.py
hackermd: Removed with bebf627a19aefebf08c0dcfb9c9058cefb2c2204

src/highdicom/pr/sop.py
hackermd: Removed via a004405bfb943ef2cb63d6aaf52cba79fb227d21

src/highdicom/pr/sop.py
hackermd: I think it would have to be `int(item.BlendingInputNumber) != (i + 1)`, but otherwise makes total sense.

src/highdicom/pr/content.py
CPBridge: In my opinion there are no issues with the current implementation, it is the logical class hierarchy given how things are laid out in the standard, optimizes code re-use and maintainability, and all terminology matches the standard precisely. However the terminology in the standard is highly confusing. The only question is whether we can eliminate some confusion by breaking with the standard and renaming some of the classes because we think that it will be easier for users to understand that way. I'm not completely opposed to this, but I am concerned that by breaking with the terminology in the standard we might cause even confusion for those who are already familiar with it. Note for example that `pydicom` has functions to `apply_modality_lut` and `apply_voi_lut` even though these operations actually apply the relevant *transformation* regardless of whether it is a LUT or not. So users of pydicom may well already be used to the confusing terminology.

Most of the confusion here is due to the naming of what is currently the `VOILUT` class. I definitely think that having this class as implemented makes sense (but maybe under different name). It has two purposes:
- It represents an item of the "FrameVOILUTSequence" in the per-frame functional groups in things like parametric map. This actually represents a transformation, not a LUT.
- It serves as a base class for the `SoftcopyVOILUT` class, which itself represents items of the "SoftcopyVOILUTSequence". Again, this actually represents a transformation, not a LUT.

With these purposes in mind, I looked for the common ground and called it `VOILUT`. I did not consider that this causes confusion with items of the `VOILUTSequence`, which appears under items of both "FrameVOILUTSequence" and "SoftcopyVOILUTSequence" and contains simply LUTs. So `VOILUT` is probably a bad name for this reason. One option is `VOITransformation`, but that seems odd as it would go into sequences called `...LUTSequence`.  The question in that case becomes what to do with it's base class `SoftcopyVOILUT`, should that become `SoftcopyVOITransformation`, despite being placed as items within "SoftcopyVOILUTSequence". Another option is `FrameVOILUT`, since this is the concrete item that it actually represents, then keep its subclass as it.

I do not see a purpose to creating a `ModalityTransformation` class. It does not currently exist and as far as I can tell this collection of elements does not exist anywhere within the standard except as a *module* called "Modality LUT" that places constraints saying that you must have either the rescale parameters or at LUT to represent a modality transform. It does not exist as an item within a sequence anywhere, as far as I am aware? I suppose we could implement it and have users create it and pass it to constructors, and then copy over the various attributes to the root level of the SOPClass dataset, like we have done for `PaletteColorLookupTable`, but I don't really see a strong reason why this would be better than how things are implemented currently. In my opinion it might make some sense to rename our private utility functions like `_add_modality_lut_attributes` -> `_add_modality_transformation_attributes`, but here we are definitely talking about the "Modality LUT" module, so maybe it's clearer as it?

I really don't see any reason to create a `VOILUT` class that is just a subclass of `LUT` without any changes. What is wrong with just passing a `LUT` where this is needed? I think the confusion here is that the existing `VOILUT` class does not do what you think.

CPBridge: So in summary, I think we should either:
1. Rename `VOILUT` -> `FrameVOILUT`, or
2. Rename `VOILUT` -> `VOITransformation` AND `SoftcopyVOILUT` -> `SoftcopyVOITransformation` 

and leave it at that

src/highdicom/pr/sop.py
CPBridge: That is accounted for in the suggestion by starting the enumeration at 1

src/highdicom/pr/content.py
CPBridge: I would go with returning bytes. This is private so no need to overthink it

src/highdicom/content.py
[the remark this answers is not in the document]
CPBridge: Yeah, it's a mess. I suppose the duplicated code is quite simple so no big deal.

It's just quite funny because my initial implementation split duplicated this logic between PaletteColorLUT and LUT  and then you asked me to merge it to reduce duplication. And now you've duplicated it again >_<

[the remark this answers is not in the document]
CPBridge: Whatever you think is best I'll go with

src/highdicom/pr/content.py
hackermd: I think the major confusion comes from the fact that the same names are used for modules (collections of attributes, e.g., [Modality LUT Module Attributes](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.11.html#table_C.11-1)), macros (collections of attributes, e.g., [Modality LUT Macro Attributes](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.11.html#table_C.11-1b)), and individual attributes (e.g., [Modality LUT Sequence](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.11.html#para_8ed414d3-e3c3-449b-8407-6384126d0d32)).

It's even worse with VOI LUT, because there is a nesting of attributes with very similar names:
- [Softcopy VOI LUT Module Attributes](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.11.8.html#table_C.11.8-1)
- [Softcopy VOI LUT Sequence](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.11.8.html#para_f5f8feb7-aa86-4a93-aa6f-c9ec483d3b68)
- [VOI LUT Macro Attributes](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.11.2.html#table_C.11-2b)
- [VOI LUT Sequence](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.11.2.html#para_6cf459c0-7173-4ade-a142-9a28ee76ced9)

Now what does `VOILUT` implement? Currently, it implements the VOI LUT Macro Attributes. These are included in Softcopy VOI LUT Sequence, but should `SoftcopyVOILUT` therefore inherit from `VOILUT`?

So far, we have implemented sequence items by Python classes, but I am not aware that we implemented macros (we definitely didn't implement modules).

It's confusing as hell that `VOILUT` does not inherit from `LUT`, but `ModalityLUT` does, and that `VOILUT` contains the `WindowCenter` and `WindowWidth` attributes, but `ModalityLUT` does not contain `RescaleSlope` and `RescaleIntercept`. I think to get some sanity into all of this, we'll have to break with the standard.

The Modality LUT and VOI LUT information entities (yes, they also have the same name!) are part of the [Pixel Transformation Sequence](https://dicom.nema.org/medical/dicom/current/output/chtml/part04/sect_N.2.html). Therefore, I thought we could implement the **macros** by `ModalityLUTTransformation` and `VOILUTTransformation` Python classes.

src/highdicom/content.py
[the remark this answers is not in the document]
CPBridge: I don't think that equidistant colors like the matplotlib function makes sense, but passing `(value, color)` tuples probably would. I think we should do it but not hold up this PR over it, just point to the standard for now.

I think we should parse it upon construction. Most of highdicom's added value is making sure things are correct so I think we should not let though incorrect things here. We wouldn't need to actually build the output array in the constructor which may save some time

src/highdicom/pr/content.py
hackermd: The standard even uses the terms "Modality LUT transformation" and "VOI LUT transformation" (see sections [N.2.1.1 Modality LUT](https://dicom.nema.org/medical/dicom/current/output/chtml/part04/sect_N.2.html#sect_N.2.1.1) and [N.2.1.3 VOI LUT](https://dicom.nema.org/medical/dicom/current/output/chtml/part04/sect_N.2.html#sect_N.2.1.3), respectively).

src/highdicom/content.py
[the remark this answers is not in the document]
CPBridge: The [standard](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.9.html#table_C.7-22) says that, for example RedPaletteColorLUTData is

> Required if segmented data is NOT used in an Image IOD or Color Palette IOD, or if the IOD is a Presentation State IOD.

The use of "segmented data" rather than "red segmented data" to me suggests that all three are the same, but it's not very clear. Also are we sure that we can even use segmented data in a presentation state, given the above?

src/highdicom/content.py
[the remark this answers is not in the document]
CPBridge: No strong feelings but I would lean to towards deducing it from the data if we can without the check

CPBridge:
@hackermd it seems you have forgotten to add a test image called `sm_image_grayscale.dcm`, so tests are failing for me

src/highdicom/pr/content.py
hackermd: Ok. Here is the implementation of the above proposal: 4d2c34b39183ee421246198e7ddc7f0511cbdc65

There are now the following classes for the different LUTs and the corresponding pixel transformations:
- `ModalityLUT` (inherits from `LUT`)
- `ModalityLUTTransformation`
- `VOILUT` (inherits from `LUT`)
- `VOILUTTransformation`
- `PresentationLUT` (inherits from `LUT`)
- `PresentationLUTTransformation`
- `ColorPaletteLUT` and `SegmentedColorPaletteLUT` (do **not** inherit from `LUT` because they require different attributes, but they have the same highdicom interface, i.e., same properties)
- `ColorPaletteLUTTransformation`

In addition, there is a `SoftcopyVOILUTTransformation` (inherits from `VOILUTTransformation`).

src/highdicom/content.py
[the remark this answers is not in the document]
hackermd: > It's just quite funny because my initial implementation split duplicated this logic between PaletteColorLUT and LUT and then you asked me to merge it to reduce duplication (though there were other issues). And now you've duplicated it again >_<

Yeah, I didn't appreciate the complexity initially..

src/highdicom/content.py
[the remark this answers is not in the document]
hackermd: I agree about the checks. We can try and see how long it takes.

src/highdicom/content.py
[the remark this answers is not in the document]
hackermd: If we parse it anyways in the constructor, then we can omit the parameter

src/highdicom/pr/content.py
hackermd: Addressed with 424cad25d733d1cca21c7997a42ddbd6732ff054

src/highdicom/pr/sop.py
hackermd: Oh...I missed that. Great!

src/highdicom/content.py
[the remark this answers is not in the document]
hackermd: Hmm...fine with me restricting it to either all or none being segmented.

We definitely need support for segmented LUT data in the presentation state. I will not include lots of LUTs with 16-bits entries. If the standard does not support that, then it should be corrected.

src/highdicom/content.py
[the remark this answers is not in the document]
hackermd: This discussion is no longer relevant (see https://github.com/herrmannlab/highdicom/pull/166#discussion_r855748603)

src/highdicom/content.py
```diff
@@ -1478,133 +1525,603 @@ def __init__(
...
+    device-indendent presentation values (P-Values).
+
+    """
+
+    def __init__(
+        self,
+        presentation_lut_shape: Optional[
+            Union[PresentationLUTShapeValues, str]
+        ] = None,
+        presentation_luts: Optional[Sequence[PresentationLUT]] = None,
```
CPBridge, line 1657: At least for GSPS, this sequence must be of length 1 (https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.11.6.html#table_C.11.6-1)

Suggest changing the type of this parameter to `Optional[PresentationLUT]`

src/highdicom/content.py
```diff
@@ -1478,133 +1525,603 @@ def __init__(
...
+        Parameters
+        ----------
+        presentation_lut_shape: Union[highdicom.pr.PresentationLUTShapeValues, str, None], optional
+            Shape of the presentation LUT, applied after the Softcopy
+            VOI LUT Transformation to create display values.
+        presentation_luts: Optional[Sequence[highdicom.LUT]], optional
+            LUTs for the presentation LUT, applied after the Softcopy
+            VOI LUT Transformation to create display values.
+
+        """  # noqa: E501
```
CPBridge, line 1670:
```suggestion
            VOI LUT Transformation to create display values.
            
        Note
        -----
        Only one of ``presentation_lut_shape`` or ``presentation_luts`` should be provided.

        """  # noqa: E501
```

May need to change this if the above parameter name changes

src/highdicom/pr/content.py
```diff
@@ -681,17 +702,26 @@ def _check_coords(
             )
 
 
-class SoftcopyVOILUT(VOILUT):
+class SoftcopyVOILUTTransformation(VOILUTTransformation):
```
CPBridge, line 705: I thought that you wanted to change this so that it no longer subclasses VOILUTTransformation but rather takes one as a parameter?

I think it's ok as is but up to you

src/highdicom/pr/sop.py
```diff
@@ -135,73 +144,36 @@ def __init__(
...
-            WindowWidth, WindowCenter, and VOILUTSequence), if any, in the
-            presentation state with values copied from the source images.
-        icc_profile: Union[bytes, None], optional
-            ICC color profile to include in the presentation state. If none is
-            provided, the profile will be copied from the referenced images.
-            The profile must follow the constraints listed in :dcm:`C.11.15
-            <part03/sect_C.11.15.html>`.
+        modality_lut_transformation: Union[highdicom.ModalityLUTTransformation, None], optional
+            Description of the Modality LUT Transformation for tranforming modality
+            dependent into modality independent pixel values
```
CPBridge, line 149:
```suggestion
            Description of the Modality LUT Transformation for transforming modality
            dependent into modality independent pixel values. If no value is provided,
            the modality transformation in the referenced images, if any, will be used.
```

src/highdicom/pr/sop.py
```diff
@@ -135,73 +144,36 @@ def __init__(
...
+        voi_lut_transformations: Union[Sequence[highdicom.pr.SoftcopyVOILUTTransformation], None], optional
+            Description of the VOI LUT Transformation for tranforming
+            modality pixel values into pixel values that are of interest to a
+            user or an application
```
CPBridge, line 153:
```suggestion
            Description of the VOI LUT Transformation for transforming
            modality pixel values into pixel values that are of interest to a
            user or an application. If no value is provided, the VOI LUT
            transformation in the referenced images, if any, will be
            used.
```

src/highdicom/pr/sop.py
CPBridge: same for parameters of the other classes

src/highdicom/pr/sop.py
CPBridge: same for parameters of other classes

src/highdicom/pr/content.py
hackermd: Yes, I initially thought it may be cleaning to call it `SoftcopyVOILUT` to make clear that it's an item of the Softcopy VOI LUT Sequence. However, it's more closely related to `VOILUTTranslation` than to `VOILUT` and therefore the name would be confusing.

The [Softcopy VOI LUT Module](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.11.8.html) contains the [VOI LUT Macro](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.11.2.html#table_C.11-2b) and additional attributes. Conceptually, the `SoftcopyVOILUTTransformation` is a special subtype of `VOILUTTransformation` that potentially restricts the transformation to a subset of images.

src/highdicom/content.py
hackermd: Addressed via fa147db72fd6598b2d9ee8e1f4e97cdbceff1362

src/highdicom/content.py
hackermd: Addressed via fa147db72fd6598b2d9ee8e1f4e97cdbceff1362

src/highdicom/content.py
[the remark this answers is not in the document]
hackermd: Addressed with b71085db106860137b7dbb0fc023eba44c636963

src/highdicom/content.py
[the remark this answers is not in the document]
hackermd: Addressed with f752b91a1aa7ded750d2c462d7de05c14b753f7b

src/highdicom/content.py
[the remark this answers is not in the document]
hackermd: Addressed with b71085db106860137b7dbb0fc023eba44c636963

PR #171, opened 2022-04-28
Change LUT word length

CPBridge:
We can continue to discuss the right thing to do here, but here's the PR to switch to storing 1 byte per entry when bits allocated is 8, if we decide to use it. I vote for merging this

### DIFF
--- a/src/highdicom/content.py
+++ b/src/highdicom/content.py
@@ -1586,8 +1586,7 @@ class LUT(Dataset):
             raise ValueError(
                 "Numpy array must have dtype uint8 or uint16."
             )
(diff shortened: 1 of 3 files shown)


hackermd:
I tend argue that we should only do this if we can set VR to `"OB"`. Otherwise, receiving applications will likely decode the data incorrectly.

However, that would require chancing the dictionary of pydicom and since that is auto-generated from the table in the standard, it would require a change in the standard.

I would leave it as is until we have further clarity. It's currently definitely not wrong. The VR is OW and the data is 16-bit.

CPBridge:
We have decided to disallow 8bit LUTs entirely for the time being. See #177. Closing this.

PR #176, opened 2022-05-06
Add factory method to construct UID from UUID

hackermd:


CPBridge (approved):
Looks ok to me. Curious about why the UID suffixes generated by pydicom are insufficient

hackermd:
> Curious about why the UID suffixes generated by pydicom are insufficient

What do you mean?

CPBridge:
What's lacking in the existing default constructor `UID()`, which uses pydicom to generate the suffix, such that this is useful?

hackermd:
You cannot just pass a UUID string to the constructor of `UID`. The `generate_uid()` function will internally generate a UUID if `prefix=None`. However, if you already have a UUID and you want to use that (instead of a new, randomly generated one), you need to construct the UID yourself. That's what this factory method does.

CPBridge:
I suppose... Not something that I'd ever thought of doing

hackermd:
For example, when converting annotations that have been stored in a database into DICOM format, the existing data may be associated with identifiers in form of UUIDs. It may then be desirable to "re-use" these identifiers in the resulting DICOM objects.

PR #177, opened 2022-05-14
Remove 8 bit LUTs

CPBridge:
Following considerable confusion about the details of the 8 bit LUTs in the presentation state PR (#139), this removes the ability to create or parse 8 bit LUTs until the uncertainty around them can be resolved. Following this, only 16 bit LUTs will be supported.


hackermd (approved):
Thanks @CPBridge! I think this is the sanest and safest approach for the time being.

PR #179, opened 2022-06-01
Fix bugs in implementation of Advanced Blending Presentation State IOD

hackermd:
There is currently a check that prevents reference of images from more than one series in presentation state instances. However, in case of the DICOM Advanced Blending Representation State IOD, images from more than one series may be blended together (as long as the images are defined in the same frame of reference). This PR relaxes the check to allow the referencing images from multiple series.

The PR also fixes the inclusion of attributes of the Palette Color Lookup Table Macro in the Advanced Blending Sequence. The attributes should be added to an item of the Palette Color Lookup Table Sequence rather than the item of the Advanced Blending Sequence.

### DIFF
--- a/src/highdicom/pr/content.py
+++ b/src/highdicom/pr/content.py
@@ -913,27 +913,23 @@ def _add_presentation_state_relationship_attributes(
     """
     # Assert referenced images are from the same series and have the same size
     ref_im = referenced_images[0]
-    ref_series_uid = ref_im.SeriesInstanceUID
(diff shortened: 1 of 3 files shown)


hackermd:
It doesn't exist yet :) I have been working with David on a correction proposal (CP) to include it into the standard. The changes in this PR are necessary to comply with the CP.

CPBridge (approved):
In that case, looks good

Issue #180, opened 2022-07-05
Provide meaningful error message if all Segmentation frames are empty


adamvonpaternos:
When using the static method of `_omit_empty_frames`, if there are no empty frames present (frames without positive pixels), then numpy thows an exception with regard to the empty `non_empty_frames` list passed into `np.stack`.

https://github.com/herrmannlab/highdicom/blob/95f39dd722ae6d357af3f942e2130d0ff4d68bfc/src/highdicom/seg/sop.py#L1037-L1078
 
May I suggest just a check on the array before proceeding? 

```
        if len(non_empty_frames) == 0:
            return (pixel_array, plane_positions, []) 
```

CPBridge:
Thanks for filing the bug report, @adamvonpaternos !

Though this seems like a simple bug, it actually gets at a more fundamental point...

If all the frames are empty, and the "omit_empty_frames" option is used, what exactly should the desired behaviour be?

I would argue that we should probably throw a more useful error message and leave it up to the user to check that the segmentation is non-empty before trying to construct a segmentation. I think creating a segmentation with no actual pixel data would be incorrect.

Thoughts @hackermd ?

CPBridge:
Actually @adamvonpaternos can you please confirm whether you are encountering this issue when there are no empty frames or all empty frames? Your text "if there are no empty frames present" suggests that there are "no empty frames" present, but your solution is a fix for "no non-empty frames" (i.e. all frames are empty).

hackermd:
> I would argue that we should probably throw a more useful error message and leave it up to the user to check that the segmentation is non-empty before trying to construct a segmentation. I think creating a segmentation with no actual pixel data would be incorrect.

I fully agree.

hackermd:
> Actually @adamvonpaternos can you please confirm whether you are encountering this issue when there are no empty frames or all empty frames? Your text "if there are no empty frames present" suggests that there are "no empty frames" present, but your solution is a fix for "no non-empty frames" (i.e. all frames are empty).

That's the conclusion I reached as well.

pieper:
> > I think creating a segmentation with no actual pixel data would be incorrect.
> 
> I fully agree.

I would think creating an empty segmentation is perfectly valid, meaning that none of the segmented structure was in the image.  Like passing an empty list is different than no list at all.

fedorov:
[...]

PR #181, opened 2022-07-06
Add warning when empty segmentation is passed with omit_empty_frames

CPBridge:
Addresses #180 

When the `omit_empty_frames` option is used for a Segmentation and an empty segmentation mask is passed (i.e. a mask with all zeros), the constructor will issue a `UserWarning` and ignore the `omit_empty_frames` option.

Add tests

### DIFF
--- a/src/highdicom/seg/sop.py
+++ b/src/highdicom/seg/sop.py
@@ -6,7 +6,6 @@ from os import PathLike
 from typing import (
     Any, cast, Dict, List, Optional, Sequence, Union, Tuple, BinaryIO
 )
-import warnings
 
 import numpy as np
 from pydicom.dataset import Dataset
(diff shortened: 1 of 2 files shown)


hackermd (commented):
Thanks @CPBridge! I am wondering whether it should be a user warning or a warning level log message. I would probably favor the latter, because it may be nice to know for which instance the warning is issued, because the fact that all frames are empty could be due to a problem in the application logic/code. What do you think?

CPBridge:
> I am wondering whether it should be a user warning or a warning level log message. I would probably favor the latter, because it may be nice to know for which instance the warning is issued, because the fact that all frames are empty could be due to a problem in the application logic/code

I have changed to a logger warning for now. But we should probably wait for a consensus on #180 before merging

hackermd (approved):

Issue #159, opened 2022-03-08
Segmentation not working for 2D X-ray mammography DICOMs


KarolBorkowski:
Generation of Segmentation (SEG) images doesn't work for mammography images due to required tags that are optional and usually not present. In particular: FrameOfReferenceUID, SliceThicknes, ImageOrientationPatient and ImagePositionPatient. I'm attaching an exemplary mammography DICOM file.

Thanks for considering it!

[test_dicom.zip](https://github.com/herrmannlab/highdicom/files/8206443/test_dicom.zip)

hackermd:
Thank @KarolBorkowski for reporting the issue and for providing a test file.

You may be able to create a `Segmentation` instance by providing `pixel_measures` to the constructor (see also https://github.com/herrmannlab/highdicom/issues/99).

KarolBorkowski:
pixel_measures is not enough since the rest of the listed tags is missing too. When I set them to some random values, the instance is generated, but I wonder how are this values used by dicom viewers and if it's safe to use any values.

fedorov:
I am not sure if anyone used SEG with X-Ray images. It is not impossible that there are some gaps or incorrect assumptions in the standard to support this use case. I want to make sure @dclunie is aware of this conversation to comment on the issues you are observing.

CPBridge:
I have never tried this myself. I had a very brief initial look into this and I do not yet have reason to believe that there are inconsistencies in the standard itself although it is certainly the case that we did not consider the case where the frame of reference uid (and hence the other attributes) are missing when we wrote the segmentation implementation. We can (and should) look into this but it won't be a totally straightforward fix.

> I wonder how are this values used by dicom viewers and if it's safe to use any values.

I imagine this is likely to cause issues with viewers, and regardless it would not be a good idea to populate this with random values unless this is in a completely isolated and controlled research setting.

CPBridge:
I have started a branch to address this issue, but unfortunately it's not so straightforward...

Currently highdicom uses a fixed list of Dimension Index Pointers (that describe the organization of the frames within the segmentation) depending on the coordinate system. For segmentations of images in the PATIENT coordinate system, this is:

- SegmentNumber
- ImagePositionPatient

For segmentations of image in the SLIDE coordinate system, this is:

- SegmentNumber
- Row Position in Total Pixel Matrix
- Column Position in Total Pixel Matrix
- X Offset in Slide Coordinate System
- Y Offset in Slide Coordinate System
[...]

PR #183, opened 2022-07-09
Allow segmentations of images without a frame of reference UID

CPBridge:
Addresses #159 

Following the discussion and extremely useful input from @fedorov and @dclunie on #159 this PR generalizes the `Segmentation` class to allow it to encode segmentations of source images that do not have frame of reference UIDs, such as projection radiography images.

There were two major changes required to make this possible.
(1) Prevent the inclusion of `PixelMeasuresSequence`, `PlanePositionSequence`/`PlanePositionSlideSequence` and `PlaneOrientationSequence` when the source images have no frame of reference within which these values could be defined.
(2) Alter the dimension indexing mechanism in the case of no frame of reference. 

The chosen approach for (2) is to allow only input series or multiframe images with a single instance/frame, and use a single dimension index value: Segment Number. This is a little limited but I imagine will cover the vast majority of use cases. Generalizing beyond this in the future (using mechanisms suggested by @dclunie in #159) may be possible but would dramatically increase complexity. Note that it does allow multiple segments of the (single) input image.

I have tested this with the test file kindly provided by @KarolBorkowski (who originally opened #159) and run the output through dciodvfy. This seems to work fine, apart from some compliance issues with the original file (notably I had to change the PatientSex in the source image file from 'U' to 'O', since 'U' is not a defined value for this attribute and highdicom raises an error as a result). I also tested on a CR image from the pydicom testfiles and that also seems to be working fine.

Any feedback from @KarolBorkowski would also be gratefully received.


src/highdicom/seg/sop.py
```diff
@@ -1036,8 +1124,8 @@ def _check_and_cast_pixel_array(
     @staticmethod
     def _omit_empty_frames(
         pixel_array: np.ndarray,
-        plane_positions: Sequence[PlanePositionSequence]
-    ) -> Tuple[np.ndarray, Sequence[PlanePositionSequence], List[int]]:
+        plane_positions: Sequence[Optional[PlanePositionSequence]]
+    ) -> Tuple[np.ndarray, List[Optional[PlanePositionSequence]], List[int]]:
```
hackermd, line 1128: Should maybe the entire list be optional? `Optional[Sequence[PlanePositionSequence]]`

hackermd (approved):
Thanks @CPBridge! This looks great. It turned out to be much more challenging than initially anticipated. I only have one very minor suggestion.

src/highdicom/seg/sop.py
CPBridge: I agree that that would seem more obvious on the face of it, especially if it were part of the public API. But actually I think it would result in quite considerably more complex branching logic both inside this method and in the surrounding code that calls it, e.g. zipped/enumerated for loops using this list. The way things are now, the same processes can continue with a list of plane position items without having to worry about whether or not those items are meaningful, and then only worry at the end about whether or not to actually include the plane positions in the dataset.

Either way is a bit ugly, but I think I favour this way

PR #184, opened 2022-07-15
Add support for instance segmentation with LABELED segmentation type

hackermd:
This PR introduces changes to support storing multiple connected components for the same segment (labeled mask) in a DICOM Segmentation instance. This is currently not supported by the DICOM standard, but is necessary for many image analysis use cases in computational pathology (e.g., instance segmentation of single cells in whole slide microscopy images).

To support these use cases, all that would be necessary is a new segmentation type, which I called "LABELED".

Since we need to encode a potentially large number of instances (e.g., a whole slide image may easily contain > 100, 000 cells), support for larger bit depths (16-bit and 32-bit) is also required. This PR therefore also introduces changes to allow for encoding of segments with 32-bit unsigned integer pixels (currently restricted to "LABELED" segmentation type to not break existing behavior).

cc @fedorov @dclunie


CPBridge (commented):
Have not had a chance to follow through all the logic yet. I believe there may be some specific situations where this implementation will not behave as expected. But I'm posting the comments that I have so far

hackermd:
We may want to define a separate Instance Segmentation IOD and implement it as a separate `highdicom.seg.InstanceSegmentation` class.

PR #185, opened 2022-07-18
Minor fix in ImageFIleReader example

Eschivo:
Added highdicom.io import in ImageFileReader example in order to have a fully functioning example.


CPBridge (approved):
Good catch, thanks for the contribution!

PR #190, opened 2022-07-20
Improve check for coordinates in annotations group

hackermd:
Currently, the constructor of `highdicom.ann.AnnotationGroup` checks whether the coordinates passed via `graphic_data` are structured correctly. Specifically, it checks whether the first and last point of a POLYGON are identical (this is required by SR but not by ANN). However, it does not directly compare the floating point values for equivalence, but checks whether they are close. This can lead to an exception in case the coordinates are indeed very close to each other.


CPBridge (approved):
makes sense

PR #191, opened 2022-07-23
Compute plane positions more efficiently for TILED_FULL dimension organization

hackermd:
If the Dimension Organization is `TILED_FULL`, the position of frames in a multi-frame image needs to be computed. The current implementation of the `compute_plane_position_slide_per_frame()` function performs this computation in an inefficient manner by calling `compute_plane_position_tiled_full()` for each frame, which internally constructs a `PixelToReferenceTransformer` object for every invocation. This PR improves the performance of the computation by constructing a `PixelToReferenceTransformer` object once and then re-using the object repeatedly. I tested this on a larger VL Whole Slide Microscopy Image instance and was able to speed up the computation from 2.38 seconds to 1.65 seconds.

### DIFF
--- a/src/highdicom/utils.py
+++ b/src/highdicom/utils.py
@@ -6,7 +6,10 @@ from pydicom.dataset import Dataset
 
 from highdicom.content import PlanePositionSequence
 from highdicom.enum import CoordinateSystemNames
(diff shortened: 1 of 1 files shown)


CPBridge (approved):
Looks good

PR #192, opened 2022-07-23
Fix decoding of graphic data in annotations group

hackermd:
This PR fixes the decoding of the `DoublePointCoordinatesData` or `PointCoordinatesData` of an Annotation Group of a Microscopy Bulk Simple Annotations instance and adds unit tests to avoid future regression.


CPBridge (approved):
Looks good

PR #193, opened 2022-08-02
Allow annotations to reference multiple images

hackermd:
An annotations object may contain multiple measurements, which may have been derived from different images. Therefore, it is important to allow reference of multiple (multi-frame) image instances.

### DIFF
--- a/src/highdicom/ann/sop.py
+++ b/src/highdicom/ann/sop.py
@@ -99,12 +99,6 @@ class MicroscopyBulkSimpleAnnotations(SOPClass):
 
         """  # noqa: E501
         src_img = source_images[0]
-        is_multiframe = hasattr(src_img, 'NumberOfFrames')
-        if is_multiframe and len(source_images) > 1:
-            raise ValueError(
-                'Only one source image should be provided in case images '
-                'are multi-frame images.'
-            )
 
         supported_transfer_syntaxes = {
             ImplicitVRLittleEndian,
--- a/src/highdicom/ann/sop.py
+++ b/src/highdicom/ann/sop.py
@@ -105,7 +105,7 @@ class MicroscopyBulkSimpleAnnotations(SOPClass):
         coordinate_type = AnnotationCoordinateTypeValues(
             annotation_coordinate_type
         )
-        if len(set([img.FrameOfReferenceUID for img in source_images])) > 1:
+        if len({img.FrameOfReferenceUID for img in source_images}) > 1:
             raise ValueError(
                 'All source images must have the same Frame of Reference UID.'
             )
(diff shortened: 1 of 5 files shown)


CPBridge (commented):
I have not read the relevant parts of the standard in detail, but this seems potentially too lax and allow the creation of objects whose meaning is ambiguous. In the case where multiple images are passed, in which image's coordinate system are the annotation graphic data defined? Should there not be a check that the different images have the same frame of reference UID (for SCOORD3D) and/or number of rows and columns (for SCOORD)?

hackermd:
> Should there not be a check that the different images have the same frame of reference UID (for SCOORD3D) and/or number of rows and columns (for SCOORD)?

That's a good idea.

I realized that we will need to differentiate between the [Referenced Image Sequence](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.37.html#para_14e0c12c-b76f-4662-aa08-6ff5904a8e9a) (Microscopy Bulk Simple Annotations Module) and [Referenced Instance Sequence](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.12.2.html#para_0eb004ad-037a-411e-a89b-42cdc1557e3a) (Common Instance Reference Module). At the moment, we are including a reference for each of the `source_images` in both sequences.

In my opinion, the design of the Microscopy Bulk Simple Annotations Module is flawed for the use of 2D annotation coordinate type (SCOORD), where coordinates are defined relative to the total pixel matrix of an individual SOP instance, because the assumption that a group of annotations applies to a single image instance may not hold - in particular if measurements are included. For example, objects may be detected in one image and the coordinates will be recorded relative to that image, but then measurements may subsequently be taken from a different image, which will have a different SOP Instance UID but the same Frame of Reference UID. Therefore, I only use 3D annotation coordinate type, where coordinates are defined relative to the frame of reference.

In conclusion, I think we should apply different checks for different annotation coordinate type values. I will update the PR accordingly.

src/highdicom/ann/sop.py
```diff
@@ -98,6 +102,21 @@ def __init__(
             of `highdicom.base.SOPClass`
 
         """  # noqa: E501
+        coordinate_type = AnnotationCoordinateTypeValues(
+            annotation_coordinate_type
+        )
+        if len(set([img.FrameOfReferenceUID for img in source_images])) > 1:
```
CPBridge, line 108: Set comprehension :o

```suggestion
        if len({img.FrameOfReferenceUID for img in source_images}) > 1:
```

CPBridge (approved):
Looks good, just one tiny suggestion

PR #194, opened 2022-08-05
Allow reference of optical path for annotations measurements

hackermd:
In case of intensity measurements for annotations, the microscopy image (and the optical path) need to be specified for the measurements. This PR adds allows including a reference to the images to which the measurements apply.

cc @dclunie

### DIFF
--- a/src/highdicom/ann/content.py
+++ b/src/highdicom/ann/content.py
@@ -11,7 +11,10 @@ from highdicom.ann.enum import (
     AnnotationGroupGenerationTypeValues,
     GraphicTypeValues,
 )
-from highdicom.content import AlgorithmIdentificationSequence
+from highdicom.content import (
+    AlgorithmIdentificationSequence,
+    ReferencedImageSequence,
+)
 from highdicom.sr.coding import CodedConcept
 from highdicom.uid import UID
 from highdicom._module_utils import check_required_attributes
@@ -25,7 +28,8 @@ class Measurements(Dataset):
         self,
         name: Union[Code, CodedConcept],
         values: np.ndarray,
-        unit: Union[Code, CodedConcept]
+        unit: Union[Code, CodedConcept],
+        referenced_images: Optional[ReferencedImageSequence] = None
     ) -> None:
         """
         Parameters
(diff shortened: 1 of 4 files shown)


src/highdicom/ann/content.py
```diff
@@ -71,6 +94,14 @@ def unit(self) -> CodedConcept:
         """highdicom.sr.CodedConcept: coded unit"""
         return self.MeasurementUnitsCodeSequence[0]
 
+    @property
+    def referenced_images(self) -> Union[ReferencedImageSequence, None]:
+        """Union[highdicom.ReferencedImageSequence, None]: referenced images"""
+        if hasattr(self, 'ReferencedImageSequence'):
+            return self.ReferencedImageSequence
```
CPBridge, line 101: Currently this will not return the type suggested by the type hint. I think you need this:

```suggestion
            return ReferencedImageSequence.from_sequence(self.ReferencedImageSequence)
```

src/highdicom/ann/content.py
```diff
@@ -633,7 +675,10 @@ def get_measurements(
         self,
         name: Optional[Union[Code, CodedConcept]] = None
     ) -> Tuple[
-        List[CodedConcept], np.ndarray, List[CodedConcept]
+        List[CodedConcept],
+        np.ndarray,
+        List[CodedConcept],
+        List[Union[ReferencedImageSequence, None]]
```
CPBridge, line 681: NB backwards incompatible change

CPBridge (commented):
This looks good to me except one minor error I spotted. Though if I remember correctly, we were waiting on a decision about changes to the standard?

There are some changes in this PR that I think we should have in master now, regardless of what happens to the optical path stuff. Specifically the changes to the docstring of the `ReferencedImageSequence` (which replace the previous versions that are basically wrong) and the addition of the `from_sequence` method for that class.

Should we break this into 2 PRs and get these changes merged ASAP?

hackermd:
> This looks good to me except one minor error I spotted. Though if I remember correctly, we were waiting on a decision about changes to the standard?

We have submitted a Correction Proposal and are currently discussing it with the DICOM Standard Committee. cc @dclunie

PR #195, opened 2022-08-14
Minor fixes to segmentation

CPBridge:
As I was reviewing #184 I noticed some minor errors in the implementation of segmentation. Just some checks that are redundant and make the code more confusing to follow as well as the typo you found in the examples. I thought I might as well push these now

@hackermd

### DIFF
--- a/src/highdicom/seg/sop.py
+++ b/src/highdicom/seg/sop.py
@@ -312,6 +312,8 @@ class Segmentation(SOPClass):
 
         if pixel_array.ndim == 2:
             pixel_array = pixel_array[np.newaxis, ...]
+        if pixel_array.ndim not in [3, 4]:
+            raise ValueError('Pixel array must be a 2D, 3D, or 4D array.')
 
         super().__init__(
             study_instance_uid=src_img.StudyInstanceUID,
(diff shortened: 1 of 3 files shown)


src/highdicom/seg/sop.py
```diff
@@ -568,17 +570,6 @@ def __init__(
...
-        if pixel_array.ndim == 2:
-            pixel_array = pixel_array[np.newaxis, ...]
-        if pixel_array.ndim not in [3, 4]:
-            raise ValueError('Pixel array must be a 2D, 3D, or 4D array.')
-
-        if pixel_array.shape[1:3] != (self.Rows, self.Columns):
-            raise ValueError(
-                'Pixel array representing segments has the wrong number of '
-                'rows and columns.'
-            )
```
CPBridge, line 580: This is true by construction of the self.Rows and self.Columns, so the check is pointless

src/highdicom/seg/sop.py
```diff
@@ -568,17 +570,6 @@ def __init__(
         self.NumberOfFrames = 0
         self.PerFrameFunctionalGroupsSequence: List[Dataset] = []
 
-        if pixel_array.ndim == 2:
-            pixel_array = pixel_array[np.newaxis, ...]
```
CPBridge, line 572: This is duplicated from above

src/highdicom/seg/sop.py
```diff
@@ -568,17 +570,6 @@ def __init__(
...
-        if pixel_array.ndim not in [3, 4]:
-            raise ValueError('Pixel array must be a 2D, 3D, or 4D array.')
```
CPBridge, line 574: Moving this earlier simplifies things a bit

hackermd (approved):

Issue #188, opened 2022-07-19
compatibility with `coverage`?


wyli:
I came across this issue when importing highdicom with `coverage`:
```bash
pip install highdicom
echo "import highdicom" > my_test.py
coverage run my_test.py
```
it seems the import always takes a long time, compared with other python packages such as numpy.

here is a minimal example in colab:
https://colab.research.google.com/drive/18qCwaHr3CWhvTmdY8dZKA2Pn0dDzlmvR
![Screenshot 2022-07-19 at 14 06 27](https://user-images.githubusercontent.com/831580/179757280-9ac62d08-062f-4f9b-975d-b6434c9693ec.png)

CPBridge:
This is speculation but I imagine that this is probably due to just a couple of files within highdicom. The package contains a large part of the DICOM standard specification scraped from the standard itself and placed into nested Python list/dictionary structures within the modules `highdicom._modules` and (to a lesser extent) `highdicom._iods`. These files are therefore very large and can cause coverage tools to work very slowly (especially the first time, when many build a cache that will speed up subsequent runs).

I'm not an expert on `coverage`, but is there some way to manually exclude those files?

wyli:
Thanks, that's probably the issue. I couldn't find a way to exclude the files. on the other hand, I think it might be useful for highdicom to consider other possibilities of packaging that static standard specification? and perhaps improve the import by loading the module attributes lazily only when the users request them?

CPBridge:
@wyli see #196 for a solution to your issue

wyli:
> @wyli see #196 for a solution to your issue

thanks!

CPBridge:
There is still room for improvement here, but marking this immediate issue as resolved

wyli:
thank you, in the colab notebook now it works fine!
```
!pip install coverage
!pip install git+https://github.com/herrmannlab/highdicom.git
!echo "import highdicom" > my_test.py
!coverage run my_test.py
```

PR #196, opened 2022-08-14
Remove module level imports of module data to speed up module import

CPBridge:
As noted in #188, importing highdicom requires importing the very large data structures in `highdicom._iods` and `highdicom._modules`. This slows down imports somewhat, but more importantly slows down coverage based tests like `coverage run` very considerably when often this is unnecessary because the data will not actually be used.

This PR swaps all imports from these modules to be lazily evaluated. With this change, the test that @wyli reported took an very long time now completes almost instantaneously. Of course coverage tests that run on functions that make use of the data in these modules will still be slow, but at least now simple tests will run faster.

As far as I'm aware, there are no significant downsides to this.

### DIFF
--- a/src/highdicom/_module_utils.py
+++ b/src/highdicom/_module_utils.py
@@ -3,13 +3,6 @@ from typing import Any, Dict, List, Optional, Sequence, Union
 
 from pydicom import Dataset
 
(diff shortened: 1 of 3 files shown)


CPBridge:
@hackermd should we consider moving this data to a JSON file (or similar) so that no coverage tool will try to assess coverage of these files?

hackermd (approved):
Thanks @CPBridge

PR #197, opened 2022-08-15
Fix typo in parameter type in Segment Description docstring

CPBridge:


CPBridge:
I just moved this to #195 instead

PR #201, opened 2022-10-10
Make pylibjpeg-libjpeg optional

CPBridge:
Highdicom is licensed under the permissive MIT license to allow users to use in their own applications, regardless of licence.

Unfortunately, it currently depends on `pylibjpeg-libjpeg`, which is GPL v3 licensed. This effectively means that many users cannot use highdicom. Luckily highdicom only uses `pylibjpeg-libjpeg` indirectly through pydicom and only in a few specific cases (decoding pixel data with a few specific transfer syntaxes).

This PR moves `pylibjpeg-libjpeg` to an optional dependency, that users may install with `pip install highdicom[libjpeg]`, and adds documentation to this effect.

It also alters the tests so that the relevant tests are automatically skipped if `pylibjpeg-libjpeg` is not installed. This necessitated a (long overdue) restructuring of the segmentation tests to split for loops within a single test to parameterized tests (using fixtures).

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -22,13 +22,10 @@ Pre-build package available at PyPi:
 
     pip install highdicom
 
-Like the underlying ``pydicom`` package, highdicom relies on functionality
-implemented in the ``pylibjpeg-libjpeg``
-`package <https://pypi.org/project/pylibjpeg-libjpeg/>`_ for the decoding of
-DICOM images with certain transfer syntaxes. Since ``pylibjpeg-libjpeg`` is
-licensed under a copyleft GPL v3 license, it is not installed by default when
-you install highdicom. To install ``pylibjpeg-libjpeg`` along with highdicom,
-use
+The library relies on the underlying ``pydicom`` package for decoding of pixel data, which internally delegates the task to either the ``pillow`` or the ``pylibjpeg`` packages.
+Since the ``pillow`` is a dependency of *highdicom* and will automatically be installed, some transfer syntax can thus be readily decoded and encoded (baseline JPEG, JPEG-2000, JPEG-LS).
+Support for additional transfer syntaxes (e.g., lossless JPEG) requires installation of the ``pylibjpeg`` package as well as the ``pylibjpeg-libjpeg`` and ``pylibjpeg-openjpeg`` packages.
+Since ``pylibjpeg-libjpeg`` is licensed under a copyleft GPL v3 license, it is not installed by default when you install *highdicom*. To install the ``pylibjpeg`` packages along with *highdicom*, use
 
 .. code-block:: none
 
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -22,7 +22,19 @@ Pre-build package available at PyPi:
 
     pip install highdicom
 
-Source code available at Github:
+Like the underlying ``pydicom`` package, highdicom relies on functionality
+implemented in the ``pylibjpeg-libjpeg``
+`package <https://pypi.org/project/pylibjpeg-libjpeg/>`_ for the decoding of
+DICOM images with certain transfer syntaxes. Since ``pylibjpeg-libjpeg`` is
+licensed under a copyleft GPL v3 license, it is not installed by default when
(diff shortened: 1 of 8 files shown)


CPBridge:
I also expanded the CI tests to run the test suite with and without pylibjpeg-libjpeg. With it, the full tests are run. Without it, tests that depend on it are skipped.

docs/installation.rst
```diff
@@ -22,7 +22,19 @@ Pre-build package available at PyPi:
...
     pip install highdicom
 
-Source code available at Github:
+Like the underlying ``pydicom`` package, highdicom relies on functionality
+implemented in the ``pylibjpeg-libjpeg``
+`package <https://pypi.org/project/pylibjpeg-libjpeg/>`_ for the decoding of
+DICOM images with certain transfer syntaxes. Since ``pylibjpeg-libjpeg`` is
+licensed under a copyleft GPL v3 license, it is not installed by default when
+you install highdicom. To install ``pylibjpeg-libjpeg`` along with highdicom,
+use
```
hackermd, line 31:
```suggestion
The library relies on the underlying ``pydicom`` package for decoding of pixel data, which internally delegates the task to either the ``pillow`` or the ``pylibjpeg`` packages.
Since the ``pillow`` is a dependency of *highdicom* and will automatically be installed, some transfer syntax can thus be readily decoded and encoded (baseline JPEG, JPEG-2000, JPEG-LS).
Support for additional transfer syntaxes (e.g., lossless JPEG) requires installation of the ``pylibjpeg`` package as well as the ``pylibjpeg-libjpeg`` and ``pylibjpeg-openjpeg`` packages.
Since ``pylibjpeg-libjpeg`` is licensed under a copyleft GPL v3 license, it is not installed by default when you install *highdicom*. To install the ``pylibjpeg`` packages along with *highdicom*, use
```

setup.py
```diff
@@ -56,7 +56,9 @@ def get_version():
         'pillow>=8.3',
         'pillow-jpls>=1.0',
         'pylibjpeg>=1.4',
-        'pylibjpeg-libjpeg>=1.3',
         'pylibjpeg-openjpeg>=1.2',
```
hackermd, line 59: What do you think about making them optional as well (I have repeatedly had trouble installing them)? The `libjpeg` optional dependency flag would then install `pylibjpeg`, `pylibjpeg-libjpeg`, and `pylibjpeg-openjpeg`.

We should have basic support for JPEG-2000 through the Pillow library.

We may even make `pillow-jpls` optional, too. We would then provide basic support for baseline JPEG and JPEG-2000 through Pillow. Support for additional transfer syntaxes would require installing optional dependencies.

setup.py
```diff
@@ -56,7 +56,9 @@ def get_version():
...
     ],
+    extras_requires={
+        'libjpeg': ['pylibjpeg-libjpeg>=1.3'],
```
hackermd, line 62:
```suggestion
        'libjpeg': [
            'pylibjpeg>=1.4',
            'pylibjpeg-libjpeg>=1.3',
            'pylibjpeg-openjpeg>=1.2'
        ],
```

setup.py
```diff
@@ -56,7 +56,9 @@ def get_version():
         'pillow>=8.3',
         'pillow-jpls>=1.0',
         'pylibjpeg>=1.4',
```
hackermd, line 58:
```suggestion
```

hackermd, line 58:
```suggestion
```

setup.py
hackermd, line 62: Shall we call this just `jpeg` rather than `libjpeg`?

hackermd (changes requested):

setup.py
CPBridge: I agree that this probably makes sense

setup.py
CPBridge: I think I prefer it as it is. If `[jpeg]` is an optional dependency, it makes it seem like without the optional dependency, no JPEG can be encoded or decoded, which is not true

hackermd (approved):

Issue #203, opened 2022-10-28
SegmentedPropertyTypeModifierCodeSequence (left/right flag) lost in round trip


pieper:
Using [this public SEG dataset](https://www.dropbox.com/s/9eesbv22hsaj4sn/ABD_LYMPH_008_SEG.dcm?dl=0) I tried the following:

```
>>> seg = highdicom.seg.segread("/opt/data/idc/seg-performance/dicom-seg/full-body-seg/ABD_LYMPH_008 Dennis Bontempi/ABD_LYMPH_008_SEG.dcm")
>>> seg.save_as("/tmp/seg.dcm")
```

Comparing the dumps of the original (left) and the newly saved file (right) I notice that the left/right modifiers are missing as in the example below.  There are some other differences but I think they are inconsequential.

![image](https://user-images.githubusercontent.com/126077/198665921-ad09f360-1246-437c-afe5-9f1af0053139.png)

CPBridge:
Yikes, thanks for reporting this @pieper . Definitely a high priority to fix!

CPBridge:
Resolved with #205

PR #205, opened 2022-10-28
Use deepcopy for CodedConcept.from_dataset()

CPBridge:
This is a fix for #203 , @pieper 

The issue was because `CodedConcept.from_dataset()` manually created a new instance of the class, which means that any attributes of the underlying dataset not supported by the `CodedConcept` class are lost.

I also had to tweak the `__eq__` to defer to the superclass implementation as a fallback as this was creating bizarre bugs when trying to serialize the result. Pydicom-level functionality needs to be able to determine equality of a `CodedConcept` to any other `pydicom.Dataset` using its own rules.

Might want to do a patch release for this: potentially a nasty one...

### DIFF
--- a/src/highdicom/sr/coding.py
+++ b/src/highdicom/sr/coding.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 import logging
 from typing import Any, Optional, Union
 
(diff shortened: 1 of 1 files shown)


pieper:
Thanks @CPBridge !

hackermd (approved):

PR #207, opened 2022-11-14
Optionally skip deepcopy in `from_dataset` methods

CPBridge:
Tweaked all `from_dataset` methods to accept a new boolean `copy` parameter (default `True`) in order to skip the copy operation to improve performance. If `False`, the dataset parameter is altered in place. This is now used in the `segread` method, and dramatically improves the performance there.

@pieper this should cut around 25% off your large seg benchmark time based on my testing

### DIFF
--- a/src/highdicom/content.py
+++ b/src/highdicom/content.py
@@ -93,7 +93,8 @@ class AlgorithmIdentificationSequence(DataElementSequence):
     @classmethod
     def from_sequence(
         cls,
-        sequence: DataElementSequence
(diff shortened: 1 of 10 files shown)


CPBridge:
Note also, there were previously a bunch of `_from_dataset` methods that did this but did not expose it publicly. In these cases I have essentially kept the same behaviour but now made it public via the `copy` parameter to the public `from_dataset` method. The result is a consistent API for all `from_dataset` methods in the library

hackermd (approved):

Issue #202, opened 2022-10-21
Slow performance reading SEG with 98 segments


pieper:
I have this notebook testing the time to run the [seg example](https://highdicom.readthedocs.io/en/latest/usage.html#parsing-segmentation-seg-images) on a file with ~100 segments.  It takes over 8 minutes to load this file and print the simple metadata.

Is this expected?  Can we expect any optimization as the library matures?  Working with the same data in a research format such as nrrd is orders of magnitude faster.

@hackermd @fedorov

pieper:
Here is the notebook:

https://colab.research.google.com/drive/1ZLqJwDIO1XKnnjOzClkSq8RIawm3sp9M?usp=sharing

CPBridge:
Hi @pieper 

The `iter_segments` was an early solution for accessing segmentation pixels, but has largely been replaced by a set of methods on the Segmentation class itself. Unfortunately this has not yet made it into the User Guide yet, though I do have a branch on it. It is fairly well documented in the [API docs for the segmentation class](https://highdicom.readthedocs.io/en/latest/package.html#highdicom.seg.Segmentation.get_segment_numbers) however. Iterating through segments is not efficient nor convenient in many situations. Would something like this be more suited to your use case? This runs considerably faster for me (though there may still be room for performance gains):

```python
import pydicom
import highdicom as hd


seg = hd.seg.segread("/tmp/ABD_LYMPH_008_SEG.dcm")
print('read file')

source_uids = seg.get_source_image_uids()
source_sops = [uids[2] for uids in source_uids]

pix = seg.get_pixels_by_source_instance(
    source_sops,
    ignore_spatial_locations=True,
    #segment_numbers=[1,2,3]  # you could choose specific segments if you like
)

print(pix.shape)
```
There are also various methods to identify which segments you need ahead of calling this method (if this is useful). See the [get_segment_numbers](https://highdicom.readthedocs.io/en/latest/package.html#highdicom.seg.Segmentation.get_segment_numbers) method

Not saying we couldn't make iter_segments more efficient, but wondering whether there may now be a better fit for what you are actually trying to achieve.

pieper:
[...]

PR #208, opened 2022-12-05
Seg Array Optimization

CPBridge:
A few improvements focused on improving the efficiency of building segmentation arrays from loaded Segmentation objects.

This was achieved through:
- Re-ordering loops in the construction of the seg_frames_matrix to avoid searching through the large `_frames_lut` more than necessary.
- Creating a specialized case for the `combine_segments` option, rather than implementing it as a postprocessing of the standard routine. This reduces memory consumption significantly as it is now O(1) rather than O(n) in the number of segments requested.
- Being more careful with dtypes, and allowing the user to specify a dtype
- Adding an option to skip checks for overlapping segments (`skip_overlap_checks`), which results in a further modest speed-up.

@pieper you should now find that your example from #202 runs significantly faster and uses much less memory than before when `combine_segments=True` (especially in the case where the full set of segments is requested).

I went a little way down the path of JITing the methods with numba, but it actually slowed things down. There may be a bit more performance improvement possibly using threading, but we are hitting on some numpy limitations


pieper:
Thanks @CPBridge !  Let me know if you need me to test before you merge or i can just wait until @hackermd has had a look through it.

CPBridge:
Thanks Steve. If you do have time to test on your hardware at some point during the process that would be very helpful. I think before or after @hackermd has a look would be fine, I don't expect the core numerical process change

CPBridge:
A couple of other points to note:

- It will still be slow to load the file initially, that will require a fix in pydicom.
- Please try with and without the `skip_overlap_checks` option and see how this affects things

hackermd:
Thanks @CPBridge. I will take a look.

> It will still be slow to load the file initially, that will require a fix in pydicom

See https://github.com/pydicom/pydicom/pull/1734

src/highdicom/seg/sop.py
```diff
@@ -1916,81 +1984,140 @@ def _get_pixels_by_seg_frame(
...
+                    if seg_frame_num >= 1:
+                        seg_frame_ind = seg_frame_num.item() - 1
+                        # Copy data to to output array
+                        if self.pixel_array.ndim == 2:
+                            # Special case with a single segmentation frame
+                            out_array[out_frm, :, :, out_seg] = \
+                                self.pixel_array
+                        else:
+                            out_array[out_frm, :, :, out_seg] = \
+                                self.pixel_array[seg_frame_ind, :, :]
```
hackermd, line 2107: Does this repeatedly decode the pixel data? It may be advantageous to store `pixel_array` as a separate variable outside of the loop instead of repeatedly accessing the attribute/property on the `self` object.

src/highdicom/seg/sop.py
```diff
@@ -2303,49 +2444,59 @@ def get_pixels_by_source_instance(
...
-                        'referenced source frames. To return an empty '
-                        'segmentation mask in this situation, use the '
-                        "'assert_missing_frames_are_empty' parameter."
-                    )
-                    raise KeyError(msg) from e
+                msg = (
+                    f'SOP Instance UID {sop_uid} does not match any '
+                    'referenced source frames. To return an empty '
+                    'segmentation mask in this situation, use the '
+                    "'assert_missing_frames_are_empty' parameter."
```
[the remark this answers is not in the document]
hackermd: For consistency

src/highdicom/seg/sop.py
```diff
@@ -2303,49 +2444,59 @@ def get_pixels_by_source_instance(
...
+        # corresponding segmentation frame is required in the output, and if so,
+        # where
+        for seg_frm, lut_row in enumerate(lut):
+            # Source frame index and segment number for this LUT row
+            lut_src, lut_segno = lut_row.ravel()
+
+            # Find indices (if any) of these value in the lists of required
+            # output source frames and segments
+            out_frm_inds = np.where(frame_indices == lut_src)[0]
+            out_seg_inds = np.where(segment_numbers == lut_segno)[0]
```
hackermd, line 2475: I need to take a closer look, but at first glance I am wondering whether this is the most performant approach..

src/highdicom/seg/sop.py
```diff
@@ -2572,42 +2737,52 @@ def get_pixels_by_source_frame(
...
+                    'An unexpected error was encountered during '
+                    'indexing of the segmentation image. Please '
+                    'file a bug report with the highdicom repository.'
+                )
+
+            # Frame is only required if there is an index for both
+            if len(out_frm_inds) > 0 and len(out_seg_inds) > 0:
+                r = out_frm_inds[0]
+                c = out_seg_inds[0]
+                seg_frames[r, c] = seg_frm + 1
```
hackermd, line 2776: This appears to be a duplication. Can we factor this out into a function?

hackermd (approved):

hackermd (changes requested):
A few suggestions regarding style.

src/highdicom/seg/sop.py
```diff
@@ -1876,6 +1932,18 @@ def _get_pixels_by_seg_frame(
             each pixel lies in the range 0.0 to 1.0. If False, the raw integer
             values are returned. If the segmentation has BINARY type, this
             parameter has no effect.
+        skip_overlap_checks: bool
+            If True, skip checks for overlap between different segments. By
+            default, checks are performed to ensure that the segments do not
+            overlap. However, this reduces performance. If checks are skipped
+            and multiple segments do overlap, the segment with the highest
+            segment number will be placed into the output array.
+        dtype: Union[type, None]
```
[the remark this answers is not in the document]
CPBridge: Resolved in 4325153

src/highdicom/seg/sop.py
```diff
@@ -61,6 +61,60 @@
...
+    if max_val < 256:
+        dtype = np.uint8
+    elif max_val < 65536:
+        dtype = np.uint16
+    else:
+        dtype = np.uint32  # should be extremely unlikely
+    return dtype
+
+
+def _check_dtype(max_val: int, dtype: type) -> None:
```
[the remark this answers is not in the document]
CPBridge: Renamed to `_check_numpy_value_representation` in f80fd92, since "value representation" on its own may cause confusion with DICOM terminology

src/highdicom/seg/sop.py
CPBridge: No, this is cached on the Dataset at the pydicom level

src/highdicom/seg/sop.py
```diff
@@ -1916,81 +1984,140 @@ def _get_pixels_by_seg_frame(
...
+            for out_frm, frm_row in enumerate(seg_frames_matrix):
+                # Loop through segmentation frames
+                for out_seg, seg_frame_num in enumerate(frm_row):
+                    if seg_frame_num >= 1:
+                        seg_frame_ind = seg_frame_num.item() - 1
+                        # Copy data to to output array
+                        if self.pixel_array.ndim == 2:
+                            # Special case with a single segmentation frame
+                            out_array[out_frm, :, :, out_seg] = \
+                                self.pixel_array
```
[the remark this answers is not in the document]
CPBridge: Can you elaborate on your concern, not sure I understand?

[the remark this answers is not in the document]
hackermd: My concern is that NumPy may do a shallow copy and that `out_array` may continue to point to the array stored at `self.pixel_array` when it is returned by the function. With the references persisting, Python may not free the memory of `self.pixel_array` (even if the instance of the class is no longer needed).

It can sometimes be beneficial in my experience to enforce a deep copy for arrays that are returned from functions/methods.

https://numpy.org/doc/stable/user/quickstart.html#view-or-shallow-copy

src/highdicom/seg/sop.py
hackermd: Good to know. I would nevertheless suggest avoiding the repeated attribute access, since it may negatively impact performance. See https://github.com/pydicom/pydicom/issues/1728#issuecomment-1312833040

hackermd: In context of the above comment (https://github.com/herrmannlab/highdicom/pull/208#discussion_r1042309582), we may even want to create a deep copy outside of the loop.

src/highdicom/seg/sop.py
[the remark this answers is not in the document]
CPBridge: I'm almost certain that his isn't true after playing around with np.shared_memory but I added a .copy() just to be safe. Although the slicing operation could create a shallow copy, I'm pretty sure when you then assign this shallow copy as a value to a slice of the output array, the data is deepcopied. Otherwise the output array could consist of a bunch of different slices of other arrays, which is not allowed

src/highdicom/seg/sop.py
CPBridge: A deepcopy out of the loop would be hugely inefficient and the comment you point to doesn't apply here since the `.pixel_array` exists as a Python-level property rather than the loopkup being deferred to pydicom.Dataset's dictionary lookup process

CPBridge:
Ok @hackermd @pieper I completely refactored the way the lookup table is stored and queried and the result is much more efficient and in my opinion much less confusing than current master and even the previously reviewed version on this branch (which was itself much more efficient than master).

Previously I had implemented a lookup table as a numpy array and then wrote some frankly quite hacky code to "query" the lookup table to find the information I needed. I realised at the time that I was essentially implementing a relational database in numpy and that this was not a good idea, but I didn't want to introduce a new dependency...

...but at some point since then I became aware that a basic sqlite client is in the Python standard library and that this could deal with in-memory databases. I have refactored the entire lookup operation to use the standard library sqlite to construct and store the lookup table. Using sql syntax is much more ergonomic and in certain situations much more efficient than my nasty numpy implementation. In particular small queries are now very very fast. And all with no new dependencies. I'm just sad I didn't think of this before.

We should also consider this sort of approach elsewhere, e.g. storing a LUT of elements of an SR.

Putting this in the context of reducing overall seg parsing time, this has dramatically improved the efficiency of the "pixel reconstruction step". I'm now getting times of < 1s on my macbook for the full set of frames and segments on @pieper 's segmentation from #202 if I use `skip_overlap_checks`. (If someone can think of a way to make those checks more efficient I would love some ideas...). I believe there are still some gains to be had in the `_build_luts` method to extract the information to place into the SQL database from the PerFramesFunctionalGroupSequence (actually entering the information into the SQL database is very fast). But this PR is getting big and I think that is a separate thread that we can leave for the future.

src/highdicom/seg/sop.py
CPBridge: See https://github.com/herrmannlab/highdicom/pull/208#issuecomment-1358495069

src/highdicom/seg/sop.py
CPBridge: See https://github.com/herrmannlab/highdicom/pull/208#issuecomment-1358495069

src/highdicom/seg/sop.py
```diff
@@ -1328,35 +1387,51 @@ def from_dataset(
...
-
-        self._source_sop_instance_uids = np.array(
-            list(self._ref_ins_lut.keys())
+        cur.executemany(
+            "INSERT INTO InstanceUIDs "
+            "(StudyInstanceUID, SeriesInstanceUID, SOPInstanceUID) "
+            "VALUES(?, ?, ?)",
+            instance_data,
         )
+        self._db.commit()
```
hackermd, line 1434: You may want to consider managing transactions using [context manager](https://docs.python.org/3/library/sqlite3.html#sqlite3-connection-context-manager).

src/highdicom/seg/sop.py
```diff
@@ -1372,11 +1447,12 @@ def _build_luts(self) -> None:
         index values.
 
         """
+        self._db = sqlite3.connect(":memory:")
```
hackermd, line 1450: You may need to be careful with the type of the `_db` attribute. We currently don't rely on mypy to check type annotations, but this would probably result in an error. See https://github.com/ImagingDataCommons/dicomweb-client/blob/3f91d8d8059a9f8fe117b1c28581ea814660478e/src/dicomweb_client/file.py#L560-L561

src/highdicom/seg/sop.py
hackermd: The same would apply to several other sections of the code.

src/highdicom/seg/sop.py
```diff
@@ -1942,7 +2044,7 @@ def _get_pixels_by_seg_frame(
             overlap. However, this reduces performance. If checks are skipped
             and multiple segments do overlap, the segment with the highest
             segment number will be placed into the output array.
-        dtype: Union[type, str, np.dtype, None]
+        dtype: Union[I
```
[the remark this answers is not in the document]
CPBridge: Oops, thanks for catching. Resolved in 6a647f9

src/highdicom/seg/sop.py
CPBridge: Not really sure what the specific concern is here. The linked code appears to need type annotations because the connection variable may be `None` and may change type during the course of the programme to `sqlite3.Connection`. However that is not the case here: the `self._db` is never `None` so the annotation is not needed. I also checked with mypy and it doesn't raise any errors related to this.

If your concern is that the name of the variable is ambiguous, I changed it to `_db_con` in 2d1752e, and also added an explicit type annotation.

src/highdicom/seg/sop.py
CPBridge: Nice, I didn't know about this. Implemented in b7f232f

src/highdicom/seg/sop.py
CPBridge: (I accidentally changed to `db_con` instead of `_db_con` 2d1752e and corrected in b7f232f)

src/highdicom/seg/sop.py
```diff
@@ -1386,35 +1387,51 @@ def from_dataset(
...
+        Instance UIDS for instances referenced in the segmentation.
 
         """
-        # Map sop uid to tuple (study uid, series uid, sop uid)
-        self._ref_ins_lut = OrderedDict()
+        cur = self._db.cursor()
+        cur.execute(
+            """
+                CREATE TABLE InstanceUIDs(
+                    SOPInstanceUID_id INTEGER PRIMARY KEY AUTOINCREMENT,
```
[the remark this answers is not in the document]
CPBridge: No it's not, but while prototyping I had it like this as I thought that using a pretty long string as a primary key (and more importantly as a foreign key in joins) would be quite inefficient compared to using an integer. But I just tested it and it seems to make no discernible difference, and it probably does make things more confusing, so I removed it in [f81e7b6](https://github.com/ImagingDataCommons/highdicom/pull/208/commits/f81e7b6b3020bd31b234049abbfd578bd67dea7c)

CPBridge:
@hackermd thanks for your comments, I have made some updates.

I also snuck in another change in [8e95876](https://github.com/ImagingDataCommons/highdicom/pull/208/commits/8e95876f85b2494f4845457f4704a719376cd22e) to change behaviour that annoyed me recently.

The code to extract a segmentation mask for a list of frames is very conservative: if you give it a frame number that is not referenced by the segmentation it will by default error out. This makes sense when thinking about SOPInstanceUIDs of single frame referenced images. But in the multiframe referenced image case, I think it reasonable to assume that if a requested frame number is not referenced in the segmentation image, but a frame with a frame number higher than the requested frame is referenced, the requested frame exists in the referenced image and the corresponding 2D mask should be returned as zeros.

src/highdicom/seg/sop.py
```diff
@@ -1823,38 +1920,118 @@ def segmented_property_types(self) -> List[CodedConcept]:
...
+
+        Parameters
+        ----------
+        sop_instance_uid: str
+            SOP Instance UID value to check.
+
+        Raises
+        ------
+        KeyError:
+            If the input sop_instance_uid is not in the database.
```
hackermd, line 1934:
```suggestion
            If the input `sop_instance_uid` is not in the database.
```

src/highdicom/seg/sop.py
```diff
@@ -1823,38 +1920,118 @@ def segmented_property_types(self) -> List[CodedConcept]:
...
 
-    def _get_src_uid_index(self, sop_instance_uid: str) -> int:
-        ind = np.argwhere(self._source_sop_instance_uids == sop_instance_uid)
-        if len(ind) == 0:
-            raise KeyError(
-                f'No such source frame: {sop_instance_uid}'
-            )
-        return ind.item()
+    def _check_sop_uid_exists(self, sop_instance_uid: str) -> None:
+        """Checks whether a SOP Instance UID is in the database.
```
hackermd, line 1924:
```suggestion
        """Check whether a SOP Instance UID is in the database.
```

src/highdicom/seg/sop.py
```diff
@@ -1328,35 +1387,50 @@ def from_dataset(
...
+                                ref_study.StudyInstanceUID,
+                                ref_series.SeriesInstanceUID,
+                                ref_ins.ReferencedSOPInstanceUID,
+                            )
                         )
 
-        self._source_sop_instance_uids = np.array(
-            list(self._ref_ins_lut.keys())
-        )
+        with self._db_con:
```
hackermd, line 1418: I wonder whether it would make sense to abstract the database out of the `SOPClass` into a separate (private) class. That class could handle the database connection and expose methods to insert, find, and retrieve information. I feel that would be cleaner and would facilitate testing.

hackermd: Such a database service class could also become useful for managing frame-level data access for other multi-frame SOP class instances in the future (Parametric Map, etc.).

src/highdicom/seg/sop.py
```diff
@@ -1823,38 +1920,118 @@ def segmented_property_types(self) -> List[CodedConcept]:
...
+
+    def _create_temporary_segment_table(
+        self,
+        segment_numbers: Sequence[int],
+        combine_segments: bool,
+        relabel: bool,
+    ) -> None:
+        """Create a temporary table of desired segment in the SQLite DB.
+
+        Assumes the self._db_con object is currently being used as a context
```
hackermd, line 1954: That seems to be quite an assumption and asking for troubling..

hackermd (commented):
Looks good to me overall. My only suggestion would be to factor out the database management logic from the SOP Class into a separate helper/service class.

src/highdicom/seg/sop.py
CPBridge: As of [8f5172f](https://github.com/ImagingDataCommons/highdicom/pull/208/commits/8f5172f2f692937eee71574ab81c4bc95e32d87e) the database related logic has been significantly refactored into a new `_SegDBManager` class

src/highdicom/seg/sop.py
CPBridge: This no longer applies after all the of the recent refactoring. All methods on the new `SegDBManager` class handle their own commits to the database

CPBridge:
@hackermd I have refactored out all the database logic and I think this is ready to go

src/highdicom/seg/sop.py
```diff
@@ -61,6 +75,723 @@
...
+        cur = self._db_con.cursor()
+        return {
+            r for r in
+            cur.execute(
+                f'SELECT DISTINCT {cols_str} FROM FrameLUT'
+            )
+        }
+
+    @contextmanager
+    def temp_table(
```
hackermd, line 421: Should this method and the following be private methods (i.e., starting with `_`)?

I would also consider renaming them `_generate_temp_table`. I generally prefer function names to start with a verb.

src/highdicom/seg/sop.py
```diff
@@ -61,6 +75,723 @@
...
+        Parameters
+        ----------
+        table_name: str
+            Name of the temporary table.
+        column_defs: Sequence[str]
+            SQL syntax strings defining each column in the temporary table, one
+            string per column.
+        column_data: Iterable[Sequence[Any]]
+            Column data to place into the table.
+
```
hackermd, line 442: Documentation of `Return` or `Yield` missing.

src/highdicom/seg/sop.py
```diff
@@ -61,6 +75,723 @@
...
+        # Yield execution to "with" block
+        yield
+
+        # Clean up table after user code executes
+        with self._db_con:
+            self._db_con.execute('DROP TABLE TemporarySegmentNumbers')
+
+    @contextmanager
+    def get_indices_by_source_instance(
```
hackermd, line 529:
```suggestion
    def iter_indices_by_source_instance(
```

src/highdicom/seg/sop.py
```diff
@@ -61,6 +75,723 @@
...
+        self,
+        source_sop_instance_uids: Sequence[str],
+        segment_numbers: Sequence[int],
+        combine_segments: bool = False,
+        relabel: bool = False,
+    ) -> Generator[Iterator[Tuple[int, int, int]], None, None]:
+        """Get frame indices to create a segmentation mask for given instances.
```
hackermd, line 536:
```suggestion
        """Iterate over segmentation frame indices for given source image instances.
```

src/highdicom/seg/sop.py
```diff
@@ -61,6 +75,723 @@
...
+
+        This yields an iterator to the underlying database result, and is
+        intended to be used as a context manager for efficiency.
+
+        Parameters
+        ----------
+        source_sop_instance_uids: str
+            SOP Instance UID of the source instances to for which segmentations
+            are requested.
```
hackermd, line 545:
```suggestion
            SOP Instance UID of the source instances for which segmentation
            image frames are requested.
```

src/highdicom/seg/sop.py
```diff
@@ -61,6 +75,723 @@
...
+        segment_numbers: Sequence[int]
+            Sequence containing segment numbers to include.
```
hackermd, line 547: I would avoid a description of the type. That information is already provided via the type declaration.

```suggestion
            Numbers of segments to include.
```

src/highdicom/seg/sop.py
```diff
@@ -61,6 +75,723 @@
...
+        ):
+            with self.temp_segment_table(
+                segment_numbers=segment_numbers,
+                combine_segments=combine_segments,
+                relabel=relabel
+            ):
+                yield self._db_con.execute(query)
+
+    @contextmanager
+    def get_indices_by_source_frame(
```
hackermd, line 616:
```suggestion
    def iter_indices_by_source_frame(
```

src/highdicom/seg/sop.py
```diff
@@ -61,6 +75,723 @@
...
+        self,
+        source_sop_instance_uid: str,
+        source_frame_numbers: Sequence[int],
+        segment_numbers: Sequence[int],
+        combine_segments: bool = False,
+        relabel: bool = False,
+    ) -> Generator[Iterator[Tuple[int, int, int]], None, None]:
+        """Get frame indices to create a segmentation mask for given frames.
```
hackermd, line 624:
```suggestion
        """Iterate over frame indices for given source image frames.
```

src/highdicom/seg/sop.py
hackermd: It may be helpful to describe somewhere why the two different methods are needed. I am assuming that the `iter_indices_by_source_instance()` is needed if a multi-frame segmentation image references multiple single-frame image instances and the `iter_indices_by_source_frame()` is needed if a multi-frame segmentation image references multiple frames of a single multi-frame image instance. Is that understanding correct?

src/highdicom/seg/sop.py
```diff
@@ -61,6 +75,723 @@
...
+        ):
+            with self.temp_segment_table(
+                segment_numbers=segment_numbers,
+                combine_segments=combine_segments,
+                relabel=relabel
+            ):
+                yield self._db_con.execute(query)
+
+    @contextmanager
+    def get_indices_by_dimension_index_values(
```
hackermd, line 706:
```suggestion
    def iter_indices_by_dimension_index_values(
```

src/highdicom/seg/sop.py
```diff
@@ -61,6 +75,723 @@
...
+        self,
+        dimension_index_values: Sequence[Sequence[int]],
+        dimension_index_pointers: Sequence[int],
+        segment_numbers: Sequence[int],
+        combine_segments: bool = False,
+        relabel: bool = False,
+    ) -> Generator[Iterator[Tuple[int, int, int]], None, None]:
+        """Get frame indices to create a segmentation mask for given frames.
```
hackermd, line 714:
```suggestion
        """Iterate over frame indices for given dimension index values.
```

src/highdicom/seg/sop.py
```diff
@@ -2228,6 +2982,19 @@ def get_pixels_by_source_instance(
...
             values are returned. If the segmentation has BINARY type, this
             parameter has no effect.
+        skip_overlap_checks: bool
+            If True, skip checks for overlap between different segments. By
+            default, checks are performed to ensure that the segments do not
+            overlap. However, this reduces performance. If checks are skipped
+            and multiple segments do overlap, the segment with the highest
+            segment number (after relabelling, if applicable) will be placed
+            into the output array.
+        dtype: Union[type, str, np.dtype, None]
```
hackermd, line 2992:
```suggestion
        dtype: Union[type, str, numpy.dtype, None]
```

src/highdicom/seg/sop.py
```diff
@@ -2285,68 +3052,46 @@ def get_pixels_by_source_instance(
...
+                indices_iterator=indices,
+                segment_numbers=np.array(segment_numbers),
+                combine_segments=combine_segments,
+                relabel=relabel,
+                rescale_fractional=rescale_fractional,
+                skip_overlap_checks=skip_overlap_checks,
+                dtype=dtype,
+            )
+
+        return output_array
```
hackermd, line 3095:
```suggestion
            return self._get_pixels_by_seg_frame(
                num_output_frames=len(source_sop_instance_uids),
                indices_iterator=indices,
                segment_numbers=np.array(segment_numbers),
                combine_segments=combine_segments,
                relabel=relabel,
                rescale_fractional=rescale_fractional,
                skip_overlap_checks=skip_overlap_checks,
                dtype=dtype,
            )

```

src/highdicom/seg/sop.py
```diff
@@ -2443,19 +3190,32 @@ def get_pixels_by_source_frame(
...
             values are returned. If the segmentation has BINARY type, this
             parameter has no effect.
+        skip_overlap_checks: bool
+            If True, skip checks for overlap between different segments. By
+            default, checks are performed to ensure that the segments do not
+            overlap. However, this reduces performance. If checks are skipped
+            and multiple segments do overlap, the segment with the highest
+            segment number (after relabelling, if applicable) will be placed
+            into the output array.
+        dtype: Union[type, str, np.dtype, None]
```
hackermd, line 3213:
```suggestion
        dtype: Union[type, str, numpy.dtype, None]
```

src/highdicom/seg/sop.py
```diff
@@ -2715,6 +3456,19 @@ def get_pixels_by_dimension_index_values(
...
```
hackermd, line 3466:
```suggestion
        dtype: Union[type, str, numpy.dtype, None]
```

CPBridge (commented):
@hackermd I have addressed the various comments, mostly just accepting all suggestions. I have attempted to improve the docstrings for the `_SegDBManager` class

hackermd (approved):

PR #211, opened 2022-12-28
Multivalue VOI window attributes in presentation states

### LINKED ISSUE #210: Failure to create GSPS when reference DICOM uses "MultiValue" for WindowCenter, WindowWidth
When I go to create a GSPS using hd.pr.GrayscaleSoftcopyPresentationState() I get an error for certain DICOM where WindowWidth and WindowCenter are not floats in the reference DICOM. 

Here's the error I get:
```
~/anaconda3/lib/python3.8/site-packages/highdicom/pr/content.py in _get_softcopy_voi_lut_transformations(referenced_images)
   1456 
   1457             if has_width:
-> 1458                 by_window[(
   1459                     ref_im.WindowWidth,
   1460                     ref_im.WindowCenter,

TypeError: unhashable type: 'MultiValue'
```

 I fixed it using the following code: 

``` Python
        if isinstance(image_dataset.WindowWidth, pydicom.multival.MultiValue):
            image_dataset.WindowWidth = image_dataset.WindowWidth[0]
        if isinstance(image_dataset.WindowCenter, pydicom.multival.MultiValue):
            image_dataset.WindowCenter = image_dataset.WindowCenter[0]
```

This fix worked for two offending DICOM, but there is one more where I'm still getting the error. To get rid of the error there, I created the GSPS with my own windowing settings (using the argument voi_lut_transformations=[SoftcopyVOILUTObject]).

CPBridge:
GSPS creation fails with some images that have value multiplicity > 1 for WindowWidth, WindowCenter and/or WindowCenterWidthExplanation, which is allowed in the standard but is fairly rare. This is because the code to copy the VOI from the source images was not designed to deal with these

This fix addresses this and should make it possible to correctly copy the VOI LUT transformation from source images with multiple values.

Resolves #210

### DIFF
--- a/src/highdicom/content.py
+++ b/src/highdicom/content.py
@@ -1867,7 +1867,7 @@ class VOILUTTransformation(Dataset):
                     )
                 for exp in window_explanation:
                     _check_long_string(exp)
-            self.WindowCenterWidthExplanation = window_explanation
+            self.WindowCenterWidthExplanation = list(window_explanation)
         if voi_lut_function is not None:
             if window_center is None:
                 raise TypeError(
--- a/src/highdicom/pr/content.py
+++ b/src/highdicom/pr/content.py
@@ -1397,16 +1397,29 @@ def _get_softcopy_voi_lut_transformations(
             for frame_number, frm_grp in enumerate(perframe_grps, 1):
                 if hasattr(frm_grp, 'FrameVOILUTSequence'):
                     voi_seq = frm_grp.FrameVOILUTSequence[0]
+
                     # Create unique ID for this VOI lookup as a tuple
                     # of the contents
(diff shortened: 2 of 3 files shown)


delton137:
Hi, I tested this (#211) and it fixes the issue! I ran it on 2 test DICOM that were giving errors before when I tried to create GSPS.

delton137:
I should have done more thorough testing before declaring victory.. I ran on a 3rd dicom and ran into a new error. See:  https://github.com/ImagingDataCommons/highdicom/issues/210#issuecomment-1367454424

delton137:
Got it working on a third DICOM now after fixing a problem on my end. Looks good.

hackermd (approved):

PR #217, opened 2023-02-19
Add Python 3.11 to unit tests matrix

CPBridge:
Add tests using Python 3.11 for all future pull requests using GitHub workflows


CPBridge:
Tests are currently failing because for python 3.11 the libjpeg packages are not available yet. I feel we should probably still merge this anyway and monitor the situation

hackermd (approved):

PR #219, opened 2023-03-08
Improve User Guide

CPBridge:
*Major* update to the user guide:

- The old user guide has now been moved to a "Quick Start" guide. I improved the SR section as it was out of date. I also added an ANN section.
- I added a "General Concepts" section, so far the only page there is "coding", but in the future I will likely add one on coordinates, and possibly other things.
- I have added new pages for each individual IOD with much more complete explanations. The following are finished (the remainder are placeholders for now):
    - SR
    - SEG
    - ANN

This branch has been going since 2021... It's way way past time to get this merged. Since this is purely doc changes and you are busy @hackermd , thoughts on just merging so that readthedocs gets updated and then you can review and tweak later?


hackermd (approved):
Very nice!

PR #220, opened 2023-03-27
Fix ReferenceImageSequence

CPBridge:
In the bulk microscopy annotations module, the ReferencedImageSequence is always set to an empty list, which is disallowed by the standard. This appears to be a simple typo/bug with a simple fix


hackermd (approved):
Good catch!

PR #227, opened 2023-05-15
Improve Segmentation Creation Efficiency

CPBridge:
Segmentation creation is slow when the number of frames is large. This PR contains several changes to improve this:

- Do not append directly to the PerFramesFunctionalGroupsSequence during construction. Surprisingly, this append operation was taking up a significant proportion of the time in each loop iteration. I changed this to append the `Dataset` to a Python list (rather than a `pydicom.Sequence`) and then cast the list to a `pydicom.Sequence` after the loop. The effect was a very dramatic speed up.
- There are several unnecessary dtype cast operations in the previous implementation. These are very expensive with large arrays. In this PR they are removed, and the original dtype is retained for longer, with the final cast to uint8 at the end.
- There are several checks on the input array, such as that the segment numbers match and that the entire array is not empty, that become very slow when the input array is large. I was able to speed some of these up by using alternative numpy operations, and combining intermediate results for some of the checks.
- The previous implementation of `_omit_empty_frames` actually altered the input array. While this makes things conceptually a bit simpler, it is an unnecessary operation on a potentially very large array. I change this to leave the input array unaltered and changed the later indexing logic to select the correct frame such that empty plane positions are ignored.

In the course of working on this, I discovered two bugs that I fixed:
- The previous implementation would not correctly apply the Maximum Fractional Value scaling in some situations. It seems that the tests had been fudged a little to let this through but I'm sure it's wrong. The tests have now been changed to enforce the correct behaviour.
- In the previous implementation, it was possible that a user would pass two or more distinct source images with the same plane position (according to the relevant dimension indices). In this situation, the segmentation frames would only be recorded for the first of the frames. In other words the implementation would silently do the wrong thing. While I suppose in principle it may be possible to allow multiple  segmentation frames with the same plane position but different SourceImageSequences, I imagine this use case is not common and not a high priority to support correctly. Therefore, in this PR I simply added a check that will raise an exception when this situation is encountered such that the library no longer silently fails as before.

I also factored out several parts of the complicated constructor to make it more readable and modular.

A further change I made was to demote some of the logger messages about frame omission etc from INFO to DEBUG. For large segmentations these messages become overwhelming and in my opinion do not belong at the INFO level.

I tested this new implementation on the large mulit-segment CT from #202 (around 650 frames with 98 segments), and saw a speed up of around 10x in creation time(!). Furthermore, I ensured that each individual change I made improved the efficiency (and not simply that the net effect of all changes improved efficiency).  

Note that when creating FRACTIONAL segmentations with a transfer syntax that requires frame compression, the efficiency gains found above are much less significant relative to the time required to compress the frames. I intend to add an option to parallelise frame compression in a future PR.


src/highdicom/seg/sop.py
```diff
@@ -1932,29 +1905,287 @@ def _omit_empty_frames(
...
+            Whether the source images have a frame of reference.
+        coordinate_system: Optional[highdicom.CoordinateSystemNames]
+            The type of coordinate system used (if any).
+
+        Returns
+        -------
+        index_values: List[int]
+            The dimension index values (except the segment number) for the
+            given plane.
+
```
hackermd, line 2037: Do we want to reference the `highdicom.seg.DimensionIndexSequence` class? That may help clarify some of the terms used.

src/highdicom/seg/sop.py
```diff
@@ -1932,29 +1905,287 @@ def _omit_empty_frames(
...
+        has_ref_frame_uid: bool
+            Whether the sources images have a frame of reference UID.
+        coordinate_system: Optional[highdicom.CoordinateSystemNames]
+            Coordinate system used, if any.
+
+        Returns
+        -------
+        pydicom.Dataset
+            Dataset representing the item of the
+            PerFrameFunctionalGroupsSequence for this segmentation frame.
```
hackermd, line 2124:
```suggestion
            Dataset representing the item of the
            Per Frame Functional Groups Sequence for this segmentation frame.
```

hackermd (changes requested):
I really like the refactoring effort and the breaking down of the large constructor method into smaller focused helper methods. I have only a few minor comments and suggestions.

src/highdicom/seg/sop.py
```diff
@@ -1793,10 +1592,178 @@ def _check_segment_numbers(described_segment_numbers: np.ndarray):
                 f'from 1. Found {described_segment_numbers[0]}. '
             )
 
+    @staticmethod
+    def _get_pixel_measures(
+        source_image: Dataset,
+        has_ref_frame_uid: bool,
```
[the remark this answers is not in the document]
CPBridge: No, this was just how it was before the refactoring and I didn't change the logic. This is better simplified, addressed in 5f66eb3

src/highdicom/seg/sop.py
```diff
@@ -1932,29 +1905,287 @@ def _omit_empty_frames(
...
+            Full segmentation array containing all segments.
+        segment_number: int
+            The segment of interest.
+        number_of_segments: int
+            Number of segments in the the segmentation.
+        segmentation_type: highdicom.seg.SegmentationTypeValues
+            Desired output segmentation type.
+        max_fractional_value: int
+            Value for scaling FRACTIONAL segmentations.
+
```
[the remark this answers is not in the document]
CPBridge: Good catch. Addressed in ce85cbd

src/highdicom/seg/sop.py
```diff
@@ -1932,29 +1905,287 @@ def _omit_empty_frames(
...
+                    segment_array *= int(max_fractional_value)
+
+        return segment_array
+
+    @staticmethod
+    def _get_dimension_index_values(
+        plane_index: int,
+        dimension_position_values: List[np.ndarray],
+        plane_position_values: np.ndarray,
+        has_ref_frame_uid: bool,
```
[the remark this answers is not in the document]
CPBridge: Actually, I think it's simpler just to take the no frame of reference uid case out of this function and back to the main `__init__` and avoid passing this at all. Addressed in 53a2392

src/highdicom/seg/sop.py
```diff
@@ -1932,29 +1905,287 @@ def _omit_empty_frames(
...
+        plane_position_values: numpy.ndarray
+            Plane positions of each plane.
+        has_ref_frame_uid: bool
+            Whether the source images have a frame of reference.
+        coordinate_system: Optional[highdicom.CoordinateSystemNames]
+            The type of coordinate system used (if any).
+
+        Returns
+        -------
+        index_values: List[int]
```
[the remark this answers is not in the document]
CPBridge: Addressed in 10340a8

src/highdicom/seg/sop.py
CPBridge: Addressed in 1774199

src/highdicom/seg/sop.py
```diff
@@ -1916,12 +1889,12 @@ def _check_and_cast_pixel_array(
     def _omit_empty_frames(
```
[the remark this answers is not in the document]
CPBridge: Indeed. The current behaviour predates this PR but your suggestion is definitely nice simplification that makes things cleaner. I implemented it in 5cbd9bf and also renamed this method to `_find_nonemtpy_frame_indices`, which now better describes what it actually does.

src/highdicom/seg/sop.py
```diff
@@ -1932,29 +1905,287 @@ def _omit_empty_frames(
...
+        """Get dimension index values for a frame.
+
+        Parameters
+        ----------
+        plane_index: int
+            Index of the plane in the sorted planes list.
+        dimension_position_values: List[numpy.ndarray]
+            Relative locations of each plane.
+        plane_position_values: numpy.ndarray
+            Plane positions of each plane.
```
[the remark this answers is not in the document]
CPBridge: Agreed. Trying to actually understand what all this does (rather than naively refactoring it keeping the logic the same) led me down quite a rabbit hole. I ended up decided that the current implementation, though correct, is needlessly complex as it maintains two lists of indices and different arrays sorted according to them. I refactored things to keep everything indexing the original list. I also changed some variable and (private) parameter names that were confusing in my opinion, and updated these docstrings. See 287047e

src/highdicom/seg/sop.py
```diff
@@ -1932,29 +1905,287 @@ def _omit_empty_frames(
...
+        number_of_segments: int
+            Number of segments in the the segmentation.
+        segmentation_type: highdicom.seg.SegmentationTypeValues
+            Desired output segmentation type.
+        max_fractional_value: int
+            Value for scaling FRACTIONAL segmentations.
+
+        """
+        if pixel_array.dtype in (np.float_, np.float32, np.float64):
+            # Based on the previous checks and casting, if we get here the
```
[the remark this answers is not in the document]
CPBridge: This sounds good the checks and casting operations in `_check_and_caste_pixel_array` are a *very* significant part of the time for the entire construction, especially when there is no compression to do. This is because they involve checks on the unique values in the array, overlap etc, on potentially very large array. Consequently we definitely do not want to be doing them once per segment. This does of course introduce some dependencies between this method (which runs in a loop for every segment) and `_check_and_caste_pixel_array` (which runs once before the loop and processes the pixel array). The encapsulation is not perfect. But I think perfect encapsulation is impossible while maintaining reasonable performance so it was not my goal. Grouping similar code into these private methods still significantly improves readability in my opinion. I added this comment here to try and document these dependencies, I also added a further comment in the docstring in c0d9f72 to make this dependency even more explicit

src/highdicom/seg/sop.py
```diff
@@ -1419,316 +1397,157 @@ def __init__(
...
+                    # Concatenate the 1D array for re-encoding at the end
+                    full_frames_list.append(
+                        segment_array[plane_index].flatten()
+                    )
 
-            self.SegmentSequence.append(segment_descriptions[i])
+        self.PerFrameFunctionalGroupsSequence = pffg_sequence
+        self.NumberOfFrames = len(pffg_sequence)
 
         if is_encaps:
```
hackermd, line 1538: This part could potentially go into a separate `_add_pixel_data` helper method at some point:

```python
def _add_pixel_data(self, frames: List[numpy.ndarray], is_encapsulated: bool) -> None
```

src/highdicom/seg/sop.py
```diff
@@ -1793,10 +1612,175 @@ def _check_segment_numbers(described_segment_numbers: np.ndarray):
...
+        Parameters
+        ----------
+        source_image: pydicom.Dataset
+            The first source image.
+        is_multiframe: bool
+            Whether the source image is multiframe.
+
+        Returns
+        -------
+        Optional[highdicom.PixelMeasuresSequence]
```
hackermd, line 1633:
```suggestion
        Union[highdicom.PixelMeasuresSequence, None]
```

src/highdicom/seg/sop.py
```diff
@@ -1807,7 +1791,7 @@ def _check_and_cast_pixel_array(
         ----------
         pixel_array: numpy.ndarray
             The segmentation pixel array.
-        described_segment_numbers: numpy.ndarray
+        number_of_segments: int,
```
hackermd, line 1794:
```suggestion
        number_of_segments: int
```

src/highdicom/seg/sop.py
```diff
@@ -1913,48 +1903,305 @@ def _check_and_cast_pixel_array(
...
+                'Encoding an empty segmentation with "omit_empty_frames" '
+                'set to True. Reverting to encoding all frames since omitting '
+                'all frames is not possible.'
+            )
+            return (list(range(pixel_array.shape[0])), True)
+
+        return (source_image_indices, False)
+
+    @staticmethod
+    def _get_segment_array(
```
hackermd, line 1949: For consistency

```suggestion
    def _get_segment_pixel_array(
```

src/highdicom/seg/sop.py
```diff
@@ -1913,48 +1903,305 @@ def _check_and_cast_pixel_array(
...
+        segment_number: int,
+        number_of_segments: int,
+        segmentation_type: SegmentationTypeValues,
+        max_fractional_value: int
+    ) -> np.ndarray:
+        """Get pixel data array for a specific segment.
+
+        This is a helper method used during the constructor. Note that the
+        pixel array is expected to have been processed using the
+        _check_and_cast_pixel_array method before being passed to this method.
```
hackermd, line 1960:
```suggestion
        ``_check_and_cast_pixel_array`` method before being passed to this method.
```

src/highdicom/seg/sop.py
```diff
@@ -1913,48 +1903,305 @@ def _check_and_cast_pixel_array(
...
+        Parameters
+        ----------
         pixel_array: numpy.ndarray
-            Pixel array with empty frames removed
-        plane_positions: List[Optional[highdicom.PlanePositionSequence]]
-            Plane positions with entries corresponding to empty frames removed.
-        source_image_indices: List[int]
-            List giving for each frame in the output pixel array the index of
-            the corresponding frame in the original pixel array
+            Full segmentation array containing all segments.
```
hackermd, line 1965:
```suggestion
            Full segmentation pixel array containing all segments.
```

src/highdicom/seg/sop.py
```diff
@@ -1793,10 +1612,175 @@ def _check_segment_numbers(described_segment_numbers: np.ndarray):
                 f'from 1. Found {described_segment_numbers[0]}. '
             )
 
+    @staticmethod
+    def _get_pixel_measures(
```
hackermd, line 1616: Using the full attribute name may be clearer

```suggestion
    def _get_pixel_measures_sequence(
```

hackermd (approved):
Thanks @CPBridge. LGTM. I left a few minor suggestions for your consideration.

src/highdicom/seg/sop.py
CPBridge: I actually rewrite this piece substantially in an upcoming pull request related to multiprocessing so I'm going to leave as is for now. Agreed that it doesn't make a huge amount of sense at the moment

src/highdicom/seg/sop.py
CPBridge: resolved in [9092a00](https://github.com/ImagingDataCommons/highdicom/pull/227/commits/9092a000b3c1518b3c04c98b07d3b6682fb9a889)

src/highdicom/seg/sop.py
CPBridge: resolved in [9092a00](https://github.com/ImagingDataCommons/highdicom/pull/227/commits/9092a000b3c1518b3c04c98b07d3b6682fb9a889)

PR #229, opened 2023-06-04
Allow duplicate entries in ReferencedSeriesSequence

CPBridge:
I have come across some segmentations where there are duplicate entries in the ReferencedSeriesSequence. Highdicom just crashes with an unhelpful message in this situation. It doesn't really make sense to have duplicate entries here, but it's not explicitly disallowed, so in this PR I deduplicate them and allow them through, and log a warning.


src/highdicom/seg/sop.py
```diff
@@ -2150,7 +2150,17 @@ def _get_ref_instance_uids(self) -> List[Tuple[str, str, str]]:
...
 
-        return instance_data
+        # There shouldn't be duplicates here, but there's no explicit rule
+        # preventing it.
+        # Since dictionary ordering is preserved, this trick deduplicates
+        # the list without changing the order
+        unique_instance_data = list(dict.fromkeys(instance_data))
+        if len(unique_instance_data) != len(instance_data):
+            logger.warning(
+                'Duplicate entries found in the ReferencedSeriesSequence.'
```
hackermd, line 2160: This log message could benefit from a few additional details, such as the SOP Instance UID of the instance and potentially also the duplicate entries. That information may help developers to debug the issue.

hackermd (approved):
Just a minor suggestion.

src/highdicom/seg/sop.py
CPBridge: Addressed in 311d2f0

PR #230, opened 2023-06-09
ANN parsing improvements

CPBridge:
Three small changes to make working with bulk annotations easier:

- Implement an `annread` function like `segread` and `srread` to load file without need to copy (which could be expensive)
- Implement a check on `AnnotationGroup.get_coordinates()` for passing an annotation_index less than one. Previous behaviour for passing 0 silently returned the last (-1) annotation, which is quite confusing
- Add an `annotation_coordinate_type` property to the MicroscopyBulkSimpleAnnotations class


hackermd (approved):
Nice!

Issue #233, opened 2023-06-14
Pydicom 2.4.0 breaks highdicom (TypeError: unhashable type: Dataset)


CPBridge:
The latest update to pydicom breaks several things in highdicom. I will add more detail later, but for now if you are experiencing issues try downgrading to pydicom==2.3.1 and if that fixes things please note the details here

erikogabrielsson:
Hi @CPBridge,

I got an [issue in WsiDicomizer ](https://github.com/imi-bigpicture/wsidicomizer/issues/74) related to this, were doing deepcopy on a SpecimenStaining (and maybe other content classes to) throws. Using 2.3.1 works.

```python
from copy import deepcopy
from highdicom import SpecimenStaining
import pydicom
import highdicom
print("pyidicom vesion: ", pydicom.__version__)
print("highdicom version", highdicom.__version__)
dataset = SpecimenStaining(["hematoxylin stain"])
deepcopy(dataset)
```
```
pyidicom vesion:  2.4.0
highdicom version 0.21.1
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[3], line 8
      6 print("highdicom version", highdicom.__version__)
      7 dataset = SpecimenStaining(["hematoxylin stain"])
----> 8 deepcopy(dataset)

File [c:\Users\er-gac\.pyenv\pyenv-win\versions\3.11.2\Lib\copy.py:153](file:///C:/Users/er-gac/.pyenv/pyenv-win/versions/3.11.2/Lib/copy.py:153), in deepcopy(x, memo, _nil)
    151 copier = getattr(x, "__deepcopy__", None)
    152 if copier is not None:
--> 153     y = copier(memo)
    154 else:
    155     reductor = dispatch_table.get(cls)

File [c:\Users\er-gac\.pyenv\pyenv-win\versions\3.11.2\Lib\site-packages\pydicom\sequence.py:86](file:///C:/Users/er-gac/.pyenv/pyenv-win/versions/3.11.2/Lib/site-packages/pydicom/sequence.py:86), in Sequence.__deepcopy__(self, memo)
     84 if memo is not None:
     85     memo[id(self)] = copied
---> 86 copied.__dict__.update(deepcopy(self.__dict__, memo))
     87 for ds in copied:
     88     ds.parent_seq = copied  # type:ignore

[...]

PR #238, opened 2023-06-19
Exclude incompatible pydicom 2.4.0 in setup.py

CPBridge:
Pydicom 2.4.0 is essentially incompatible with highdicom, see #233 and https://github.com/pydicom/pydicom/issues/1813

The fix introduced in https://github.com/pydicom/pydicom/pulls/1814 and was released in pydicom 2.4.1. We should therefore exclude the specific problematic version of pydicom (2.4.0) in our setup.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ setuptools.setup(
     package_dir={'': 'src'},
     python_requires='>=3.6',
     install_requires=[
-        'pydicom>=2.3.0',
+        'pydicom>=2.3.0,!=2.4.0',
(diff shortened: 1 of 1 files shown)


CPBridge:
Note that the issue is [now fixed](https://github.com/pydicom/pydicom/pull/1814 ) in pydicom and pydicom versions >= 2.4.1 should again be compatible with highdicom. So it's just this one version short-lived (2.4.0) that we need to exclude from highdicom's dependencies.

hackermd (approved):

PR #240, opened 2023-06-21
Convert plane positions to plane position slide sequence

### LINKED ISSUE #231: get_plane_positions_of_image returns list of pydicom.Sequence instead of highdicom.PlanePositionSequence
The function [`get_plane_positions_of_image`](https://github.com/ImagingDataCommons/highdicom/blob/master/src/highdicom/seg/content.py#L459) specifies that it should return a `List[PlanePositionSequence]`. Instead, the call [`item.PlanePositionSequence`](https://github.com/ImagingDataCommons/highdicom/blob/master/src/highdicom/seg/content.py#L459) causes the return of a `List[pydicom.Sequence]`. 

That sequence has the name PlanePositionSequence but it is no instance of the class `highdicom.PlanePositionSequence`, which later can cause the error ["Can only compare instances of PlanePositionSequence"](https://github.com/ImagingDataCommons/highdicom/blob/b1ccd3e46a5fda77859b6d9211ce054212f57da2/src/highdicom/content.py#L470) when comparing plane positions in the [seg/sop.py](https://github.com/ImagingDataCommons/highdicom/blob/b1ccd3e46a5fda77859b6d9211ce054212f57da2/src/highdicom/seg/sop.py#LL1406C33-L1406C33) file. 

I am planning to submit a merge request to fix this by using the function [`highdicom.PlanePositionSequence.from_sequence`](https://github.com/ImagingDataCommons/highdicom/blob/b1ccd3e46a5fda77859b6d9211ce054212f57da2/src/highdicom/content.py#L497) to convert the `pydicom.Sequence` to a `highdicom.PlanePositionSequence`.

thomas-albrecht:
Here is a new branch to fix #231

### DIFF
--- a/src/highdicom/seg/content.py
+++ b/src/highdicom/seg/content.py
@@ -481,8 +481,9 @@ class DimensionIndexSequence(DataElementSequence):
             )
         elif self._coordinate_system == CoordinateSystemNames.SLIDE:
             if hasattr(image, 'PerFrameFunctionalGroupsSequence'):
-                plane_positions = [
+                plane_positions = [PlanePositionSequence.from_sequence(
                     item.PlanePositionSlideSequence
+                )
                     for item in image.PerFrameFunctionalGroupsSequence
                 ]
             else:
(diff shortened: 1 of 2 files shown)


CPBridge (approved):

thomas-albrecht:
Thanks for the tip. I typically always just merge and never rebase. 
But that makes a lot of sense. I would have needed the -f because rebase messes with the history somehow.

hackermd (approved):

Issue #235, opened 2023-06-15
It prompts Failed to read frame #18, when using highdicom to read the specified dicom file.


Marsxpc:
hello, you can reproduce this issue by using the script below:
(environment: windows system and highdicom 0.21.1)

from highdicom.io import ImageFileReader
from pydicom import dcmread, pixel_data_handlers
import numpy as np
import matplotlib.pyplot as plt
path = r'C:\Users\Desktop\XX000090'
frame_idx = 18
with ImageFileReader(path) as image_obj:
    arr = image_obj.read_frame(frame_idx, correct_color=False)
plt.imshow(arr)
plt.show()

error message:
Failed to read frame #18.  File "C:\Users\test.py", line 12, in <module>
    arr = image_obj.read_frame(frame_idx, correct_color=False)
  File "C:\Users\AppData\Local\Programs\Python\Python39\lib\site-packages\highdicom\io.py", line 574, in read_frame
    frame_data = self.read_frame_raw(index)
  File "C:\Users\AppData\Local\Programs\Python\Python39\lib\site-packages\highdicom\io.py", line 545, in read_frame_raw
    raise IOError(f'Failed to read frame #{index}.')
   
 Note: Using pydicom to read this dicom file is OK

Marsxpc:
[XX000090.zip](https://github.com/ImagingDataCommons/highdicom/files/11753333/XX000090.zip)

CPBridge:
Hi @Marsxpc thanks for reporting this issue. I took a quick look and I while I can read the file ok on my linux machine, there is definitely something strange going on: the colours appear to be incorrect and there is some distortion of the frames. I'm not sure whether this is a problem with the library or whether the file is somehow corrupted. I will have to dig in further later.

Can you load the entire file correctly with pydicom (just access the pixel_array)? That would be a useful point of reference to figure out what's going on. Thanks

CPBridge:
Ah sorry, I just noticed your comment about pydicom

Marsxpc:
Hi @CPBridge
First, let me talk about my test results, the following two pictures are the results when I set the frame_idx to 0 and 1
![frame_1](https://github.com/ImagingDataCommons/highdicom/assets/43091577/f201bbf7-4fd8-4d3c-b670-89a0000e384d)
![frame_0](https://github.com/ImagingDataCommons/highdicom/assets/43091577/d584beec-945c-4e16-b9a3-7f8c455a7216)
[...]

Issue #241, opened 2023-07-04
ImageFileReader does not instantiate image object (sometimes)


cheyvaert-medexprim:
Hello, 

I run on python shell the example from the documentation:

```
from pydicom.data import get_testdata_file
from highdicom.io import ImageFileReader

test_filepath = get_testdata_file('eCT_Supplemental.dcm')

with ImageFileReader(test_filepath) as image:
    print(image.metadata.SOPInstanceUID)
    for i in range(image.number_of_frames):
        frame = image.read_frame(i)
        print(frame.shape)
```

It works well.

BUT:

If I tried to do the same with one of my own files:

number_of_frames returns 1, the other method returns empty values, sometimes with error or warning.

BUT:
If called twice the number_of_frames method:

with ImageFileReader(test_filepath) as image:
    image.number_of_frames # calling number_of_frames twice
    image.number_of_frames # calling number_of_frames twice
    print(image.metadata.SOPInstanceUID)
    for i in range(image.number_of_frames):
        frame = image.read_frame(i)
        print(frame.shape)

It's working suddenly well.

The image I tried to open weight 2Go,  and the modality is SM.

CPBridge:
Hi @cheyvaert-medexprim, thanks for using the library and reporting this bug. It will help a lot if you provide an deidentofoed example dataset that demonstrates the issue. I'm afraid I do not understand what you mean in this sentence:

> The image I tried to open weight 2Go

Could you explain? Is this some open data?

CPBridge:
If you can't provide the files, it would be useful if you could confirm a few details:
- what is the SOPClassUID of the file?
- what is the TransferSyntaxUID?
- what are the values of Rows, Columns and NumberOfFrames?

I suspect this may be related to #235

CPBridge:
Could you try the branch on #242 ? (Making sure you have pillow<10.0.0 and >=8.3 installed to avoid #243)

cheyvaert-medexprim:
After a few local test the bug is coming from #243 (it is set by using pillow 2.9.5) thank you. I had the authorization to send you the image today. Just tell me where I can send it to you.

Also:

    what is the SOPClassUID of the file? => 1.2.840.10008.5.1.4.1.1.77.1.6
    what is the PhotometricInterpretation? => YBR_FULL_422
    what is the TransferSyntaxUID? => 1.2.840.10008.1.2.4.50
    what are the values of Rows, Columns and NumberOfFrames? =>  256, 256,  6

[...]

PR #242, opened 2023-07-13
Correctly account for Cr and Cb subsampling for native YBR_FULL_422

CPBridge:
Addresses #235

Occasionally the YBR_FULL_422 photometric interpretation is used for native encoded pixels. When this occurs, the calculation of bytes per frame needs to account for the subsampling of the colour (Cr and Cb) channels. The logic is given [here](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.3.html#sect_C.7.6.3.1.2) as:

> Rows (0028,0010) * Columns (0028,0011) * Number of Frames (0028,0008) * 2

This seems to work correctly with the sample file shared in #235.

### DIFF
--- a/src/highdicom/io.py
+++ b/src/highdicom/io.py
@@ -492,6 +492,12 @@ class ImageFileReader(object):
             #   are packed into 12.5 -> 13 bytes
             return n_pixels // 8 + (n_pixels % 8 > 0)
         else:
+            if self.metadata.PhotometricInterpretation == 'YBR_FULL_422':
+                # Account for subsampling of CB and CR when calculating
+                # expected number of samples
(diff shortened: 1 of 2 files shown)


CPBridge:
Note that issues #235 and #241 seem to be resolved by this change. However, there is more to think through before merging:

- Dealing with the other subsampled photometric interpretations
- Adding tests
- Performing colour space conversion to RGB when colour correction is requested

CPBridge:
Added a test and pushing this out so that it makes it into the next release (0.22.0). I will try to fix the colour space issue soon, but for now users can retrieve the frames without colour correction and then use eg pillow to convert YBR to RGB

CPBridge:
Merging, since tests are failing for reasons unrelated to this PR

Issue #243, opened 2023-07-13
Errors with pillow>=10.0.0: AttributeError: module 'PIL.ImageCms' has no attribute 'INTENT_RELATIVE_COLORIMETRIC'


CPBridge:
Pillow 10.0.0 included some breaking changes to the organisation of constants in the `PIL.ImageCms` module. Values that were previously module-level constants are now enums, e.g. `PIL.ImageCms.INTENT_RELATIVE_COLORIMETRIC` is now `PIL.ImageCms.Intent.RELATIVE_COLORIMETRIC`. The result is that functionality associated with color is broken when using highdicom with Pillow>=10.0.0.

If you are having issues, please downgrade to pillow<=10.0.0

CPBridge:
This was resolved in the latest release 0.22.0

PR #244, opened 2023-07-13
work around pillow 10.0.0 changes

CPBridge:
Fix for #243 :

- Require pillow >=10.0.0
- Update `highdicom.color` to use the new organisation of constants found in pillow==10.0.0

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ setuptools.setup(
     install_requires=[
         'pydicom>=2.3.0,!=2.4.0',
         'numpy>=1.19',
-        'pillow>=8.3',
+        'pillow>=10.0.0',
         'pillow-jpls>=1.0',
     ],
     extras_require={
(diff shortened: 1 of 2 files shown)


CPBridge:
ok so it appears that pillow 10.0.0 does not support python 3.7. This is going to be mmore complex that I had thought. We will have to either drop support for 3.7 or put some hacks in place that can deal with both pillow<10.0.0 and pillow>=10.0.0. I am reluctant to drop support for 3.7 personally, as it is still in widespread use

CPBridge:
OK so this is ugly but I added logic that works around the breaking change at pillow 10.0.0 so that we can continue to support python 3.7. While this doesn't smell great, in my opinion, it is worth it to avoid dropping an entire version of python just to avoid this tiny little API change. I note that over the last 6 months around 5-10% of our users are using 3.7: https://pypistats.org/packages/highdicom. What do you think @hackermd ?

hackermd (approved):

PR #245, opened 2023-07-27
Multiprocessing for frame encoding in Segmentation constructor

CPBridge:
Add the ability to use concurrency to encode frames when creating segmentations.

When using encapsulated transfer syntaxes, the process of encoding frames takes a significant amount of the total time for creation of a Segmentation object. For large multiframe segmentations such as those of WSIs, this can take several minutes. Parallelising the encoding of frames is an obvious place for significant performance gains.

In this PR, I add the ability to use multiprocessing for frame encoding using the Python standard library's `concurrent.futures` module. The user specifies a `workers` parameter governing the number of worker processes to spawn (with the default remaining no worker threads). Alternatively they may "bring their own" worker pool by providing an instance of any `concurrent.futures.Executor`. This has two potential benefits. First it allows a worker pool to be re-used in programs where a large number of segmentations needs to be created to save to overhead of setting up the worker pools. Secondly, it gives the user more control over the multiprocessing context if they want it. For example, in principle, they could pass a ThreadPoolExecutor to use multi-threading rather that multiprocessing.

I also made a few other tweaks to the segmentation array processing for further efficiency gains. Most notably, I adjusted the `_get_segment_pixel_array` method to return the array for a single frame and a single segment (as opposed to a all frames and a single segment). This makes the looping logic a bit simpler and it turns out that this is slightly faster, presumably because it avoid the need to allocate memory for large arrays.

### DIFF
--- a/src/highdicom/seg/sop.py
+++ b/src/highdicom/seg/sop.py
@@ -2010,7 +2010,7 @@ class Segmentation(SOPClass):
         segmentation_type: SegmentationTypeValues,
         max_fractional_value: int
     ) -> np.ndarray:
-        """Get pixel data array for a specific segment.
+        """Get pixel data array for a specific segment and plane.
 
         This is a helper method used during the constructor. Note that the
         pixel array is expected to have been processed using the
@@ -2020,9 +2020,9 @@ class Segmentation(SOPClass):
         Parameters
         ----------
         pixel_array: numpy.ndarray
-            Full segmentation pixel array containing all segments for a single
-            plane. Array is therefore either (Rows x Columns x Segments) or
-            (Rows x Columns) in case of a "label map" style array.
+            Segmentation pixel array containing all segments for a single plane.
+            Array is therefore either (Rows x Columns x Segments) or (Rows x
+            Columns) in case of a "label map" style array.
         segment_number: int
             The segment of interest.
         number_of_segments: int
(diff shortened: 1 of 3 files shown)


src/highdicom/seg/sop.py
```diff
@@ -1523,29 +1591,49 @@ def __init__(
...
+                            **encode_frame_kwargs,
+                        )
+                        frame_futures_list.append(future)
                 else:
-                    # Concatenate the 1D array for re-encoding at the end
-                    full_frames_list.append(
-                        segment_array[plane_index].flatten()
-                    )
+                    # Concatenate the 1D array for encoding at the end
+                    full_frames_list.append(segment_array.flatten())
```
hackermd, line 1614: I wonder whether a separate variable is needed or whether frames (compressed or not) could just be stored in a list called `frames`? How the list of frames then needs to be handled (encapsulation or native encoding) is determined by `is_enaps`.

The variable for the futures could be called `frames_futures`.

In my opinion, that would make it easier to follow.

hackermd (commented):
Cool! I really like that parallel encoding of frames using multiple workers. It makes me wonder whether we should use multiple workers for decoding of frames as well.

src/highdicom/seg/sop.py
CPBridge: I did it this way primarily to appease mypy, but I guess it probably hurts readability. Implemented 5c5cbb6 as you suggest

CPBridge:
Yes I have definitely thought about decoding with multiple workers too, one for the to-do list.

PR #260, opened 2023-10-09
Remove redundant cast

DimitriPapadopoulos:
Casting a variable to its type hint seems redundant, although it can be seen as defensive programming.


CPBridge:
This is done for the purpose of static analysis tools such as mypy. We used to have `mypy` as part of our CI/CD checks until changes in pydicom created so many issues that it became untenable. At some point though we hope to return to including these checks so I'm going to keep this here.

PR #261, opened 2023-10-09
Creating a list from a list is equivalent to a copy

DimitriPapadopoulos:
In that case, I find calling `.copy()` is more readable.

However, do we really **need** this copy? The list `unique_instance_data` is initialised in this function and then returned. What's the point of copying it?

### DIFF
--- a/src/highdicom/seg/sop.py
+++ b/src/highdicom/seg/sop.py
@@ -2174,7 +2174,7 @@ class Segmentation(SOPClass):
                 f'duplicated referenced SOP Instance UID items: {display_str}.'
             )
 
-        return list(unique_instance_data)
+        return unique_instance_data.copy()
 
     def _build_luts(self) -> None:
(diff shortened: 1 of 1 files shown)


src/highdicom/seg/sop.py
```diff
@@ -2174,7 +2174,7 @@ def _get_ref_instance_uids(self) -> List[Tuple[str, str, str]]:
                 f'duplicated referenced SOP Instance UID items: {display_str}.'
             )
 
-        return list(unique_instance_data)
+        return unique_instance_data.copy()
```
CPBridge, line 2177: I agree, I think this `list()` call is vestigial, let's remove outright
```suggestion
        return unique_instance_data
```

CPBridge (changes requested):

CPBridge (approved):

PR #262, opened 2023-10-09
Use list comprehension where possible

DimitriPapadopoulos:


src/highdicom/seg/sop.py
```diff
@@ -2595,12 +2595,9 @@ def segmented_property_categories(self) -> List[CodedConcept]:
...
         """
-        categories = []
-        for desc in self.SegmentSequence:
-            if desc.segmented_property_category not in categories:
-                categories.append(desc.segmented_property_category)
-
-        return categories
+        return list(
+            {desc.segmented_property_category for desc in self.SegmentSequence}
+        )
```
CPBridge, line 2600: As a set comprehension this may change order, which is undesirable and may additionally break existing user code. Let's revert this to how it was, I don't think this can be achieved with a comprehension

src/highdicom/seg/sop.py
```diff
@@ -2613,12 +2610,9 @@ def segmented_property_types(self) -> List[CodedConcept]:
...
         """
-        types = []
-        for desc in self.SegmentSequence:
-            if desc.segmented_property_type not in types:
-                types.append(desc.segmented_property_type)
-
-        return types
+        return list(
+            {desc.segmented_property_type for desc in self.SegmentSequence}
+        )
```
CPBridge, line 2615: As above, this changes order. We should avoid this

CPBridge (changes requested):
2 of these 3 changes could break user code and should be reverted

src/highdicom/seg/sop.py
DimitriPapadopoulos: I understand. I was under the impression that order is not important here (in which case a set could be returned outright) but that's not documented.

DimitriPapadopoulos: Fixed.

CPBridge: Thanks. You are right that it is not documented and so users should not be relying on a particular order. Still, it is probably what many will expect and I think it's safer not to change behaviour now for the sake of code neatness

CPBridge (approved):

PR #265, opened 2023-10-09
`io.open()` → `open()`

DimitriPapadopoulos:
From https://docs.python.org/3/library/io.html#io.open:
> This is an alias for the builtin [`open()`](https://docs.python.org/3/library/functions.html#open) function.


CPBridge (approved):
Thanks

PR #267, opened 2023-10-11
Add Python 3.12 CI tests

DimitriPapadopoulos:


CPBridge (approved):

CPBridge (commented):
We will have to wait until jpegls is working on 3.12 before we can add this I think

DimitriPapadopoulos:
I understand this depends on [pillow-jpls](https://github.com/planetmarshall/pillow-jpls), doesn't it?

I have opened issue https://github.com/planetmarshall/pillow-jpls/issues/20.

erikogabrielsson:
Is there a reason for pillow-jpls not being an extra like pylibjpeg et al?

DimitriPapadopoulos:
I think this is because some kind of JPEG library is needed. From the [Installation guide](https://github.com/ImagingDataCommons/highdicom/blob/master/docs/installation.rst):
> The library relies on the underlying `pydicom` package for decoding of pixel data, which internally delegates the task to either the `pillow` or the `pylibjpeg` packages. Since `pillow` is a dependency of _highdicom_ and will automatically be installed, some transfer syntax can thus be readily decoded and encoded (baseline JPEG, JPEG-2000, JPEG-LS). Support for additional transfer syntaxes (e.g., lossless JPEG) requires installation of the `pylibjpeg` package as well as the `pylibjpeg-libjpeg` and `pylibjpeg-openjpeg` packages. Since `pylibjpeg-libjpeg` is licensed under a copyleft GPL v3 license, it is not installed by default when you install _highdicom_.

CPBridge:
That's correct. The main reason that pylibjpeg is optional is licence considerations. Since there are no such issues with `pillow-jpls`, we didn't make it optional. However if things like this holdup with 3.12 occur, it may make sense to make it optional too.

Generally we have been a bit wary of this approach of having a confusing mess of optional dependencies

erikogabrielsson:
Thanks for the clarification @CPBridge. I agree that optional dependencies can be messy and that you should consider making it optional if it looks like its no longer maintained.

DimitriPapadopoulos:
So the idea would be to add a `try:` before `import pillow_jpls` and emit an error message suggesting to install the module if `except ImportError`?
https://github.com/ImagingDataCommons/highdicom/blob/0f43f8e805bfc9be5256976f8eb984c7e45c3c5e/src/highdicom/frame.py#L260-L261

erikogabrielsson:
I (too) made a [PR to pillow-jpls](https://github.com/planetmarshall/pillow-jpls/pull/24) for Python 3.12 and Pillow 10 support.

Regarding reading mpressed transfer syntaxes other than standard jpeg and jpeg 2000, [wsidicom](https://github.com/imi-bigpicture/wsidicom/blob/main/wsidicom/codec/decoder.py) uses [imagecodecs](https://github.com/cgohlke/imagecodecs) for decoding and encoding jpeg 12 bit, jpeg lossless, and jpeg ls.

CPBridge:
Thanks @erikogabrielsson . I had sort of hoped that pillow-jpls would come back to life and save us from having to work around this, but that is looking increasingly less likely. I will look into whether `imagecodecs` could be a good alternative.

erikogabrielsson:
From what I understand, highdicom only uses pillow-jpls to enable encoding? Decoding can only be done if having the `libjpeg` extra with `pylibjpeg-libjpeg`.
I did a quick test with:
```python
return jpegls_encode(array)
```
in the [jpeg ls statement in `frame.py`](https://github.com/erikogabrielsson/highdicom/blob/imagecodecs/src/highdicom/frame.py) and it passed the tests.

erikogabrielsson:
> Thanks @erikogabrielsson . I had sort of hoped that pillow-jpls would come back to life and save us from having to work around this, but that is looking increasingly less likely. I will look into whether `imagecodecs` could be a good alternative.

Looks like the issue is now [addressed](https://github.com/planetmarshall/pillow-jpls/pull/25). Hopefulle we will se a build for 3.12 soon.

CPBridge:
Merging this now that the dependency issue is fixed

PR #268, opened 2023-10-11
http:// → https://

DimitriPapadopoulos:
I have updated only `http://` URLs that are actually redirected to an `https://` target.


CPBridge (approved):
Thanks

Issue #275, opened 2023-12-17
Issues with jpeg 2000 and  `encode_frame`


erikogabrielsson:
Looking at the `encode_frame`-method i can see some issues:
- [In DICOM Jpeg 2000 encoded frames should not be in JP2 format](https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_A.4.4.html), which is the default when saving using Pilow. `"no_jp2": True` should be added to the settings.
- If I understand the parameter description correctly, `photometric_interpretation` is the intended color format in the output frame. The current implementation only allows `YBR_FULL` for color data, and the allowed values are [`YBR_RCT`, `YBR_ICT`, `YBR_FULL` and `RGB`](https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_8.2.4.html). According to the standard `RGB` and `YBR_FULL` should not have any component transform, while `YBR_RCT` and `YBR_ICT`should. To get Pillow to do the component transform to `YBR`, you need to add  `"mct": True` to the settings.

The parameters for saving Jpeg 2000 with Pillow is described [here](https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#jpeg-2000).

Not sure about the intended use of the `encode_frame`-method, but as it is used in `highdicom` to produce `PixelData` elements at least the jp2-issue should be fixed.

erikogabrielsson:
Hi @CPBridge,
No problem with the delay.

1. It looks good'.
2. From what I remember, my issue was "maybe also YBR_RCT and YBR_ICT should be supported" :)

CPBridge:
The first point has now been resolved with #280, which will be inthe next release.

We do not have plans at this time to support the other photometric interpretations, since they are not widely used at the moment, so I am closing this issue. If people can point to a use case for those values, then it may be worth implementing them, but I do not know of any

PR #280, opened 2024-03-27
Remove JP2 wrapper from JPEG 2000 encoding

CPBridge:
Fix for #275

@erikogabrielsson

### DIFF
--- a/src/highdicom/frame.py
+++ b/src/highdicom/frame.py
@@ -50,7 +50,7 @@ def encode_frame(
         Number of bits that need to be allocated per pixel sample
     bits_stored: int
         Number of bits that are required to store a pixel sample
-    photometric_interpretation: int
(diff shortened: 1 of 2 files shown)


CPBridge:
Tagging @hackermd for some help here.

hackermd:
I think @erikogabrielsson is right and the content should be encoded as J2K (raw codestream) rather JP2 (file format). The `no_jp2` flag appears to achieve that goal.

hackermd:
@CPBridge I think we need to update the unit test and check for the magic number of the JPEG2000 raw codestream (`b"\xFF\x4F\xFF\x51"`) instead:

https://github.com/ImagingDataCommons/highdicom/blob/v0.22.0/tests/test_frame.py#L174

CPBridge:
Thanks both, I will put out a minor release relatively soon with this fix

PR #287, opened 2024-05-28
Enforce ruff/pyupgrade rules (UP)

DimitriPapadopoulos:
This is a suggestion of Scientific Python [Repo-Review](https://learn.scientific-python.org/development/guides/repo-review/?repo=ImagingDataCommons%2Fhighdicom&branch=master).

### DIFF
--- a/bin/create_iods_modules.py
+++ b/bin/create_iods_modules.py
@@ -105,9 +105,9 @@ if __name__ == '__main__':
     except IndexError:
         raise ValueError('Path to directory must be provided.')
     if not os.path.exists(directory):
-        raise OSError('Path does not exist: "{}"'.format(directory))
+        raise OSError(f'Path does not exist: "{directory}"')
     if not os.path.isdir(directory):
-        raise OSError('Path is not a directory: "{}"'.format(directory))
+        raise OSError(f'Path is not a directory: "{directory}"')
 
     now = datetime.datetime.now()
     current_date = datetime.datetime.date(now).strftime('%Y-%m-%d')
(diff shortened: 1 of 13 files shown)


CPBridge:
Hi @DimitriPapadopoulos 

Thanks for the PR. We currently use flake8 as a linter, which overall does a pretty good job. However these other tools appear to catch some other potentially useful things so I'm happy to make these changes.

However there is one issue in that our flake8 settings require <=80 character lines (@hackermd 's has some strong feelings about this...), and one of the tools (ruff?) appears to be reformatting lines to be longer than this limit, which is causing our current CI pipeline to fail. I'm not keen to change the line length limit as it has been in place since the start and we would have inconsistently formatted code everywhere. Is it possible to change the settings to prevent reformatting beyond 80 chars?

Also, would it make sense to add these tools to the CI so that we continue to remain in compliance?

DimitriPapadopoulos:
I agree flake8 does a good job, changing might not a priority. For what it's worth, the Scientific Python Development Guide mostly suggests:
1. moving from `setup.py`/`setup.cfg` to `pyproject.toml`, which is probably the most important item,
2. using _ruff_ for linting and formatting, with a minimal set of rules including _bugbear_ and _pyupgrade_,
3. using _pre-commit_ - I am not a fan of _pre-commit_ but eventually got used to it.

I'm not sure these suggestions have anything to do with _compliance_. I expect they will remain mere suggestions in the foreseeable future. But then I must admit I have become used to code formatted with _ruff_, and find it's a good idea to follow the trend.

To answer your question, yes, we can set _ruff_ to use 80 chars/line. The default for _ruff_ is 88 and a default maximum of 88 to 100 chars/line seems more reasonable nowadays. But again, we can set it to 80.

You may use only the _ruff_ linter. If you do use the _ruff_ formatter, it will probably change some code any way - so that would probably be a good moment to switch to 88 chars/line or more.

While the bugbear rules do detect actual issues (like a missing `assert` in tests), the pyupgrade rules mostly renovate the code and may improve performance - although I suspect such "random" performance improvements are usually not perceptible at the scale of a whole Python module.

DimitriPapadopoulos:
I'll look into fixing line length next week.

CPBridge (approved):

PR #290, opened 2024-06-02
Enforce ruff rules (RUF)

DimitriPapadopoulos:


CPBridge (approved):
Some good catches here

PR #292, opened 2024-06-02
Enforce ruff/flake8-pyi rules (PYI)

DimitriPapadopoulos:


CPBridge (approved):
@DimitriPapadopoulos 

Thanks for these various changes. I'm wondering whether you would be able to share the specific configurarion/command line options you used for these various tools so that I can look into integrating them in the CI?

DimitriPapadopoulos:
I used `ruff --extend--select PYI` for this specific merge request. Ideally I should have used `ruff --config 'line-length = 80' --extend--select PYI` instead of fixing possible line length issues after linting. Since you're using _flake8_, you just need to add to your test requirements the relevant plugin — in this case [flake8-pyi](https://github.com/PyCQA/flake8-pyi).

DimitriPapadopoulos:
More generally, for _ruff_ rules that implement the rules of _flake8_ plugins, just install the plugin:
* B → [flake8-bugbear](https://github.com/PyCQA/flake8-bugbear)
* PYI -> [flake8-pyi](https://github.com/PyCQA/flake8-pyi)

For _ruff_ rules that implement the rules of other tools, add the relevant tool to your test dependencies and add them to the CI jobs:
* UP → [pyupgrade](https://github.com/asottile/pyupgrade)
* FLY → [flynt](https://github.com/ikamensh/flynt)

DimitriPapadopoulos:
If you do want to enforce such rules, I would recommend moving to _ruff_ altogether.

Issue #288, opened 2024-06-02
Move from `setup.py` to `pyproject.toml`?


DimitriPapadopoulos:
Would it make sense to move `setup.py` and `requirements*.txt` to `pyproject.toml`?

CPBridge:
Completed in #293

PR #293, opened 2024-06-16
setup.py → pyproject.toml

DimitriPapadopoulos:
Fxies #288.


CPBridge:
Thanks, looks good but I'll run a few careful tests before we merge

CPBridge (changes requested):
Thanks for taking the time to do this!

A couple of things:
1. It appears that the license (and less importantly the platforms) did not make it across from `setup.py` to `pyproject.toml`.
2. We should remove the `requirements_doc.txt` and `requirements_test.txt`  as they are now redundant, right? Or is there some reason you left them in place?

DimitriPapadopoulos:
1. The license is part of the [classifiers](https://github.com/DimitriPapadopoulos/highdicom/blob/pyproject.toml/pyproject.toml#L20):
   ```
       "License :: OSI Approved :: MIT License",
   ```
   I have not duplicated it in a `license` field because the [Python Packaging User Guide](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license) reads:
   > If you are using a standard, well-known license, it is not necessary to use this field. Instead, you should use one of the [classifiers](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#classifiers) starting with `License ::`.
   
   Same for platforms, they are documented in the classifiers. There's no key dedicated to platforms anyway.
2. I've removed `requirements_doc.txt` and `requirements_test.txt`.

CPBridge:
Thanks! That makes sense.

But now the github actions are failing because they depend on the old `requirements_test.txt`. Could you please make the simple change to [this line](https://github.com/ImagingDataCommons/highdicom/blob/b7f81f6e2aaa525342f058913f1cf49ade8b6457/.github/workflows/run_unit_tests.yml#L30) of `run_unit_tests.yml` to fix and get the CI pipeline running again?

DimitriPapadopoulos:
The CI jobs pass, I have also updated the documentation.

CPBridge:
> I have also updated the documentation.

Ah yes, good catch. Thanks for your help!

CPBridge (approved):

DimitriPapadopoulos:
Unfortunately, I missed an occurrence of `requirements_docs.txt`:
https://github.com/ImagingDataCommons/highdicom/blob/3f7a507958a49f7a2438f492a6a3a7a70eb3a45d/.readthedocs.yaml#L27

CPBridge:
Yeah I just noticed that too, I'll take care of it. Thanks!

DimitriPapadopoulos:
Perhaps you need to restore `requirements_docs.txt`. It looks like `pyproject.toml` won't be supported by readthedocs:
https://github.com/readthedocs/readthedocs.org/issues/8599

DimitriPapadopoulos:
My wrong, this is covered by [packages](https://docs.readthedocs.io/en/stable/config-file/v2.html#packages):
> ```yaml
> version: 2
> 
> python:
>   install:
>     - method: pip
>       path: .
>       extra_requirements:
>         - docs
> ```
> With the previous settings, Read the Docs will execute the next commands:
> ```console
> pip install .[docs]
> ```

CPBridge:
Yup, I came across and implemented the same solution. Builds just passed on readthedocs

PR #307, opened 2024-09-26
Add coordinate for Measurement in SR

### LINKED ISSUE #306: Add SCOORD to a Measurement in SR (TID 320)
Hi, is it possible to add spatial coordinates to a Measurement in SR?

I'm looking at [TID 320 Image or Spatial Coordinates](https://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_320), specifically row 3 and 4.

If it's not currently possible, I could work on a PR to add it.

But I could use some pointers on how to best implement it.
As I understand, the `referenced_images` parameter in `sr.Measurement` implements the first row of TID 320 (inferred from image).
What's the best solution to add the other options? Should a new parameter in `sr.Measurement` be added for each one? Or the `referenced_images` type could be changed to `Optional[Sequence[SourceImageForMeasurement | SourceScoordForMeasurement]]`?

Fedalto:
This adds support for row 3 and 4 from [TID 320](https://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_320).

Resolves #306

### DIFF
--- a/src/highdicom/sr/content.py
+++ b/src/highdicom/sr/content.py
@@ -730,6 +730,36 @@ class CoordinatesForMeasurement(ScoordContentItem):
         )
         self.ContentSequence = [source_image]
 
+    @classmethod
+    def from_dataset(
+            cls,
+            dataset: Dataset,
+            copy: bool = True,
+    ) -> 'CoordinatesForMeasurement':
+        """Construct object from an existing dataset.
+
+        Parameters
+        ----------
+        dataset: pydicom.dataset.Dataset
+            Dataset representing an SR Content Item with value type SCOORD
+        copy: bool
+            If True, the underlying dataset is deep-copied such that the
+            original dataset remains intact. If False, this operation will
+            alter the original dataset in place.
+
+        Returns
+        -------
+        highdicom.sr.CoordinatesForMeasurement
+            Constructed object
+
+        """
+        if copy:
+            dataset_copy = deepcopy(dataset)
+        else:
+            dataset_copy = dataset
+        item = super()._from_dataset_base(dataset_copy)
+        return cast(CoordinatesForMeasurement, item)
+
 
 class ImageRegion(ScoordContentItem):
 
--- a/src/highdicom/sr/content.py
+++ b/src/highdicom/sr/content.py
@@ -443,11 +443,7 @@ class SourceImageForRegion(ImageContentItem):
                     '1-based.'
                 )
         super().__init__(
-            name=CodedConcept(
-                value='111040',
-                meaning='Original Source',
-                scheme_designator='DCM'
-            ),
+            name=codes.SCT.Source,
             referenced_sop_class_uid=referenced_sop_class_uid,
             referenced_sop_instance_uid=referenced_sop_instance_uid,
             referenced_frame_numbers=referenced_frame_numbers,
--- a/src/highdicom/sr/content.py
+++ b/src/highdicom/sr/content.py
@@ -707,6 +707,30 @@ class SourceSeriesForSegmentation(UIDRefContentItem):
         return cast(SourceSeriesForSegmentation, item)
 
 
+class CoordinatesForMeasurement(ScoordContentItem):
+
+    """Content item representing spatial coordinates of a measurement"""
+
+    def __init__(
+            self,
+            graphic_type: Union[GraphicTypeValues, str],
+            graphic_data: np.ndarray,
+            source_image: SourceImageForRegion,
+    ) -> None:
+        graphic_type = GraphicTypeValues(graphic_type)
+        super().__init__(
+            name=CodedConcept(
+                value='121112',
+                meaning='Source of Measurement',
+                scheme_designator='DCM'
+            ),
+            graphic_type=graphic_type,
+            graphic_data=graphic_data,
+            relationship_type=RelationshipTypeValues.INFERRED_FROM,
+        )
+        self.ContentSequence = [source_image]
+
+
 class ImageRegion(ScoordContentItem):
 
     """Content item representing an image region of interest in the
--- a/src/highdicom/sr/content.py
+++ b/src/highdicom/sr/content.py
@@ -716,14 +716,11 @@ class CoordinatesForMeasurement(ScoordContentItem):
             graphic_type: Union[GraphicTypeValues, str],
             graphic_data: np.ndarray,
             source_image: SourceImageForRegion,
+            purpose: Union[CodedConcept, Code] = codes.SCT.Source,
     ) -> None:
         graphic_type = GraphicTypeValues(graphic_type)
         super().__init__(
-            name=CodedConcept(
-                value='121112',
-                meaning='Source of Measurement',
-                scheme_designator='DCM'
-            ),
+            name=purpose,
             graphic_type=graphic_type,
             graphic_data=graphic_data,
             relationship_type=RelationshipTypeValues.INFERRED_FROM,
--- a/src/highdicom/sr/content.py
+++ b/src/highdicom/sr/content.py
@@ -764,7 +764,7 @@ class CoordinatesForMeasurement(ScoordContentItem):
         else:
             dataset_copy = dataset
         item = super()._from_dataset_base(dataset_copy)
-        return cast(CoordinatesForMeasurement, item)
+        return cast(cls, item)
 
 
 class CoordinatesForMeasurement3D(Scoord3DContentItem):
--- a/src/highdicom/sr/content.py
+++ b/src/highdicom/sr/content.py
@@ -714,6 +714,19 @@ class CoordinatesForMeasurement(ScoordContentItem):
             source_image: SourceImageForRegion,
             purpose: Union[CodedConcept, Code] = codes.SCT.Source,
     ) -> None:
+        """
+        Parameters
+        ----------
+        purpose: Union[highdicom.sr.CodedConcept, pydicom.sr.coding.Code]
+            Concept name
+        graphic_type: Union[highdicom.sr.GraphicTypeValues, str]
+            Name of the graphic type
+        graphic_data: numpy.ndarray
+            Array of ordered spatial coordinates, where each row of the array
+            represents a (Column,Row) pair
+        source_image: highdicom.sr.SourceImageForRegion
+            Source image to which `graphic_data` relates
+        """
         graphic_type = GraphicTypeValues(graphic_type)
         super().__init__(
             name=purpose,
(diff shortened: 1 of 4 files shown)


src/highdicom/sr/content.py
```diff
@@ -707,6 +707,30 @@ def from_dataset(
...
 
+class CoordinatesForMeasurement(ScoordContentItem):
+
+    """Content item representing spatial coordinates of a measurement"""
+
+    def __init__(
+            self,
+            graphic_type: Union[GraphicTypeValues, str],
+            graphic_data: np.ndarray,
+            source_image: SourceImageForRegion,
```
Fedalto, line 718: This has nothing to do with ImageRegion, but I reused `SourceImageForRegion` here as it needs to be an image with `SELECTED FROM` relationship. Not sure if there's a better option.

src/highdicom/sr/templates.py
```diff
@@ -2386,7 +2386,7 @@ def __init__(
         finding_sites: Optional[Sequence[FindingSite]] = None,
         method: Optional[Union[CodedConcept, Code]] = None,
         properties: Optional[MeasurementProperties] = None,
-        referenced_images: Optional[Sequence[SourceImageForMeasurement]] = None,
+        referenced_images: Optional[Sequence[Union[SourceImageForMeasurement, CoordinatesForMeasurement]]] = None,
```
Fedalto, line 2389: I created another option and used the same `referenced_images` parameter.

Another solution I see would be to expand `SourceImageForMeasurement` and implement all options from TID 320 in it.

If a `SCOORD` is given, use row 3 and 4. If `SCOORD3D`, then make it use row 6 (I have not implemented this). If no coordinate is passed, use row 1 (which is the current SourceImageForMeasurement functionality)

CPBridge (changes requested):
Hi @Fedalto, thanks for the nice PR!

Generally speaking, it's rather unfortunate that the way this was originally implemented (with only row 1 of TID 320 being used) makes it a little harder to implement this in a backwards-compatible manner.

A couple of other requests, if you are willing:
- Ideally we would add a `referenced_coordinates` property to the `Measurement` class to retrieve the coordinates from the object? This would be very similar to how the [`referenced_images`](https://github.com/ImagingDataCommons/highdicom/blob/cb62737d13ea1bda4625ca0862ec466ecce2a917/src/highdicom/sr/templates.py#L2612) property works now (though, given some of other comments, I suppose it would be best not to specify a concept name, as in fact we probably should not have done for `referenced_images`).
- Unfortunately the `Measurement` class not only implements TID 300 as it is used within TID 1501 but also rows 5 and after of [TID 1491](https://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1419) (which in turn is used within TID 1410 and TID 1411). This was to simplify code and reduce duplication, but unfortunately the two are not exactly the same, and these coordinates are not included in TID 1419.  I think this is why this is why these options were not included in the first place. Though technically the template is extensible, I think it would be best for the [`_ROIMeasurementsAndQualitativeEvaluations`](https://github.com/ImagingDataCommons/highdicom/blob/cb62737d13ea1bda4625ca0862ec466ecce2a917/src/highdicom/sr/templates.py#L3136) class to check any measurements it is passed do not contain `SCOORD` or `SCOORD3D` content items, since those templates are supposed to work rather differently (with the region defined as part of the TID 1410 and TID 1411 templates).

Note that I also target your PR at a new `v0.24.0` branch rather than `master`.

src/highdicom/sr/templates.py
Fedalto: > Then you would need to document and enforce that using referenced_images and referenced_coordinates are mutually exclusive.

I don't think this is the case.
See row 13 of TID 300. Measurement, it has `VM 1-n`, so I understand that `referenced_images` and `referenced_coordinates` can both be used inside a single `Measurement`.

CPBridge: Hmm, yes I guess you are right. They are mutually exclusive within TID 320, but there's no requirement that multiple uses of TID320 within TID300 have the same type

CPBridge:
@Fedalto is this ready for another review?

Fedalto:
Most of it. 

I just realized that TID 1411 also does not use TID 300, which I thought it did. So the check I did in 1410 (Planar ROI Measurements and Qualitative Evaluations) needs indeed to be in `_ROIMeasurementsAndQualitativeEvaluations` as you had suggested.
I'll make this change today and will let you know

Fedalto:
Hi @CPBridge. This should be ready for another review.
Thanks.

src/highdicom/sr/content.py
```diff
@@ -707,6 +703,57 @@ def from_dataset(
...
+            purpose: Union[CodedConcept, Code] = codes.SCT.Source,
+    ) -> None:
+        graphic_type = GraphicTypeValues(graphic_type)
```
CPBridge, line 717: Could you add a docstring for the `__init__` method?

src/highdicom/sr/content.py
```diff
@@ -707,6 +703,57 @@ def from_dataset(
...
+        highdicom.sr.CoordinatesForMeasurement
+            Constructed object
+
+        """
+        if copy:
+            dataset_copy = deepcopy(dataset)
+        else:
+            dataset_copy = dataset
+        item = super()._from_dataset_base(dataset_copy)
+        return cast(CoordinatesForMeasurement, item)
```
CPBridge, line 754: Going forward I'm doing this for better inheritance behaviour. If I remember correctly, a change in the `0.24.0` dev branch makes this change for all such classes.

```suggestion
        return cast(cls, item)
```

There's also the matter of then how to correctly type the return type in this case. But don't worry I'll make that change before release.

tests/test_sr.py
```diff
@@ -3068,6 +3093,24 @@ def test_constructed_with_multiple_references(self):
...
+                graphic_data=np.array([[1, 1], [2, 2]]),
+                source_image=self._image_for_region
+            )]
+        )
+        with pytest.raises(ValueError) as exc:
+            PlanarROIMeasurementsAndQualitativeEvaluations(
+                tracking_identifier=self._tracking_identifier,
+                referenced_region=self._region,
+                measurements=[measurement],
+            )
```
CPBridge, line 3112: Can we add a test somewhere that the new `referenced_coordinates` property works as expected, and returns the right number of items and correct type?

CPBridge (changes requested):
Hi @Fedalto, thanks for your work on this. There are a few comments above.

Also, if we are adding referenced coordinates I would prefer to add the `SCOORD3D` case at the same time (though I'm not concerned about the "R-INFERRED" variants, which honestly confuse me). Would you be happy to add this?

I think it would be pretty straightforward since it does not need the nested source image content item. We could just pass a plain `Scoord3DContentItem` into the `Measurement` init method, i.e. make `referenced_coordinates` of type `Optional[Sequence[Union[CoordinatesForMeasurement, Scoord3DContentItem]]]`, and also have the corresponding property search for and return items of value type `SCOORD` or `SCOORD3D`.

But since this would force the user to pick a concept name/purpose and I think having sane defaults is extremely important in this highly complex SR stuff, I would actually lean towards defining a new `CoordinatesForMeasurement3D` class, which is basically just a `Scoord3DContentItem` derived class with a little default value for the purpose (just like the existing `CoordinatesForMeasurement` class. I'm also happy to work on this if you prefer.

src/highdicom/sr/templates.py
```diff
@@ -2431,6 +2433,8 @@ def __init__(
             and an indication of its selection from a set of measurements
         referenced_images: Union[Sequence[highdicom.sr.SourceImageForMeasurement], None], optional
             Referenced images which were used as sources for the measurement
+        referenced_coordinates: Union[Sequence[highdicom.sr.CoordinatesForMeasurement], None], optional
```
CPBridge, line 2436: Oh one more thing, can we add a quick note to say that this is not valid for measurements that will later be included in Planar/Volumetric ROI Measurement groups?

Fedalto:
Hi @CPBridge. 
I made the requested changes. I believe this is ready for another review.
Thanks

src/highdicom/sr/templates.py
```diff
@@ -10,6 +10,7 @@
 
 from highdicom.sr.coding import CodedConcept
 from highdicom.sr.content import (
+    CoordinatesForMeasurement,
```
CPBridge, line 13:
```suggestion
    CoordinatesForMeasurement,
    CoordinatesForMeasurement3D,
```

src/highdicom/sr/templates.py
```diff
@@ -2387,6 +2389,7 @@ def __init__(
         method: Optional[Union[CodedConcept, Code]] = None,
         properties: Optional[MeasurementProperties] = None,
         referenced_images: Optional[Sequence[SourceImageForMeasurement]] = None,
+        referenced_coordinates: Optional[Sequence[Union[CoordinatesForMeasurement, Scoord3DContentItem]]] = None,
```
CPBridge, line 2392:
```suggestion
        referenced_coordinates: Optional[Sequence[Union[CoordinatesForMeasurement, CoordinatesForMeasurement3D]]] = None,
```

src/highdicom/sr/templates.py
```diff
@@ -2431,6 +2434,11 @@ def __init__(
             and an indication of its selection from a set of measurements
         referenced_images: Union[Sequence[highdicom.sr.SourceImageForMeasurement], None], optional
             Referenced images which were used as sources for the measurement
+        referenced_coordinates: Union[Sequence[Union[highdicom.sr.CoordinatesForMeasurement, Scoord3DContentItem]], None], optional
```
CPBridge, line 2437:
```suggestion
        referenced_coordinates: Union[Sequence[Union[highdicom.sr.CoordinatesForMeasurement, highdicom.sr.CoordinatesForMeasurement3D]], None], optional
```

src/highdicom/sr/templates.py
```diff
@@ -2620,6 +2636,24 @@ def referenced_images(self) -> List[SourceImageForMeasurement]:
         )
         return [SourceImageForMeasurement.from_dataset(m) for m in matches]
 
+    @property
+    def referenced_coordinates(self) -> List[Union[CoordinatesForMeasurement, Scoord3DContentItem]]:
```
CPBridge, line 2640:
```suggestion
    def referenced_coordinates(self) -> List[Union[CoordinatesForMeasurement, CoordinatesForMeasurement3D]]:
```

src/highdicom/sr/templates.py
```diff
@@ -2620,6 +2636,24 @@ def referenced_images(self) -> List[SourceImageForMeasurement]:
...
+        scoord_matches = find_content_items(
+            self[0],
+            value_type=ValueTypeValues.SCOORD
+        )
+        coord = [CoordinatesForMeasurement.from_dataset(m) for m in scoord_matches]
+        scoord3d_matches = find_content_items(
+            self[0],
+            value_type=ValueTypeValues.SCOORD3D
+        )
+        coord.extend([Scoord3DContentItem.from_dataset(m) for m in scoord3d_matches])
```
CPBridge, line 2654:
```suggestion
        coord.extend([CoordinatesForMeasurement3D.from_dataset(m) for m in scoord3d_matches])
```

src/highdicom/sr/templates.py
```diff
@@ -2620,6 +2636,24 @@ def referenced_images(self) -> List[SourceImageForMeasurement]:
         )
         return [SourceImageForMeasurement.from_dataset(m) for m in matches]
 
+    @property
+    def referenced_coordinates(self) -> List[Union[CoordinatesForMeasurement, Scoord3DContentItem]]:
+        """List[Union[highdicom.sr.CoordinatesForMeasurement, Scoord3DContentItem]]:
```
CPBridge, line 2641:
```suggestion
        """List[Union[highdicom.sr.CoordinatesForMeasurement, highdicom.sr.CoordinatesForMeasurement3D]]:
```

tests/test_sr.py
```diff
@@ -2807,6 +2834,28 @@ def test_construction_with_optional_parameters(self):
...
+        measurement = Measurement(
+            name=self._name,
+            value=self._value,
+            unit=self._unit,
+            referenced_coordinates=[scoord, scoord_3d],
+        )
+
+        assert len(measurement.referenced_coordinates) == 2
+        assert scoord in measurement.referenced_coordinates
+        assert scoord_3d in measurement.referenced_coordinates
```
CPBridge, line 2857:
```suggestion
        assert scoord_3d in measurement.referenced_coordinates
        assert isinstance(measurement.referenced_coordinates[0],
        assert isinstance(measurement.referenced_coordinates[1], CoordinatesForMeasurement3D)
```

CPBridge (approved):
Thanks @Fedalto, a nice contribution!

I have tweaked things a bit to use the `CoordinatesForMeasurement3D` class instead of the `Scoord3DContentItem` as this will be less confusing for users I think. However a `Scoord3DContentItem` should still work

Issue #309, opened 2024-10-03
LongCodeValue can cause a failure in search_tree


rhaxton:
In highdicom.sr.utils.py find_content_items()::search_tree() near line:

https://github.com/ImagingDataCommons/highdicom/blob/69281fcfe980065113eb58f4ed9407a41e4e3845/src/highdicom/sr/utils.py#L96

fetching name_code.CodeValue can fail if the ContentSequence uses either LongCodeValue or URNCodeValue instead of CodeValue.

https://dicom.innolitics.com/ciods/comprehensive-sr/sr-document-content/0040a730/0040a168/00080100

Something like:

```
if hasattr(name_code, 'CodeValue'):
    code_value = name_code.CodeValue
elif hasattr(name_code, 'LongCodeValue'):
    code_value = name_code.LongCodeValue
else:
    code_value = name_code.URNCodeValue
```

might be needed here?

CPBridge:
Hi @rhaxton, thanks for reporting the issue. Your suggestion makes sense to me, though I haven't thoroughly thought through whether it might have any other unintended consequences.

Would you be interested in submitting a pull request with your change, and adding a test for this case?

rhaxton-ctx:
Yes, I can try to work something up soon for this

CPBridge:
Resolved by #310

PR #310, opened 2024-10-04
Fix for #309

rhaxton:
Changed sr_document.dcm test case to have a LongCodeValue tag for Language of Content

Changed search_tree in find_content_items to try all three CodeValue tags (CodeValue, LongCodeValue, URNCodeValue)


CPBridge:
Hi @rhaxton, thanks for the PR. I think your code changes are good, except for the failing style checks: https://github.com/ImagingDataCommons/highdicom/actions/runs/11187492783/job/31122527504?pr=310

However, I have one nitpick... I notice that you have changed the attribute used for language code in the example SR document to Long Code Value but do not change the value (`121049`). However, to be 100% strict about this, that is an invalid usage because the standard states ([part 3 chapter 8 section 8.1](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/chapter_8.html#sect_8.1)):

> The Long Code Value (0008,0119) or URN Code Value (0008,0120) is only used for codes that exceed the 16 character size limit of Code Value (0008,0100). If the code value length exceeds 16, the Code Value (0008,0100) shall not be present. If the code value length is 16 characters or less, the Code Value (0008,0100) shall contain the code and neither Long Code Value (0008,0119) nor URN Code Value (0008,0120) shall be present.

I.e. you cannot use Long Code Value here because the value is < 16 characters. While this isn't relevant for the current PR under discussion, I think it is important that any test files we distribute are as correct as possible, as they will likely be used for tests for all sorts of other behaviour and also serve as exemplars on how to construct these files.

Can you think of other correct usage of Long Code Value that we could use here instead? Or alternatively we could revert the change to the file and create a tweaked version in memory during the test by setting it with some long string nonsense value.

CPBridge (changes requested):
See above comment with requested changes

rhaxton:
I reverted the test case file and added two test cases that modify in memory.  I reread the test file in each case to avoid modifying / testing / reverting the shared data, hopefully that is OK?

CPBridge (approved):
Looks good, thanks for your help @rhaxton

PR #315, opened 2024-10-24
Fix SR if `SpacingBetweenSlices` is not set

Fedalto:
[`SpacingBetweenSlices`](https://dicom.innolitics.com/ciods/ct-image/image-plane/00180088) is an optional DICOM tag, so it can not be present. Only add it to `ImageLibraryEntryDescriptors` if available.


CPBridge (approved):
Looks good to me, thanks for the help @Fedalto

PR #318, opened 2025-01-31
fix color error

aporter-turing:
This PR fixes the convert legacy error related to adding RGB dicoms

### DIFF
--- a/src/highdicom/legacy/sop.py
+++ b/src/highdicom/legacy/sop.py
@@ -405,7 +405,8 @@ def _convert_legacy_to_enhanced(
             bits_allocated=ds.BitsAllocated,
             bits_stored=ds.BitsStored,
             photometric_interpretation=ds.PhotometricInterpretation,
-            pixel_representation=ds.PixelRepresentation
+            pixel_representation=ds.PixelRepresentation,
+            planar_configuration = ds.PlanarConfiguration
         )
         for ds in sf_datasets
     ]


CPBridge:
Thanks @aporter-turing for reporting this and for your suggested fix. Unfortunately while your code may "work" and successfully produce a DICOM object, my reading of the standard is that that object would not be a valid Legacy Converted Enhanced MR Image.

There is a constraint that the PhotometricInterpretation have the value "MONOCHROME2", would would appear to rule out any color images being included in the Legacy Converted Enhanced MR Image altogether ([source](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_A.70.3.html#sect_A.70.3.1))

> The Photometric Interpretation (0028,0004) defined in [Section C.8.15.2](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.15.2.html) is MONOCHROME2.

So instead, we should probably check this constraint and raise a more intuitive error message if any input image is not "MONOCHROME2".

CPBridge:
Oops yes the link was wrong but I'm afraid the same is true for legacy converted MR ([source](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_A.71.3.html#sect_A.71.3.1)).

If you want a multiframe MR image, I thenk you would need to use the [Enhancnd MR Color Image](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_A.36.4.3.html#table_A.36-5), but highdicom does not support this at the moment (maybe someday)

CPBridge:
Closing in favour of #321

Issue #324, opened 2025-02-11
MeasurementReport -- `imaging_measurements` could not be None


nicolay-r:
Dear highdicom team,

This is simple bug of accessing non assigned variable in the case when `imaging_measurements` is `None` here:
https://github.com/ImagingDataCommons/highdicom/blob/a4dfcb5e3bb6630b69548e1c89cb9040d7f0ea20/src/highdicom/sr/templates.py#L4297
```bash
...
  File ".venv/lib/python3.10/site-packages/highdicom/sr/templates.py", line 4297, in __init__
    item.ContentSequence.append(container_item)
UnboundLocalError: local variable 'container_item' referenced before assignment
```

**Steps to reproduce:** 
code example for reproduction which is a modified version of the one in the documentation here:
https://highdicom.readthedocs.io/en/latest/quickstart.html#creating-structured-report-sr-documents
```python
import highdicom as hd
from pydicom.sr.codedict import codes


observer_person_context = hd.sr.ObserverContext(
    observer_type=codes.DCM.Person,
    observer_identifying_attributes=hd.sr.PersonObserverIdentifyingAttributes(
        name='Foo'
    )
)
observer_device_context = hd.sr.ObserverContext(
    observer_type=codes.DCM.Device,
    observer_identifying_attributes=hd.sr.DeviceObserverIdentifyingAttributes(
        uid=hd.UID()
    )
)
observation_context = hd.sr.ObservationContext(
    observer_person_context=observer_person_context,
    observer_device_context=observer_device_context,
)

# Create the report content
measurement_report = hd.sr.MeasurementReport(
    observation_context=observation_context,
    procedure_reported=codes.LN.CTUnspecifiedBodyRegion,
)
```

CPBridge:
Ooof yes. Thanks for submitting this issue @nicolay-r 

I think the solution is that creating a MeasurementReport without any measurements should not be allowed (it would certainly be an odd thing to do, which is presumably why this has gone unnoticed). The [table for TID1500](http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1500) states that either an imaging measurement be included, or a qualitative evaluation, or a derived measurement (rows 6, 10, and 12). Since highdicom does not currently support the latter two options, at least one imaging measurement should be required. But of course this should be handled more gracefully than an unbound variable error.

I will work on a fix soon

nicolay-r:
[...]

PR #325, opened 2025-02-13
Disallow measurement report with no measurement groups

CPBridge:
Fix for #324 (@nicolay-r)

Explicitly disallow passing None for the `imaging_measurements` parameter of the `MeasurementReport` class. Previously this failed with a bug relating to an unbound variable.

Note that I have still kept the `imaging_measurements` as an "optional" parameter in terms of the function signature in order to allow expansion in the future to add support for qualitative evaluations or image derived measurements (rows 10 and 12 of the [TID1500 table](http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1500)).

I further looked through static analysis results on this file in case there were other cases with similar "indentation problems" leading to potentially unbound locals. I found one in the `ImageLibrary` class that would prevent multiple image library groups being added to the library. This is fixed also

### DIFF
--- a/src/highdicom/sr/templates.py
+++ b/src/highdicom/sr/templates.py
@@ -5134,58 +5134,63 @@ class ImageLibrary(Template):
 
         """
         super().__init__()
+
+        if len(datasets) == 0:
+            raise ValueError(
+                "Argument 'datasets' must contain at least one item."
+            )
+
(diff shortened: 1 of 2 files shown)


nicolay-r:
@CPBridge Thank you for you time and hard work on this! :rocket:

PR #340, opened 2025-04-06
Set COVERAGE_CORE to speed up tests on 3.12

CPBridge:
Github actions has always been extremely slow on Python 3.12 since it was added to the tests.

I tracked this down to the pytest coverage plugin (`pytest-cov`), which in turn appears to be due to problems with the cpython interpreter itself. It really struggles with the huge `_iods.py` and `_modules.py` files, but simple attempts to exclude them from the coverage measurements didn't work (I'm not sure why).

However, setting the environment variable `COVERAGE_CORE=sysmon`  (see [docs](https://coverage.readthedocs.io/en/latest/cmd.html)) improves performance very dramatically.

I also updated the versions of the testing related dependencies

### DIFF
--- a/.github/workflows/run_unit_tests.yml
+++ b/.github/workflows/run_unit_tests.yml
@@ -18,6 +18,10 @@ jobs:
         python-version: ["3.10", "3.11", "3.12", "3.13"]
         dependencies: [".", "'.[libjpeg]'"]
 
+    env:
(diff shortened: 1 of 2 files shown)


CPBridge:
It seems that the improvement is not that dramatic, but probably worth having

Issue #355, opened 2025-05-26
Support for DICOM Templates TID 3802 Cardiovascular Patient History, TID 3700 ECG Report and TID 2000 Basic Diagnostic Imaging Report


HalvarLilienthal:
Hi,

at @Cardio-AI we've extended Highdicom to support the following DICOM Templates:
- TID 3802 Cardiovascular Patient History
- TID 3700 ECG Report
- TID 2000 Basic Diagnostic Imaging Report

As we've already discussed with you, we'd like to contribute our work to this project by creating a new feature branch for a pull request.
Our contribution consists of ~40 new classes of template implementations, content items, and units, it should contain everything that is needed to represent these three templates in Highdicom.
While we implement tests and verify we enforced your coding style guideline and PEP8 compliance, we'd like to know if and where in your codebase you'd like us to add these extensions.

PR #365, opened 2025-06-24
Restructure templates file into a folder and add TID 2000, TID 3700 and TID 3802

HalvarLilienthal:
Implements #355 

This pull request restructures the [src/highdicom/sr/templates.py](https://github.com/Cardio-AI/highdicom/blob/master/src/highdicom/sr/templates.py) file into a `src/highdicom/sr/templates` folder.

Additionally it adds support for the DICOM templates
- TID 2000 Basic Diagnostic Imaging Report
- TID 3700 ECG Report
- TID 3802 Cardiovascular Patient History

as well as all templates these three depend on.


CPBridge:
Thank you!

It will probably be a while before I have time to review this in depth: I want to get out version 0.26.0 soon with some important bug fixes, and that will require a bit of work on my part. Then I will return to look at this

CPBridge (changes requested):
Hi @HalvarLilienthal 

Sorry for such a long wait for this, and thanks again for the PR.

I have gone through the `common`, `tid1500`, and `tid2000` files and I'm sharing several important comments here. I have not yet gone through the `tid3700` or `tid3802` files, but I will try to do this over the next week.

At a high level, I like the restructuring into files for each top-level template. I think one of the most important concerns is the way you have been using context groups (CIDs), which are not template but define lists of codes that are suggested for use at certain places within a template.

src/highdicom/sr/templates/tid3700.py
```diff
@@ -0,0 +1,1213 @@
...
+
+
+class PriorECGStudy(Template):
+    """:dcm:`TID 3702 <part16/sect_TID_3702.html>`
+    Prior ECG Study
+    """
+
+    def __init__(
+        self,
+        comparison_with_prior_study_done: Union[CodedConcept, Code],
```
CPBridge, line 885: The allowed values for this [CID](https://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_CID_230.html) are just "Yes", "No", and "Undetermined". Maybe to reduce boilerplate, we just allow users to pass a bool here and construct the corresponding code if they do?

src/highdicom/sr/templates/tid3700.py
```diff
@@ -0,0 +1,1213 @@
...
+                    'have type CompositeContentItem.'
+                )
+            content.append(prior_report_for_current_patient)
+        if source_of_measurement is not None:
+            if not isinstance(source_of_measurement, WaveformContentItem):
+                raise TypeError(
+                    'Argument "source_of_measurement" must have type ' +
+                    'WaveformContentItem.'
+                )
+            content.append(source_of_measurement)
```
CPBridge, line 941: Unfortunately, this approach allows the user to construct content items with the incorrect relationship type, and it isn't checked. We might have to create specialized subtypes unfortunately.

src/highdicom/sr/templates/tid3700.py
```diff
@@ -0,0 +1,1213 @@
...
+        if len(content) > 0:
+            item.ContentSequence = content
```
CPBridge, line 943: Check is unnecessary since some items are required. This occurs in several places. I haven't left a comment on all of them

src/highdicom/sr/templates/tid3700.py
```diff
@@ -0,0 +1,1213 @@
...
+class ECGFinding(Template):
+    """:sct:`EV 271921002`
+    ECG Finding
+    """
+
+    def __init__(
+        self,
+        value: Union[CodedConcept, Code],
+        equivalent_meaning_of_value: Optional[str] = None,
+        ecg_findings: Optional[Sequence["ECGFinding"]] = None
```
CPBridge, line 956: I see why you've done it but think we need to avoid this recursive approach for at least two reasons.

- The lower level findings should have relationship type "INFERRED FROM" and the upper level findings should have "CONTAINS". Therefore you cannot use the same class for both. In the current implementation, the relationship types at the lower level are incorrect.
- Your current approach does not prevent more than two levels of findings, though this would not be allowed by the template.

A separate issue is that the parameter name `ecg_findings` does not in my opinion clearly convey the purpose of the lower level findings as being findings from which the upper level findings are inferred. The parameter should probably be renamed to something like `inferred_from`?

I think you will either need to define yet another class for the lower level "inferred from" findings, or have the user pass plain codes for the lower level findings and construct those content items internally. The latter approach does not allow for "equivalent meaning of value" text at the lower level, however.

src/highdicom/sr/templates/tid3700.py
```diff
@@ -0,0 +1,1213 @@
...
+        ecg_finding_text: Optional[str] = None,
+        ecg_finding_codes: Optional[Sequence[ECGFinding]] = None
+    ) -> None:
+        item = ContainerContentItem(
+            name=codes.DCM.QualitativeAnalysis,
+            template_id='3717',
+            relationship_type=RelationshipTypeValues.CONTAINS
+        )
+        content = ContentSequence()
+        if ecg_finding_text is None and ecg_finding_codes is None:
```
CPBridge, line 1012: You should also check that they are not both empty sequences

src/highdicom/sr/templates/tid3700.py
```diff
@@ -0,0 +1,1213 @@
...
+
+class ECGQualitativeAnalysis(Template):
+    """:dcm:`TID 3717 <part16/sect_TID_3717.html>`
+    ECG Qualitative Analysis
+    """
+
+    def __init__(
+        self,
+        ecg_finding_text: Optional[str] = None,
+        ecg_finding_codes: Optional[Sequence[ECGFinding]] = None
```
CPBridge, line 1004: Minor, but perhaps there could just be one parameter `ecg_findings: Sequence[str | ECGFinding]` and then you could check the type internally to figure out what to do with them.

CPBridge: Also, maybe "ecg" in the parameter names isn't needed, given the context?

src/highdicom/sr/templates/tid3700.py
```diff
@@ -0,0 +1,1213 @@
...
+
+class SummaryECG(Template):
+    """:dcm:`TID 3719 <part16/sect_TID_3719.html>`
+    Summary, ECG
+    """
+
+    def __init__(
+        self,
+        summary: Optional[str] = None,
+        ecg_overall_finding: Optional[Union[Code, CodedConcept]] = None
```
CPBridge, line 1057: include link to [CID 3677](https://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_CID_3677.html)

src/highdicom/sr/templates/tid3700.py
```diff
@@ -0,0 +1,1213 @@
...
+    ) -> None:
```
CPBridge, line 1058: Prevent construction if both parameters are `None`?

CPBridge (changes requested):
@HalvarLilienthal 

ok, I made it through TID3700 and left several more comments. I still need to look through 3802.

HalvarLilienthal:
Hi @CPBridge, thank you so much for your in depth review! I'll work on your improvement ideas over the next couple of weeks, use your suggestions where I agree and respond to the topics open for discussion whenever I find the time. Sorry for all the inconsistencies, multiple people contributed on our end but I am confident we can bring these classes up to your standard.

src/highdicom/sr/templates/common.py
```diff
@@ -0,0 +1,1394 @@
...
+        ]
+
+
+class PressureUnit(CIDUnits):
+    def __init__(self, mmHg: Union[int, None] = None,
+                 kPa: Union[int, None] = None) -> None:
+        self.units = [
+            {'value': mmHg, 'code_value': "mm[Hg]", 'code_meaning': "mmHg"},
+            {'value': kPa, 'code_value': "kPa", 'code_meaning': "kPa"}
+        ]
```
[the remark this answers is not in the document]
CPBridge: See also [this comment](https://github.com/ImagingDataCommons/highdicom/pull/365#discussion_r2314272585)

src/highdicom/sr/templates/tid2000.py
```diff
@@ -0,0 +1,317 @@
...
+
+class EquivalentMeaningsOfConceptNameText(TextContentItem):
+    """:dcm:`TID 1210 <part16/chapter_A.html#sect_TID_1210>`
+    Equivalent Meaning(s) of Concept Name Text
+    """
+
+    def __init__(self,
+                 value: str,
+                 language_of_value: Optional[LanguageOfValue] = None
+                 ) -> None:
```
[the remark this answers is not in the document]
HalvarLilienthal: Fixed this and all other occurrences in 228ae340419e7d150368a8f9d942bd07dd89b38b

CPBridge:
@HalvarLilienthal  I just realised that I inadvertently closed this PR by deleting the branch it was based on. My apologies for that. This doesn't mean that I am not still interested in incorporating these changes if you have time to work on it: but it would need to be re-based on either master or the current development branch (currently v0.28.0dev)

PR #373, opened 2025-08-22
Change to definition of SR evidence

CPBridge:
Fixes for issues related to SR from @deepakri201 's conversion project:

- Replace use of `pydicom.sr.coding.codes.SCT.Source` with a manual code to remove the unwanted "(attribute)" part of the meaning string. Also fix some outdated codes found in doc pages
- Change the behavior of the `highdicom.sr.utils.collect_evidence` function in regard to how it determines which evidence goes in the CurrentRequestProcedureEvidenceSequence and which into the PertinentOtherEvidenceSequence. Previously this was based on whether the item is found in the SR content tree. Now it is based on whether the item belongs to the same study as the new SR document.

### DIFF
--- a/src/highdicom/sr/utils.py
+++ b/src/highdicom/sr/utils.py
@@ -290,8 +290,12 @@ def collect_evidence(
         for ref in references
     }
     evd_uids = set()
-    same_study_group: Mapping[tuple[str, str], list[Dataset]] = defaultdict(list)
-    other_study_group: Mapping[tuple[str, str], list[Dataset]] = defaultdict(list)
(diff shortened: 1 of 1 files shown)


fedorov:
> Replace use of `pydicom.sr.coding.codes.SCT.Source` with a manual code

@CPBridge why not try to fix it in pydicom?

And on that note, do you know if pydicom is taking those from here: https://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_O.html ? Since in that table most of the items have the unwanted qualifiers ...

Issue #371, opened 2025-08-07
Add RecommendedDisplayCIELabValue parameter to Annotation Group


CPBridge:

CPBridge:
Can also use this implementation to compare: https://github.com/DCMTK/dcmtk/blob/master/dcmiod/libsrc/cielabutil.cc

PR #374, opened 2025-08-26
Add CIELabColor to AnnotationGroup

CPBridge:
- Add conversions between RGB and CIELab
- Add the option to include a recommended display color in an annotation group

Addresses #371


tests/test_ann.py
```diff
@@ -161,6 +162,8 @@ def test_construction(self):
             ),
         ]
 
+        color = CIELabColor(40.0, 50.0, 6.0)
```
fedorov, line 165: Maybe add a comment what that color looks like, just so that the users know what to expect?

PR #394, opened 2026-02-20
Feature/itk volume conversions

mccle:
Depends on #387 for optional dependency importing.

Introduces 'to' and 'from' methods for the itk and SimpleITK libraries on the `highdicom.Volume` class.


CPBridge (changes requested):

src/highdicom/volume.py
```diff
@@ -3583,6 +3585,210 @@ def pad_array(array: np.ndarray, cval: float) -> float:
...
+                ' Safely recasting to float32.'
+            )
+            array = array.astype(np.float32)
+
+        elif array.dtype == np.float128:
+            f64 = np.finfo(np.float64)
+            if array.min() >= f64.min and array.max() <= f64.max:
+                warnings.warn(
+                    'SimpleITK does not support float128 data.'
+                    ' Safely recasting to float64.'
```
CPBridge: This is not really safe as it still involves a loss of precision (I think...)

CPBridge: Also I think "casting" would be preferred to "recasting"

.github/workflows/run_unit_tests.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         # and the likelihood of 3.12-specific bugs is considered low at this
         # stage
         python-version: ["3.10", "3.11", "3.13", "3.14"]
-        dependencies: [".", "'.[libjpeg]'"]
+        dependencies: [".", "'.[libjpeg]'", "'.[itk]'", "'.[sitk]'"]
```
CPBridge: This is going to run tests a lot more than needed I think. I think this would suffice:


```suggestion
        dependencies: [".", "'.[libjpeg,itk,sitk]'"]
```

I.e. once with none of the optional dependencies and once with all of them. Since they shouldn't really interact with each other, I think this is fine

tests/test_sitk.py
```diff
@@ -0,0 +1,583 @@
...
+            match=(
+                'SimpleITK does not support float128 data.'
+                ' Safely recasting to float64 is not possible.'
+            )
+        ):
+            sitk_roundtrip = Volume.from_sitk(volume.to_sitk())
+
+    else:
+        sitk_roundtrip = Volume.from_sitk(volume.to_sitk())
+        assert sitk_roundtrip.dtype == volume.dtype
```
CPBridge, line 504: Wouldn't it be a good idea to test for equality of the arrays here? (except in cases where casting may lead to small differences)

tests/test_sitk.py
```diff
@@ -0,0 +1,583 @@
...
+
+
+@pytest.mark.parametrize(
+    'zip_url,nifti_url',
+    [
+        (
+            'https://github.com/neurolabusc/dcm_qa_mprage/raw/refs/heads/main/In/2_t1_mp2rage_sag_p3_32.zip',  # noqa: E501
```
CPBridge: I suggest that is might be neater to factor out `"https://github.com/neurolabusc/dcm_qa_mprage/raw/refs/heads/main"`

tests/test_sitk.py
mccle: I can add equivalence tests to all of the "safe" casts and us `np.allclose` for the `float128` to `float64` cast. However, in this case should the original array also be updated to be of random values as opposed to `np.zeros`?

docs/volume.rst
```diff
@@ -731,60 +731,97 @@ Writing a volume to a NIfTI file:
...
+
+    vol = hd.Volume.from_itk(
+        itk_im=itk_im,
+        coordinate_system='PATIENT',
+        frame_of_reference_uid=None
     )
 
-Creating an ITK Image from a Volume:
+
+SImpleITK
```
CPBridge: typo
```suggestion
SimpleITK
```

src/highdicom/volume.py
```diff
@@ -3583,6 +3585,307 @@ def pad_array(array: np.ndarray, cval: float) -> float:
             channels=self._channels,
         )
 
+    def to_sitk(self) -> 'SimpleITK.Image':  # noqa: F821
+        """Convert the Volume to ``SimpleITK.Image`` format.
+
+        The Volume is converted to a 3D ``SimpleITK.image``. If
```
```suggestion
        The Volume is converted to a 3D ``SimpleITK.Image``. If
```

docs/volume.rst
```diff
@@ -731,60 +731,97 @@ Writing a volume to a NIfTI file:
...
+Conversion is essentially the same as the ITK interface, also using the "LPS"
+convention and reversing the order of dimensions of the array. This permutation is
+again handled automatically by highdicom and requires no intervention by the user.
+
+Creating a SimpleITK Image from a Volume:
 
 .. code-block:: python
 
-    import itk
+    import SimpleITK as sitk
```
CPBridge: This is unnecessary right?

CPBridge (changes requested):

docs/volume.rst
mccle: Yes, the import should be handled in the to method. I can remove them.

tests/test_sitk.py
mccle: There were a few unique roots, but all of them have been factored out into `DCM_QA_*` constants.

CPBridge (approved):

PR #399, opened 2026-03-19
Remove packaging version specifier

CPBridge:
The packaging version specifier is currently quite a recent version, meaning that it could be quite restrictive on the compatibility of highdicom with other packages. This is likely unnecessary because we use only very basic functionality from `packaging`, and could be rather frustrating (e.g. I already found it makes highdicom incompatible with a somewhat recent `mlflow` version).

I propose we remove the version specifier, unless you have a more specific suggestion. What do you think @mccle?


CPBridge:
Merging after discussion with @mccle and using `packaging>=17.1` lower limit

Issue #406, opened 2026-04-04
Feature request: Support for DICOM Supplement 247 Eyecare Measurement Templates (TID 6001–6009)


oftalmos-org:
## Summary

DICOM Supplement 247 \"Eyecare Measurement Templates\" was included in DICOM 2025c (July 2025). It defines structured report templates specifically for ophthalmic measurements that should replace the current practice of encoding these measurements in generic TID 1500 reports.

We are building a multi-vendor OCT → DICOM pipeline (Optopol Revo FC130, Zeiss Cirrus HD-OCT) that currently uses TID 1500 via highdicom. We would like to migrate to the correct ophthalmic templates once highdicom supports them.

## TIDs needed

| TID  | Name                                      | Use case                        |
|------|-------------------------------------------|---------------------------------|
| 6001 | Ophthalmology Measurements Group          | Root container                  |
| 6004 | Circumpapillary RNFL Key Measurements     | RNFL circle scan (OCT)          |
| 6005 | Macular Thickness Key Measurements        | ETDRS grid, CMT (OCT)           |
| 6006 | Ganglion Cell Layer Key Measurements      | mGCIPL (OCT)                    |
| 6002 | Visual Field Key Measurements             | Static perimetry (HFA, PTS 925) |
| 6003 | Optic Disc Key Measurements               | C/D ratio, ONH                  |
| 6007 | Corneal Topography Key Measurements       | Topography                      |

## Reference

- Supplement 247 PDF: https://dicom.nema.org/Dicom/News/January2025/docs/sups/sup247.pdf
- DICOM PS3.16 templates: TID 6001–6009

## Question

Is there a planned release or branch for Supplement 247 support? If not, we would be interested in contributing an implementation.

CPBridge:
Hi @oftalmos-org 

My position on SR templates is that there are so many and we do not have the resources to possibly implement all of them unfortunately. We give TID1500 special status because it is a very general-purpose template. The templates you listed are not a priority for us, sorry.

That said, if you or people like you have certain templates that are important to you/them and are willing to put together high-quality implementations, then I would gratefully accept them into the library. If you want to go ahead with this, I am also happy to advise on how to go about it if you have questions. Please make sure to try to follow the style of the existing templates and re-use components where you can. I would also look at the discussion on #355 and #365 where a contributor previously worked on adding a different set of templates. It seems that may be abandoned now unfortunately, but the discussion including the design decisions taken and some of the missteps made there too would be useful for you.

[...]

PR #407, opened 2026-04-11
feat: restructure sr/templates → package; add Supplement 247 TIDs 6001/6004/6005

oftalmos-org:
## Summary

Implements DICOM Supplement 247 "Eyecare Measurement Templates" (incorporated into PS3.16 2025c) and restructures `highdicom/sr/templates` from a single module into a package, as discussed in #355.

### Templates added

| TID | Class | Description |
|-----|-------|-------------|
| TID 6001 | `OphthalmologyMeasurementsGroup` | Sub-template: single-eye measurement group; finding site constrained to Eye (81745001, SCT); laterality mandatory (Row 3 `>>` nested inside Finding Site item) |
| TID 6004 | `CircumpapillaryRNFLKeyMeasurements` | Root template: circumpapillary RNFL thickness via OCT (CID 42x3) |
| TID 6005 | `MacularThicknessKeyMeasurements` | Root template: ETDRS macular grid thickness via OCT (CID 42x4) |

### Package restructure (`sr/templates/`)

- `templates.py` → `templates/tid1500.py` (rename only, no content changes)
- `templates/__init__.py` re-exports all previous public names — fully backwards-compatible
- `templates/tid6000.py` — new module for Supplement 247 TIDs

### Provisional concept codes

Supplement 247 was balloted with several DCM codes shown as `nnnXXX` placeholders (final values pending WG-09 assignment). They are stored here under private scheme `99OPHTHALMO` (≤ 16 chars, VR SH) with inline comments marking them as draft.

Codes pending a companion PR to pydicom (`pydicom.sr.codedict`):

- Container titles: `nnn102` ("RNFL Key Measurements"), `nnn103` ("Macular Thickness Key Measurements")
- CID 42x3 RNFL: `nnn400`–`nnn404` (average/quadrant thicknesses), `nnn406` (ROI radius)
- CID 42x4 Macular: `nnn250` ("Average macular thickness")
- ETDRS grid (LOINC, stable, not yet in pydicom convenience dict): `LN 57108-3` through `LN 57118-2`

No `coding_ophthalmology.py` was created, per CPBridge guidance in #406.

### Tests

New file `tests/test_sr_ophthalmic.py` — 41 tests:

- `TestOphthalmologyMeasurementsGroup` (14): laterality nesting (Row 3 `>>`), finding site default/custom, tracking identifier, multiple measurements, validation errors
- `TestCircumpapillaryRNFLKeyMeasurements` (13): container code, template ID, bilateral, 99OPHTHALMO scheme check, full quadrant encoding
- `TestMacularThicknessKeyMeasurements` (12): LOINC scheme check, `uL` unit for total volume, full 10-item ETDRS grid, bilateral
- `TestRoundtrip` (2): TID 6004 and TID 6005 each serialised to `ComprehensiveSR` → `dcmwrite` → `dcmread` → values verified

All 41 new tests pass. All 294 pre-existing `test_sr.py` tests pass.

### Validation

Validated with real clinical OCT data from a production Optopol Revo FC130 unit. CMT = 281.4 µm.

This PR also implements the templates/ package restructure discussed in #355 and designed by @CPBridge.


oftalmos-org:
@CPBridge This PR is ready for review. All 41 new ophthalmic SR tests pass alongside the 294 existing test_sr.py tests. Happy to address any feedback.

CPBridge:
Great! I will take a look at this soon, sorry if it takes a few weeks as I have a lot on my plate at the moment

src/highdicom/sr/templates/tid6000.py
```diff
@@ -0,0 +1,635 @@
...
+from pydicom.sr.coding import Code
+
+from highdicom.sr.coding import CodedConcept
+from highdicom.sr.enum import RelationshipTypeValues
+from highdicom.sr.value_types import (
+    CodeContentItem,
+    ContainerContentItem,
+    ContentSequence,
+    TextContentItem,
+    UIDRefContentItem,
```
CPBridge: Unused import
```suggestion
```

src/highdicom/sr/templates/tid6000.py
```diff
@@ -0,0 +1,635 @@
...
+    Template,
+)
+
+# ---------------------------------------------------------------------------
+# Supplementary UCUM unit not yet in pydicom.sr.codedict
+# ---------------------------------------------------------------------------
+#: UCUM microliter — used for macular total volume (``LN 57118-2``).
+#: Not exposed as a convenience attribute in pydicom 3.x; defined here until
+#: pydicom adds it to ``codes.UCUM``.
+UCUM_MICROLITER: Code = Code('uL', 'UCUM', 'microliter')
```
CPBridge: I would rather not have this as part of highdicom's public API.

src/highdicom/sr/templates/tid6000.py
```diff
@@ -0,0 +1,635 @@
...
+    :class:`CircumpapillaryRNFLKeyMeasurements` (TID 6004) and
+    :class:`MacularThicknessKeyMeasurements` (TID 6005).
+    """
+
+    def __init__(
+        self,
+        laterality: CodedConcept | Code,
+        measurements: Sequence[Measurement],
+        finding_site: CodedConcept | Code | None = None,
+        tracking_identifier: str | None = None,
```
CPBridge: We already have an implementation of TID 4108 "Tracking Identifier" template in what is now `tid1500.py`. It would seem to make more sense to use that here, since it includes both the text and UID identifiers.

src/highdicom/sr/templates/tid6000.py
```diff
@@ -0,0 +1,635 @@
...
+            Free-text identifier for tracking this measurement group across
+            studies or analyses.  Encoded as
+            ``(112039, DCM, "Tracking Identifier")`` with relationship type
+            HAS_OBS_CONTEXT (row 6 of TID 60x1).
+
+        """  # noqa: E501
+        super().__init__()
+
+        if finding_site is None:
+            finding_site = codes.cid4209.Eye
```
CPBridge: My reading of the template is that this value is fixed as "Eye" and it would be wrong to use anything else (this is the meaning of having an EV, =enumerated value in the value set constraint). So I think this should just be hardcoded with no `finding_site` parameter allowing the user to change it

src/highdicom/sr/templates/tid6000.py
```diff
@@ -0,0 +1,635 @@
...
+    :class:`MacularThicknessKeyMeasurements` (TID 6005).
+    """
+
+    def __init__(
+        self,
+        laterality: CodedConcept | Code,
+        measurements: Sequence[Measurement],
+        finding_site: CodedConcept | Code | None = None,
+        tracking_identifier: str | None = None,
+    ) -> None:
```
CPBridge: While generally I don't think we need to include every optional row of the template's table, I think the "source of measurement" (row 11) would be a particularly important one to include such that measurements can be traced back to the image in which they were made

src/highdicom/sr/templates/tid6000.py
```diff
@@ -0,0 +1,635 @@
...
+        # item, NOT of the container.  The double-indent (>>) in Table TID
+        # 60x1 encodes this nested modifier relationship.
+        site_item.ContentSequence = ContentSequence()
+        lat_item = CodeContentItem(
+            name=CodedConcept(
+                value='272741003',
+                scheme_designator='SCT',
+                meaning='Laterality',
+            ),
+            value=laterality,
```
CPBridge: CID244 is non-extensible, so I think it probably makes sense to constrain the value to the left/right codes (i.e. raise a value error if another value is passed)?

src/highdicom/sr/templates/tid6000.py
```diff
@@ -0,0 +1,635 @@
...
+
+        if not measurement_groups:
+            raise ValueError(
+                "Argument 'measurement_groups' must contain at least one item."
+            )
+        if len(measurement_groups) > 2:
+            raise ValueError(
+                "Argument 'measurement_groups' must contain at most two items "
+                "(one per eye)."
+            )
```
CPBridge: Why this restriction? The VM for this row is 1-n and the notes explicitly state

> TID 2120 may be invoked once for each eye, or multiple times for a trend series for a single eye

src/highdicom/sr/templates/tid6000.py
```diff
@@ -0,0 +1,635 @@
...
+        )
+        report = CircumpapillaryRNFLKeyMeasurements(
+            algorithm_id=algo,
+            measurement_groups=[group],
+        )
+    """
+
+    def __init__(
+        self,
+        algorithm_id: AlgorithmIdentification,
```
CPBridge: Since we already have a class for TID1002 ObserverContext, it would be straightforward to add this as an optional attribute

src/highdicom/sr/templates/tid6000.py
```diff
@@ -0,0 +1,635 @@
...
+        # Row 8 — Measurements (MC, 1-n, CONTAINS via TID 300).
+        for meas in measurements:
+            item.ContentSequence.extend(meas)
+
+        self.append(item)
+
+
+# ---------------------------------------------------------------------------
+# TID 6004 (Sup247 TID 60x4) — Retinal Nerve Fiber Layer Key Measurements
+# ---------------------------------------------------------------------------
```
CPBridge: I would prefer not to have these superfluous "header" comments

src/highdicom/sr/templates/tid6000.py
```diff
@@ -0,0 +1,635 @@
...
+
+        if not measurement_groups:
+            raise ValueError(
+                "Argument 'measurement_groups' must contain at least one item."
+            )
+        if len(measurement_groups) > 2:
+            raise ValueError(
+                "Argument 'measurement_groups' must contain at most two items "
+                "(one per eye)."
+            )
```
CPBridge: As above, I don't think that this restriction is in line with the standard

src/highdicom/sr/templates/tid6000.py
```diff
@@ -0,0 +1,635 @@
...
+        item.ContentSequence.extend(language_of_content_item_and_descendants)
+
+        # Row 4 — Algorithm Identification (M, TID 4019, HAS_OBS_CONTEXT).
+        item.ContentSequence.extend(algorithm_id)
+
+        # Row 5 — Ophthalmology Measurements Group (M, 1-2, CONTAINS).
+        for group in measurement_groups:
+            item.ContentSequence.extend(group)
+
+        super().__init__([item], is_root=True)
```
CPBridge: How about row 7 (the symmetry measurement)?

CPBridge (changes requested):
Hi @oftalmos-org 

I'm sorry it took me so long to get to this. The last few months have been very busy with personal things, but now I have my head around the templates and what you are trying to do here, I hope future iterations will be much quicker.

I have three high-level comments, and then a bunch of smaller ones below.

1. Firstly, and most importantly, it appears that Supplement 247 was approved into the standard in version 2025b (according to [David Clunie's page](https://dclunie.com/dicom-status/status.html)). It seems your implementation was based on an earlier draft and there have been some changes. Most notably the TIDs have changed:

    - TID 6001 in the draft become [TID 2120](https://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_TID_2120.html)
    - TID 6004 became [TID2123](https://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_TID_2123.html)
    - TID 6005 became [TID2124](https://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_TID_2124.html)

    Would you please update this in line with the approved and released version of the standard?

    Relatedly, I would quite like to avoid releasing the codes required as part of the public API of highdicom, since the intention would be to use them only temporarily. To that end, what is the status of the PRs to add required codes to pydicom? I took a look and was not able to see one.

2. TID 2120 (formally TID 6001 in the draft version) is described as "a proper subset of TID 1501 with some optional extensions". Highdicom already has class representing TID1501, namely `highdicom.sr.MeasurementsAndQualitativeEvaluations`, so I'm wondering whether this class should be a subclass of that class (or possibly a subclass of its base class `highdicom.sr._MeasurementsAndQualitativeEvaluations)`. That would seem to remove some redundant code and allow this class to take advantage of some of the optional parts that are already implemented. It would also fix some of the other issues I highlight below

3. It took me a while to get my head around what I think is a new pattern in the standard of invoking templates with entire context groups defining the required measurements that must be present within a measurement group. It is a little tricky to fit this pattern into the way we built templates in the past. It seems that the approach that you have taken is to document which measurements are required within the top level classes (CircumpapillaryRNFLKeyMeasurement and MacularThicknessKeyMeasurements). I am concerned that this is insufficient, and that many (most?) SRs produced would likely be non-compliant due to a user failing to follow these requirements. This is at odds with our philosophy that if a highdicom constructor finishes successfully, the new object is compliant with the standard.

    I see two potential ways around this:

    1. We keep things more or less as they are currently structured, but add logic to the constructors of these two top level classes to check that the measurement groups passed to them meet all the requirements for the template, and give clear error messages if they do not. This would be easier to implement, but I do think putting the whole burden on the user to pass the right combination of parameters will be quite confusing to the user.
    2. We implement one specialized subclass of the OpthalmologyMeasurementsGroup (or MeasurementsAndQualitativeEvaluations, in line with the second point above) for each invocation of that template. These classes would accept as parameters the measurement values and place them into measurements with the correct codes. We might want to remove some of other optional parameters of the Measurement (TID300) class, or assume they are shared between all measurements in the group, to simplify things a bit. Then the top-level classes accept these specialized subclasses, rather than any MeasurementGroup. While this would be more complex to implement (and there would certainly be more details to figure out), I think the result would be a much more simplified process for the user. What do you think?

oftalmos-org:
Hi @CPBridge ,

Thank you for the detailed review — and no worries about the time, I really appreciate you working through this carefully and myself as well can surely relate to having a busy mind/life.

I have mainly three responses:

1. TID renumbering: Confirmed, we were working from the draft. We will audit the full PR against the ratified 2025b TIDs (2120/2123/2124) and update all references before the next iteration. Regarding the pydicom PR for the required codes — we have not opened one yet. We will do that in parallel and link it here.

2. Inheritance from TID 1501: We agree that subclassing MeasurementsAndQualitativeEvaluations (or its base _MeasurementsAndQualitativeEvaluations) is the right approach since TID 2120 is defined as a proper subset of TID 1501. Before we commit to one or the other: do you have a preference between the public class and the private base? Our instinct is the private base gives more flexibility for the ophthalmic-specific extensions, but we want your guidance.

3. Conformance enforcement: We prefer Option B — specialized subclasses for each template invocation that accept measurement values directly (e.g. rnfl_superior, rnfl_inferior as float parameters) rather than pre-built Measurement objects. The principle that a successful constructor guarantees a compliant object is the right design goal and Option A would leave a confusing API. We understand this is more work and we are committed to it.

We will come back with the updated TIDs, the pydicom PR linked, and a revised architecture implementing Option B with proper inheritance from TID 1501.

Thanks again,
Noel

CPBridge:
Thanks Noel, sounds excellent!

Regarding your question 2 about using the public class or the private base class. To answer this, it helps to first understand why it was done this way in the first place. The base class `_MeasurementsAndQualitativeEvaluations` captures all the behaviour shared by TIDS 1501 ("Measurements And Qualitative Evaluations"), 1410  ("Planar ROI Measurements And Qualitative Evaluations"), and 1411 ("Volumetric ROI Measurements And Qualitative Evaluations"). While the latter two are more complex, as currently implemented there is only one element of TID 1501 that is not shared by the others, and that is the `source_images` parameter, which corresponds to row 10b of the TID1501 table, and the corresponding property that accesses the source images.

Since TID2120 does include `source_images` (row 11), I think it makes sense to base the class directly on the public `MeasurementsAndQualitativeEvaluations` class so that you inherit the behaviour relating to sources image. I don't foresee any problems adding further ophthalmology-specific options in the future, since it seems that the standard is trying to keep it as a proper subset. However, I could be convinced to change my mind if you try to implement this and come into problems I haven't seen.

There is one little problem already though, which I just noticed.
The source images supported by the `MeasurementsAndQualitativeEvaluations` class currently uses the code SCT 260753009 "Source Image" as suggested by the note for row 10 of TID 1501, whereas TOD 2120 uses DCM 121112 "Source of Measurement". However, looking more into this, I think the existing implementation of `MeasurementsAndQualitativeEvaluations` should use DCM 121112 per its invocation with ImagePurpose = CID 7551 in its parent TID1500.

I have created [this PR](https://github.com/ImagingDataCommons/highdicom/pull/412) (into v0.28.0dev branch) to address this issue. You will need to rebase on that when I merge it. Let me know if you see any problems!

oftalmos-org:
Hi Chris,

Thank you for the clarification on the base class — that makes perfect sense. Inheriting source_images from MeasurementsAndQualitativeEvaluations directly is exactly the right move, and I appreciate you digging into the why behind the class hierarchy.
I've seen your PR for the DCM 121112 fix and will rebase on v0.28.0dev once it merges.
I'm now working on the revised implementation with the updated TIDs (2120/2123/2124), Option B specialized subclasses, the inheritance from MeasurementsAndQualitativeEvaluations, and a companion PR to pydicom for the required codes. I'll be back with a substantive update soon.

Thanks again for your time and patience with this.
Noel

oftalmos-org:
@mafolgueras could you please help us with the architecture decisions please?

oftalmos-org:
Dear @CPBridge,

Following up on PR #407 — I'm trying to map out how Transducin 
(the open-source pipeline that generated the corpus we're validating 
the highdicom implementation against) should structure its AI 
classification outputs as DICOM SRs.

Transducin produces two types of results per eye:
  1. Quantitative measurements (CMT, RNFL thickness) — already handled 
     by the SR builder aligned with PR #407 templates.
  2. Qualitative AI classification (e.g., Beckman AMD stage: 
     Early / Intermediate / Late / nAMD) — this is the new piece.

For the qualitative classification output, should I:
  a) Use TID 1500 with a QualitativeEvaluation content item 
     (CODE type, private coding scheme "RETINAOS" for Beckman stages 
     until WG-09 standardizes them), nested under the same 
     MeasurementsAndQualitativeEvaluations group from PR #407?
  b) Or open a separate Comprehensive SR that references the same 
     source OCT SOP Instance?

Also: is there precedent in highdicom for private coding schemes 
alongside SCT/DCM within the same TID 1500 group, or is that 
considered bad practice?

Thanks

CPBridge:
Hi @oftalmos-org 

For the first question from the limited information provided I think both would be appropriate options. Certainly it is fine to store a mix of measurements and qualitative evaluates in the same group, as long as the image or region they are derived from is the same and other descriptive metadata (observer context etc) is shared. Of course there's also a third option of multiple measurement groups within the same SR, which would make sense if most metadata are shared but the images or regions differ

I am a little unsure what you are asking in the second question. No there is no precedent *within* highdicom for private codes. But highdicom doesn't tell you what codes to use, except where particular codes are required by the standard. It sounds like you are maybe asking whether it's okay for you to use highdicom to create an SR for your project that mixes private and standard codes. If so, yes that is fine and done quite often

oftalmos-org:
Thank you Chris — both answers are very helpful.

For Q1: the use case is a single macular study where all findings (CMT, ETDRS sectors, RNFL, and a future qualitative AMD classification) derive from the same source OCT instance with shared observer context — so a single group with mixed measurements and qualitative evaluations is the right approach. Good to know that multiple groups within the same SR is also an option for cases where the source image or region differs (e.g., a macular measurement group and a disc measurement group in the same SR).

For Q2: the deeper question was about the mechanics of adding a qualitative finding with a private code — specifically a Beckman AMD stage classification in this case — as a CodeContentItem or similar alongside the existing SCT/DCM numeric measurements. The 99OFTALMOS scheme is provisional and intended to be replaced by Supplement 247 codes (CID 4285/4283) once those are fully supported in highdicom via PR #407. In the interim, could you point to the right highdicom pattern for a QualitativeEvaluation with a private CodedConcept as the value — is it simply a CodeContentItem with a CodedConcept using a private scheme_designator, registered via CodingSchemeIdentificationSequence in the SR header?

On PR #407 itself — I will rebase on your v0.28.0dev branch once the Source of Measurement fix merges and address the remaining review items then.

oftalmos-org:
**Dear @CPBridge,**

Just wanted to drop a quick update and say THANK YOU. Your guidance on this PR and on #406 has been invaluable, and working through the Supplement 247 structure with you taught me a lot about how these templates are meant to compose.

The real-world use referenced in this PR is now documented in a published preprint:

**Jaurrieta Hinojos JN et al. Transducin: An Open-Source Python Pipeline for Parsing the Undocumented Optopol Revo FC130 .OPT Format and Generating DICOM Structured Reports. medRxiv 2026. https://doi.org/10.64898/2026.07.14.26357256**

The pipeline [https://github.com/oftalmos-org/transducin] uses TID 1500/1501 via highdicom in production across 475 scans from two clinical sites. The plan is to migrate to TID 6004/6005 once this PR merges. Validation dataset on Zenodo: **https://doi.org/10.5281/zenodo.20031717**

**highdicom** is cited in the paper, and you're acknowledged by name — none of this would have been possible without the library you've built. 


**Once more thank you, 
Noel**

CPBridge:
Hi @oftalmos-org thanks for this information, I'm glad this was productive. It's cool to see what people are doing with the library and I appreciate the recognition. I will have to read your paper in more detail

Sorry that I never replied to your earlier question here:

> In the interim, could you point to the right highdicom pattern for a QualitativeEvaluation with a private CodedConcept as the value — is it simply a CodeContentItem with a CodedConcept using a private scheme_designator, registered via CodingSchemeIdentificationSequence in the SR header?

Yes it is simply a CodeContentItem with a CodedConcept using a private scheme_designator. We do not actually have an implementation of CodingSchemeIdentificationSequence. I do not believe it is required, but would definitely be considered a best practice. I created #451 about this and will try and get it into the next release as it should be pretty straightforward.

> On PR https://github.com/ImagingDataCommons/highdicom/pull/407 itself — I will rebase on your v0.28.0dev branch once the Source of Measurement fix merges and address the remaining review items then.

A small note of housekeeping that we just released `0.28.0` so now please rebase on `v0.29.0dev` branch

oftalmos-org:
Hi Chris,

Thanks for the clarification on private CodedConcepts. It is good to know CodingSchemeIdentificationSequence isn't required, and #451 is a welcome addition.

Here is the preprint DOI if you'd like it handy when you get to reading: 
https://doi.org/10.64898/2026.07.14.26357226 (medRxiv, CC BY 4.0). 

highdicom is cited as a core dependency throughout.
Noted on rebasing to v0.29.0dev — will do and keep in touch. 

Have a great week,
Noel

CPBridge:
Sorry, is this PR ready for review now? What's the status? @oftalmos-org

oftalmos-org:
> Sorry, is this PR ready for review now? What's the status? @oftalmos-org

Hi Chris @CPBridge ,

Status update on 
[ImagingDataCommons/highdicom#407](https://github.com/ImagingDataCommons/highdicom/pull/407)
 — all three of your review items are now addressed, plus the rebase:

1. TID renumbering (2025b ratified standard)
Updated TID 6001/6004/6005 → TID 2120/2123/2124 throughout. All previously-provisional Sup247 draft codes now use their final ratified DCM/LOINC values (container codes 131242/131243, RNFL measurements 131264–131272, macular 131255, etc.), so the private 99OPHTHALMO scheme is gone entirely — no companion pydicom PR needed after all, since the standard shipped final code values directly. Laterality also moved from CID 244 to CID 247 per the ratified TID 2120 table.

2. Base class inheritance
OphthalmologyMeasurementsGroup now subclasses MeasurementsAndQualitativeEvaluations (TID 1501) directly, reusing FindingSite and TrackingIdentifier (TID 4108) rather than hand-rolling content items. One correction this required: the base class hardcodes TemplateIdentifier='1501' on the container, but this group actually invokes TID 2120, so I override it back after construction — otherwise the SR would misdeclare which template was used.

3. Conformance enforcement (Option B)
Implemented as discussed — a successful construction now guarantees a compliant object:

RNFLSectorMeasurementsGroup / RNFLClockfaceMeasurementsGroup split TID 2123's two group roles (sector method + CID 4283 values vs. fixed clockface method + CID 4284 positions), with sector_method validated against the 5 CID 4282 codes.
The bilateral RNFL-symmetry measurement (Row 7) is enforced per the standard's actual MC wording: required once both eyes are represented (across sector + clockface groups combined), rejected otherwise.
MacularMeasurementsGroup gets the same named-parameter treatment for TID 2124's ETDRS grid.

Rebase
One thing worth flagging: v0.29.0dev looks like it's fallen behind master — it's still just one commit past v0.28.0dev (the version bump to 0.29.0, from 2026-07-23), while master has kept moving since. Rebasing onto v0.29.0dev literally left the PR conflicting against master (a two-file version-string conflict). So I rebased onto master directly instead — it already contains everything v0.29.0dev has via v0.28.0dev, and the PR now shows mergeable with no conflicts.

Still open from your review: I haven't implemented TID 1002 Observer Context on the two root templates, or the optional Row 9-16 extras (TABLE, image quality, visual explanation, comment). Happy to add any of those if you'd like them before merge — otherwise I'd treat them as follow-up.

Let me know if the Option B design choices above look right, particularly the strict symmetry enforcement — happy to relax it if that's too aggressive.

Thanks again for your patience on this one.

Noel

PR #410, opened 2026-05-18
Volume dtype restriction

mccle:
Added checks to the volume init and setter methods to ensure volumes have float, integer, or bool types. A new test was added to verify that volumes cannot be instantiated with an array of an unsupported data type nor changed after instantiation to a new data type that is unsupported.

### DIFF
--- a/src/highdicom/volume.py
+++ b/src/highdicom/volume.py
@@ -2871,8 +2871,8 @@ class Volume(_VolumeBase):
             [np.floating, np.integer, np.bool_]
         ]):
             raise ValueError(
-                "Array must have a dtype of float, integer,"
-                f" or bool, received '{value.dtype}'."
+                "Argument 'array' must have an integer, floating point, "
+                f" or boolean dtype, received '{array.dtype}'."
             )
 
         self._array = value
--- a/src/highdicom/volume.py
+++ b/src/highdicom/volume.py
@@ -2504,6 +2504,15 @@ class Volume(_VolumeBase):
                 "Argument 'array' must be at least three dimensional."
             )
 
+        if not any([
+            np.issubdtype(array.dtype, dtype) for dtype in
+            [np.floating, np.integer, np.bool_]
+        ]):
+            raise ValueError(
+                "Argument 'array' must have a dtype of float, integer,"
+                f" or bool, received '{array.dtype}'."
+            )
+
         if channels is None:
             channels = {}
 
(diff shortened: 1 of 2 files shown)


src/highdicom/volume.py
```diff
@@ -2504,6 +2504,15 @@ def __init__(
                 "Argument 'array' must be at least three dimensional."
             )
 
+        if not any([
+            np.issubdtype(array.dtype, dtype) for dtype in
+            [np.floating, np.integer, np.bool_]
+        ]):
+            raise ValueError(
+                "Argument 'array' must have a dtype of float, integer,"
+                f" or bool, received '{array.dtype}'."
```
CPBridge: This suggests that the dtype must be one of these options exactly. How about:
 
```suggestion
                "Argument 'array' must have an integer, floating point, "
                f" or boolean dtype, received '{array.dtype}'."
```

src/highdicom/volume.py
```diff
@@ -2856,6 +2865,16 @@ def array(self, value: np.ndarray) -> None:
...
                 "Array must match the shape of the existing array."
             )
+
+        if not any([
+            np.issubdtype(value.dtype, dtype) for dtype in
+            [np.floating, np.integer, np.bool_]
+        ]):
+            raise ValueError(
+                "Array must have a dtype of float, integer,"
+                f" or bool, received '{value.dtype}'."
```
CPBridge: As above

```suggestion
                "Argument 'array' must have an integer, floating point, "
                f" or boolean dtype, received '{array.dtype}'."
```

tests/test_volume.py
```diff
@@ -1341,3 +1341,77 @@ def test_match_geometry_segmentation():
...
+            str
+        ]:
+            with pytest.raises(
+                ValueError,
+                match=(
+                    "Array must have a dtype of float, integer,"
+                    " or bool, received"
+                )
+            ):
+                volume.array = np.zeros((10, 10, 10), dtype=disallowed_dtype)
```
CPBridge: What does this test? Isn't it the same regardless of `dtype`, and therefore run loads of times unnecessarily?

Would it be cleaner to have two tests: one for valid dtypes and the other for invalid dtypes?

CPBridge (changes requested):
A couple of minor suggestions

tests/test_volume.py
mccle: Assuming that the starting dtype is unimportant for array reassignment, it should be safe to avoid nesting them. I have split them into separate tests for the init dtypes and for the reassign dtypes.

CPBridge (approved):

PR #411, opened 2026-05-19
Add nibabel volume conversions and tests

mccle:


src/highdicom/volume.py
```diff
@@ -3892,6 +3892,136 @@ def from_itk(
...
+                )
+                array = array.astype(np.float64)
+
+            else:
+                raise ValueError(
+                    'NiBabel does not support float128 data.'
+                    ' Casting to float64 is not possible.'
+                )
+
+        nifti = nib.Nifti1Image(
```
mccle: nibabel has several other image classes that I believe could represent a volume, listed [here](https://github.com/nipy/nibabel/blob/master/nibabel/imageclasses.py#L33). Not all of them are relevant, but the MINC and MGH formats seem worth consideration. I think at the very least, we should support both `Nifti1Image` and `Nifti2Image`, meaning the `to_nib` method should already be changed to allow the user to specify the class to which they wish to convert. Additionally, several of these classes can accept an array, affine, and dtype exactly as the `Nifti1Image` class or with extremely minor changes.

However, I do not think all of the formats support the same data types. Are you interested in supporting other formats @CPBridge? If so, would it be best to group different classes into separate `to_*` methods or have any branching logic be included in a single `to_nib` method?

pyproject.toml
```diff
@@ -50,6 +50,7 @@ libjpeg = [
 ]
 itk = ["itk>=5.4.0"]
 sitk = ["SimpleITK>=2.2.1"]
+nib = ["nibabel>=4.0.0"]
```
CPBridge: I think it would be clearer to use the full name


```suggestion
nibabel = ["nibabel>=4.0.0"]
```

src/highdicom/volume.py
```diff
@@ -3892,6 +3920,238 @@ def from_itk(
...
+    def to_nibabel(
+        self,
+        image_class: Literal[
+            'Nifti1Image',
+            'Nifti2Image',
+            'MGHImage',
+            'Minc1Image',
+            'Minc2Image',
+            'AnalyzeImage'
+        ] = 'Nifti1Image'
```
CPBridge: Elsewhere in the library, we use `enum.Enum`s for a defined set of options. I think it would be better to stick to that convention here. `Literal` is only useful for type checking. It does nothing at runtime.

src/highdicom/volume.py
```diff
@@ -3892,6 +3920,238 @@ def from_itk(
...
+    ) -> 'nibabel.spatialimages.SpatialImage':  # noqa: F821
+        """Convert the volume to a ``nibabel.spatialimages.SpatialImage``
+        format.
+
+        The Volume is converted to one of several 3D Image classes (
+        ``nibabel.Nifti1Image``, ``nibabel.Nifti2Image``, ``nibabel.MGHImage``,
+        ``nibabel.Minc1Image``, ``nibabel.Minc2Image``, ``nibabel.AnalyzeImage``
+        ), defaulting to ``nibabel.Nifti1Image``. If its array'current datatype
```
CPBridge: Typo

```suggestion
        ), defaulting to ``nibabel.Nifti1Image``. If its array's current datatype
```

src/highdicom/volume.py
```diff
@@ -3892,6 +3920,238 @@ def from_itk(
...
+            raise ValueError(
+                'NiBabel conversion does not currently support'
+                ' volumes with multiple channels.'
+            )
+
+        if np.issubdtype(self.dtype, np.bool_):
+            self = self.astype(np.uint8)
+
+        dtype = self.dtype
+        image_class = getattr(nib, image_class)
```
CPBridge: I would guard this with a check on the values, and issue a more helpful error message if an inappropriate value is passed. As noted above, this would probably be better handled with an enum (there are examples of this all over the library, for example the [segmentation_type](https://github.com/ImagingDataCommons/highdicom/blob/a4c27ddedc390d5520b460e23f7caef7ed351b23/src/highdicom/seg/sop.py#L172) parameter of the Segmentation constructor)

src/highdicom/volume.py
```diff
@@ -3892,6 +3920,238 @@ def from_itk(
...
+                np.uint32: np.int32,
+                np.uint64: np.int32,
+                np.int8: np.int16,
+                np.int64: np.int32,
+                np.float16: np.float32,
+                np.float128: np.float64
+            }
+
+        else:
+            dtype_map = {}
```
CPBridge: This will just cause confusion downstream I would think

src/highdicom/volume.py
```diff
@@ -3892,6 +3920,238 @@ def from_itk(
...
+        coordinate_system: highdicom.CoordinateSystemNames | str
+            Coordinate system (``"PATIENT"`` or ``"SLIDE"``) in which the volume
+            is defined.
+        frame_of_reference_uid: Union[str, None], optional
+            Frame of reference UID for the frame of reference, if known.
+
+        Returns
+        -------
+        highdicom.Volume:
+            Volume constructed from the `itk.Image`.
```
```suggestion
            Volume constructed from the Nibabel image.
```

src/highdicom/volume.py
```diff
@@ -3892,6 +3920,238 @@ def from_itk(
...
+            self.get_affine('RAS'),
+            header=header
+        )
+
+        return image
+
+    @classmethod
+    def from_nibabel(
+        cls,
+        nib_im: 'nibabel.spatialimages.SpatialImage',  # noqa: F821
```
CPBridge: We have generally been verbose with parameter names


```suggestion
        nibabel_image: 'nibabel.spatialimages.SpatialImage',  # noqa: F821
```

src/highdicom/volume.py
```diff
@@ -3892,6 +3920,238 @@ def from_itk(
...
+
+        Returns
+        -------
+        highdicom.Volume:
+            Volume constructed from the `itk.Image`.
+
+        Raises
+        ------
+        ValueError
+            When the volume is not 3D (multiple channels are unsupported).
```
CPBridge: It's not a volume at this stage

```suggestion
            When the input image is not 3D (multiple channels are unsupported).
```

tests/test_nib.py
```diff
@@ -0,0 +1,1283 @@
...
+    pytest.skip("Optional dependency not available", allow_module_level=True)
+
+DCM_QA_MPRAGE = 'https://github.com/neurolabusc/dcm_qa_mprage/raw/refs/heads/main'  # noqa: E501
+DCM_QA_ME = 'https://github.com/neurolabusc/dcm_qa_me/raw/refs/heads/master'
+DCM_QA_PDT2 = 'https://github.com/neurolabusc/dcm_qa_pdt2/raw/refs/heads/main'
+
+
+def read_multiframe_ct_volume():
+    im = imread(get_testdata_file('eCT_Supplemental.dcm'))
+    return im.get_volume()
```
CPBridge: There's some repetition in some of these functions I think. It might make sense to factor these out into the `tests/utils.py` file?

CPBridge (changes requested):
Few suggestions

src/highdicom/volume.py
mccle: This branch covers Minc1Image and Minc2Image, as the dtypes are all supported and do not need to be remapped. Would `elif image_class in [nib.Minc1Image, nib.Minc2Image]` be preferable as it is more explicit?

tests/test_nib.py
mccle: Ok, I can factor a few of these out from the sitk and itk tests as well. I believe some of them may be in the original volume tests as well.

mccle: Should I also update these in this PR or keep it focused on purely nibabel and standardize these changes in a followup PR after all of the initial external dependency PRs are merged?

src/highdicom/volume.py
CPBridge: Ah okay, that's fine, I misunderstood

tests/test_nib.py
CPBridge: I think this PR is fine unless you think that will create problems

CPBridge (approved):

PR #422, opened 2026-06-23
Treat None Rescale Intercept/Slope as missing without raising exception #418

### LINKED ISSUE #418: Exception when RescaleSlope/Intercept are present but empty (null)
I've run into a problem when a Dicom file contains both `RescaleSlope` and `RescaleIntercept`, but because they're empty, `highdicom` fails with an exception:
```
     if (
         'RescaleSlope' in sub_ds or
         'RescaleIntercept' in sub_ds
         ):
                 modality_slope_intercept = (
>                   float(sub_ds.get('RescaleSlope', 1.0)),
>                   float(sub_ds.get('RescaleIntercept', 0.0))
                  )
 TypeError: float() argument must be a string or a real number, not 'NoneType'

src/highdicom/image.py:631/632: TypeError
```

To reproduce (from repo project root):
```bash
cp data/test_files/ct_image.dcm data/test_files/ct_image_blank_rescale.dcm
dcmodify -m "(0028,1052)=" data/test_files/ct_image_blank_rescale.dcm

dcmdump data/test_files/ct_image_blank_rescale.dcm | grep Rescale
# (0028,1052) DS (no value available)                     #   0, 0 RescaleIntercept
# (0028,1053) DS [1]                                      #   2, 1 RescaleSlope
```
running pytests for `tests/test_image.py` will then fail with the exception.

The cause is that the tag is present and `sub_ds.get(..., ...)` will return `None` rather than the default value.

The quickest fix would be  along the lines:
```python
rescale_slope = sub_ds.get('RescaleSlope', None)
rescale_intercept  = sub_ds.get('RescaleIntercept', None)

if rescale_slope or rescale_intercept:
    modality_slope_intercept = (
        float(rescale_slope or 1.0),
        float(rescale_intercept or 0.0)
    )
```

but maybe at this point you'd want to guard against not just empty values but also non-parseable values.

Happy to submit a PR, just let me know whether something quick like above is fine, or you'd prefer something more "proper" (maybe a helper `get_float_or_none()` which deals with edge cases.)

realcundo:
Treat None Rescale Intercept/Slope as missing without raising exception. 

Fixes: #418

I wasn't sure what kind of tests you prefer, so I wrote one that expects parsing to succeed without exceptions and  `_effective_slope_intercept` to be `None` for all combinations of missing/None rescale slope and intercept tags.

Confirmed this test was failing with exception before the code fix.

Happy to update if you think this should be tested differently.

### DIFF
--- a/src/highdicom/image.py
+++ b/src/highdicom/image.py
@@ -623,14 +623,18 @@ class _CombinedPixelTransform:
                         else:
                             sub_ds = ds
 
-                        if (
-                            'RescaleSlope' in sub_ds or
-                            'RescaleIntercept' in sub_ds
-                        ):
+                        # Rescale intercept and slope should not be empty,
+                        # but this has been observed in some data and
+                        # should be handled by assuming slope=1, intercept=0
+                        rescale_slope = sub_ds.get('RescaleSlope', None)
+                        rescale_intercept = sub_ds.get('RescaleIntercept', None)
+
+                        if rescale_slope or rescale_intercept:
                             modality_slope_intercept = (
-                                float(sub_ds.get('RescaleSlope', 1.0)),
-                                float(sub_ds.get('RescaleIntercept', 0.0))
+                                float(rescale_slope or 1.0),
+                                float(rescale_intercept or 0.0)
                             )
+
                             self.applies_to_all_frames = (
                                 self.applies_to_all_frames and is_shared
                             )
--- a/tests/test_image.py
+++ b/tests/test_image.py
@@ -1017,6 +1017,31 @@ def test_combined_transform_pmap_rwvm_lut():
         tf = _CombinedPixelTransform(pmap, output_dtype=np.float32)
 
 
+def test_empty_rescale_slope_intercept_behaves_as_missing():
+    # Construct a CT image with None RescaleSlope/RescaleIntercept
+    file_path = Path(__file__)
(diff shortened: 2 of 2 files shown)


src/highdicom/image.py
```diff
@@ -623,14 +623,15 @@ def __init__(
...
                             sub_ds = ds
 
-                        if (
-                            'RescaleSlope' in sub_ds or
-                            'RescaleIntercept' in sub_ds
-                        ):
+                        rescale_slope = sub_ds.get('RescaleSlope', None)
+                        rescale_intercept = sub_ds.get('RescaleIntercept', None)
+
+                        if rescale_slope or rescale_intercept:
```
CPBridge, line 632: Minor request, can we leave a comment about why we did this 


```suggestion
                        if rescale_slope or rescale_intercept:
                            # Rescale intercept and slope should not be empty,
                            # but this has been observed in some data and
                            # should be handled by assuming slope=1,
                            # intercept=0

```

CPBridge (changes requested):
One minor thing

src/highdicom/image.py
realcundo: Applied verbatim (at the top of the block rather than in the `if` body .

CPBridge (approved):
Thanks @realcundo

PR #431, opened 2026-07-25
Fix native floating-point frame decoding

yinkev:
## Summary

- decode native `FloatPixelData` and `DoubleFloatPixelData` frames through `ImageFileReader`
- pass the detected pixel-data keyword into the shared frame decoder
- avoid requiring integer-only `BitsStored` and `PixelRepresentation` attributes for floating-point pixel modules
- add multi-frame float32 and float64 regression coverage, including negative fractional values

## Root cause

`ImageFileReader` detected floating-point pixel tags while scanning the file, but `read_frame()` discarded that information. It then required integer Pixel Data attributes and called `decode_frame()` with its integer default keyword, so valid floating-point DICOM images failed before decoding.

## Validation

- regression fails on `upstream/master` with missing `BitsStored`
- `python -B -m pytest tests/test_io.py::TestImageFileReader::test_read_float_pixel_data -q`
- `python -B -m flake8 src/highdicom/io.py tests/test_io.py`
- `git diff --check`

### DIFF
--- a/src/highdicom/io.py
+++ b/src/highdicom/io.py
@@ -31,6 +31,7 @@ from pydicom.tag import (
 )
 from pydicom.uid import UID, DeflatedExplicitVRLittleEndian
 
+from highdicom.enum import PixelDataKeywords
 from highdicom.frame import decode_frame
 from highdicom.color import ColorManager
 from highdicom.uid import UID as hd_UID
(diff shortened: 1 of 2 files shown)


yinkev:
Replaced by #441 to conform to this repository's documented contributor workflow and branch-naming convention. The replacement uses the same signed commit and identical code diff: https://github.com/ImagingDataCommons/highdicom/pull/441

PR #432, opened 2026-07-25
Fix per-channel min-max normalization for 4-D volumes

yinkev:
## Summary

- honor `per_channel=True` when a volume has exactly one channel dimension
- normalize each channel independently across the three spatial axes
- add regression coverage for channels with substantially different intensity ranges

## Root cause

`normalize_min_max()` enabled channel-wise reduction only when `number_of_channel_dimensions > 1`. A standard 4-D volume has exactly one channel dimension, so it silently fell back to global normalization despite the explicit option.

## Validation

- regression fails against `upstream/master`
- full `tests/test_volume.py`: 117 passed
- Flake8 on touched files
- `git diff --check`

### DIFF
--- a/src/highdicom/volume.py
+++ b/src/highdicom/volume.py
@@ -3428,7 +3428,7 @@ class Volume(_VolumeBase):
 
         if (
             per_channel and
-            self.number_of_channel_dimensions > 1
+            self.number_of_channel_dimensions > 0
         ):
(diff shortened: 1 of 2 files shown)


yinkev:
Replaced by #442 to conform to this repository's documented contributor workflow and branch-naming convention. The replacement uses the same signed commit and identical code diff: https://github.com/ImagingDataCommons/highdicom/pull/442

PR #433, opened 2026-07-25
Preserve dtype in per-channel volume padding

yinkev:
## Summary

- preserve the source volume dtype when padding independently per channel
- add `uint16` and `float32` regression coverage
- verify distinct channel-wise minimum padding values and unchanged interior voxels

## Root cause

The per-channel padding path preallocated its output with `np.zeros()` without a dtype, which silently promoted all non-`float64` volumes to `float64`.

## Validation

- regression fails against `upstream/master` for both tested dtypes
- pad subset: 26 passed
- Flake8 on touched files
- `git diff --check`

### DIFF
--- a/src/highdicom/volume.py
+++ b/src/highdicom/volume.py
@@ -3646,7 +3646,10 @@ class Volume(_VolumeBase):
                 for s, (p1, p2) in zip(self.spatial_shape, full_pad_width)
             ]
             # preallocate output array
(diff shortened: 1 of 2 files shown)


yinkev:
Replaced by #443 to conform to this repository's documented contributor workflow and branch-naming convention. The replacement uses the same signed commit and identical code diff: https://github.com/ImagingDataCommons/highdicom/pull/443

PR #434, opened 2026-07-25
Validate ImageFileReader frame indices

yinkev:
## Summary

- reject negative frame indices and indices equal to `NumberOfFrames`
- validate before indexing the offset table or seeking into pixel data
- document the `ValueError` contract on raw and decoded frame access

## Root cause

The previous condition used `index > number_of_frames`. It therefore allowed Python negative indexing and missed the first invalid upper-bound value, `index == number_of_frames`.

## Validation

- regression fails on `upstream/master`
- Flake8 on touched files
- `git diff --check`

### DIFF
--- a/src/highdicom/io.py
+++ b/src/highdicom/io.py
@@ -693,12 +693,18 @@ class ImageFileReader:
 
         Raises
         ------
+        ValueError
+            When `index` is less than 0 or greater than or equal to the number
+            of frames.
         OSError
             When frame could not be read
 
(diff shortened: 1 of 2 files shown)


yinkev:
Replaced by #444 to conform to this repository's documented contributor workflow and branch-naming convention. The replacement uses the same signed commit and identical code diff: https://github.com/ImagingDataCommons/highdicom/pull/444

PR #435, opened 2026-07-25
Reject non-invertible volume affines

yinkev:
## Summary

- reject non-invertible affine matrices during `Volume` and `VolumeGeometry` construction
- retain the existing scaled-orthogonality and final-row validation
- clarify the constructor contract and add coverage for both public types

## Root cause

The approximate orthogonality check can accept rank-deficient matrices whose dependent column is sufficiently small. Those objects then fail later when spatial operations attempt to invert the affine.

## Validation

- regression fails for both constructors on `upstream/master`
- Flake8 on touched files
- `git diff --check`

### DIFF
--- a/src/highdicom/volume.py
+++ b/src/highdicom/volume.py
@@ -272,9 +272,10 @@ class _VolumeBase(ABC):
         affine: numpy.ndarray
             4 x 4 affine matrix representing the transformation from voxel
             indices to the frame-of-reference coordinate system. The top left 3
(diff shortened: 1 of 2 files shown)


yinkev:
Replaced by #445 to conform to this repository's documented contributor workflow and branch-naming convention. The replacement uses the same signed commit and identical code diff: https://github.com/ImagingDataCommons/highdicom/pull/445

PR #441, opened 2026-07-25
Fix native floating-point frame decoding

### LINKED ISSUE #436: ImageFileReader fails to decode native floating-point pixel data
## Problem

`ImageFileReader` detects `FloatPixelData` and `DoubleFloatPixelData` while scanning a DICOM file, but `read_frame()` later applies integer Pixel Data assumptions. Valid floating-point images fail because they do not provide integer-only attributes such as `BitsStored` and `PixelRepresentation`.

## Expected behavior

Native float32 and float64 frames should be decoded through the shared frame decoder with their detected pixel-data keyword, preserving values and dtype.

## Reproduction

A two-frame Parametric Map using `FloatPixelData` or `DoubleFloatPixelData` fails before decoding with a missing `BitsStored` error.

## Proposed scope

Propagate the detected pixel-data keyword into frame decoding and add multi-frame float32/float64 regression coverage.

yinkev:
Closes #436.

## Summary

- decode native `FloatPixelData` and `DoubleFloatPixelData` frames through `ImageFileReader`
- pass the detected pixel-data keyword into the shared frame decoder
- avoid requiring integer-only `BitsStored` and `PixelRepresentation` attributes for floating-point pixel modules
- add multi-frame float32 and float64 regression coverage, including negative fractional values

## Root cause

`ImageFileReader` detected floating-point pixel tags while scanning the file, but `read_frame()` discarded that information. It then required integer Pixel Data attributes and called `decode_frame()` with its integer default keyword, so valid floating-point DICOM images failed before decoding.

## Validation

- regression fails on `upstream/master` with missing `BitsStored`
- `python -B -m pytest tests/test_io.py::TestImageFileReader::test_read_float_pixel_data -q`
- repository-documented `pytest --flake8`: 2,022 passed, 178 expected optional-dependency skips
- `python -B -m flake8 src/highdicom/io.py tests/test_io.py`
- `git diff --check`

### DIFF
--- a/src/highdicom/io.py
+++ b/src/highdicom/io.py
@@ -31,6 +31,7 @@ from pydicom.tag import (
 )
 from pydicom.uid import UID, DeflatedExplicitVRLittleEndian
 
+from highdicom.enum import PixelDataKeywords
 from highdicom.frame import decode_frame
 from highdicom.color import ColorManager
 from highdicom.uid import UID as hd_UID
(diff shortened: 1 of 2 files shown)


CPBridge (approved):
Thanks for catching this

PR #442, opened 2026-07-25
Fix per-channel min-max normalization for 4-D volumes

### LINKED ISSUE #437: normalize_min_max ignores per_channel for one channel dimension
## Problem

`Volume.normalize_min_max(per_channel=True)` enables channel-wise reduction only when `number_of_channel_dimensions > 1`. A normal 4-D volume has exactly one channel dimension, so the explicit option silently falls back to global normalization.

## Expected behavior

When any channel dimension is present, `per_channel=True` should normalize each channel independently across the three spatial axes.

## Reproduction

Two channels with disjoint intensity ranges do not both map to `[0, 1]`; the second channel retains a nonzero minimum.

## Proposed scope

Correct the channel-dimension predicate, make per-channel range arithmetic safe for constant and full-range integer channels, and add focused 4-D regressions.

yinkev:
Closes #437.

## Summary

- honor `per_channel=True` when a volume has exactly one channel dimension
- normalize each channel independently across the three spatial axes
- make integer extrema/range arithmetic safe for constant and full-range channels
- add float32, constant-`uint16`, and full-range-`int16` regression coverage

## Root cause

`normalize_min_max()` enabled channel-wise reduction only when `number_of_channel_dimensions > 1`. In addition, its zero-range workaround wrote a possibly fractional output range into the input-typed peak-to-peak array, and signed integer subtraction could overflow before promotion.

## Validation

- all three regressions fail on the relevant pre-fix implementation
- repository-documented `pytest --flake8`: 2,023 passed, 179 expected optional-codec skips
- `git diff --check`

### DIFF
--- a/src/highdicom/volume.py
+++ b/src/highdicom/volume.py
@@ -3428,7 +3428,7 @@ class Volume(_VolumeBase):
 
         if (
             per_channel and
-            self.number_of_channel_dimensions > 1
+            self.number_of_channel_dimensions > 0
         ):
             imin = self.array.min(axis=(0, 1, 2), keepdims=True)
             imax = self.array.max(axis=(0, 1, 2), keepdims=True)
--- a/src/highdicom/volume.py
+++ b/src/highdicom/volume.py
@@ -3438,11 +3438,11 @@ class Volume(_VolumeBase):
 
             peak_to_peak = imax - imin
 
-            # Avoid division by zero without assigning a fractional output
-            # range into an integer array.
+            # Constant channels map to output_min; keep their inert scale at
+            # 1.0 while avoiding division by zero.
             safe_peak_to_peak = np.where(
                 peak_to_peak == 0.0,
-                1,
+                output_range,
                 peak_to_peak,
             )
             scale_factor = output_range / safe_peak_to_peak
--- a/src/highdicom/volume.py
+++ b/src/highdicom/volume.py
@@ -3432,12 +3432,20 @@ class Volume(_VolumeBase):
         ):
             imin = self.array.min(axis=(0, 1, 2), keepdims=True)
             imax = self.array.max(axis=(0, 1, 2), keepdims=True)
+            if np.issubdtype(self.array.dtype, np.integer):
+                imin = imin.astype(np.float64)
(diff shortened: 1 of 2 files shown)


src/highdicom/volume.py
```diff
@@ -3428,16 +3428,24 @@ def normalize_min_max(
...
-            # Avoid division by zerp
-            peak_to_peak[peak_to_peak == 0.0] = output_range
-            scale_factor = output_range / peak_to_peak
+            # Avoid division by zero without assigning a fractional output
+            # range into an integer array.
+            safe_peak_to_peak = np.where(
+                peak_to_peak == 0.0,
+                1,
+                peak_to_peak,
+            )
```
CPBridge, line 3447: Shouldn't the `1` here be `output_range` such that `scale_factor` ends up being 1.0 after the following line?


```suggestion
            safe_peak_to_peak = np.where(
                peak_to_peak == 0.0,
                output_range,
                peak_to_peak,
            )
```

CPBridge (commented):
Thanks for this @yinkev . I have one question here

src/highdicom/volume.py
CPBridge: Also, relatedly, I think we should include some tests of this behaviour with `output_range` that is not the "trivial" default of (0.0, 1.0)

yinkev: Yes — agreed. The constant channel still maps to output_min because its centered values are zero, but using output_range makes the inert scale factor exactly 1.0 and expresses the intent correctly. Updated in the latest commit. The focused local pytest setup is currently blocked by the repository dependency resolution conflict between requires-python >=3.10 and sphinx-autodoc-typehints 3.9.9 requiring >=3.12; CI will run the existing per-channel normalization cases.

yinkev: Added that coverage as well. I strengthened the constant-integer-channel regression to use output_min=-2.0 and output_max=3.0, then assert the constant channel maps entirely to -2.0 while the varying channel spans the full [-2.0, 3.0] range.

CPBridge (approved):
Thank you @yinkev looks good

PR #443, opened 2026-07-25
Preserve dtype in per-channel volume padding

### LINKED ISSUE #438: Volume.pad with per_channel promotes arrays to float64
## Problem

The per-channel branch of `Volume.pad()` preallocates its output with `np.zeros()` without specifying a dtype. `uint16`, `float32`, and other non-float64 volumes are silently promoted to float64.

## Expected behavior

Padding should preserve the source volume dtype while calculating channel-specific constant values.

## Reproduction

Padding a multi-channel `uint16` or `float32` volume with `mode=MINIMUM, per_channel=True` returns a float64 array.

## Proposed scope

Allocate the result using `self.array.dtype` and add integer and floating-point regressions.

yinkev:
Closes #438.

## Summary

- preserve the source volume dtype when padding independently per channel
- add `uint16` and `float32` regression coverage
- verify distinct channel-wise minimum padding values and unchanged interior voxels

## Root cause

The per-channel padding path preallocated its output with `np.zeros()` without a dtype, which silently promoted all non-`float64` volumes to `float64`.

## Validation

- regression fails against `upstream/master` for both tested dtypes
- pad subset: 26 passed
- Flake8 on touched files
- `git diff --check`

### DIFF
--- a/src/highdicom/volume.py
+++ b/src/highdicom/volume.py
@@ -3646,7 +3646,10 @@ class Volume(_VolumeBase):
                 for s, (p1, p2) in zip(self.spatial_shape, full_pad_width)
             ]
             # preallocate output array
-            new_array = np.zeros([*out_spatial_shape, *self.channel_shape])
+            new_array = np.zeros(
+                [*out_spatial_shape, *self.channel_shape],
+                dtype=self.array.dtype,
+            )
             for cind in itertools.product(
                 *[range(n) for n in self.channel_shape]
             ):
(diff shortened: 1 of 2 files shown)


CPBridge:
Did you mean to leave this as a draft? Looks like a complete solution @yinkev

yinkev:
Good catch. That was accidental. I’ve marked it ready for review now. Thanks for catching that!

CPBridge (approved):
Thanks @yinkev

PR #444, opened 2026-07-25
Validate ImageFileReader frame indices

### LINKED ISSUE #439: ImageFileReader accepts invalid frame indices
## Problem

The frame bounds check uses `index > number_of_frames`. It therefore allows negative Python indexing and fails to reject `index == number_of_frames` before offset-table access.

## Expected behavior

Valid frame indices are exactly `0 <= index < NumberOfFrames`; other values should raise a clear `ValueError` before seeking or decoding.

## Proposed scope

Correct the predicate, document the exception contract, and add lower- and upper-bound regressions.

yinkev:
Closes #439.

## Summary

- reject negative frame indices and indices equal to `NumberOfFrames`
- validate before indexing the offset table or seeking into pixel data
- document the `ValueError` contract on raw and decoded frame access

## Root cause

The previous condition used `index > number_of_frames`. It therefore allowed Python negative indexing and missed the first invalid upper-bound value, `index == number_of_frames`.

## Validation

- regression fails on `upstream/master`
- Flake8 on touched files
- `git diff --check`

### DIFF
--- a/src/highdicom/io.py
+++ b/src/highdicom/io.py
@@ -697,12 +697,18 @@ class ImageFileReader:
 
         Raises
         ------
+        ValueError
+            When `index` is less than 0 or greater than or equal to the number
+            of frames.
         OSError
             When frame could not be read
 
         """
-        if index > self.number_of_frames:
-            raise ValueError('Frame index exceeds number of frames in image.')
+        if index < 0 or index >= self.number_of_frames:
+            raise ValueError(
+                'Frame index must be greater than or equal to 0 and less '
+                'than the number of frames in the image.'
+            )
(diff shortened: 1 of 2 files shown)


CPBridge:
Is this ready for review? @yinkev

CPBridge:
Could you please resolve the merge conflicts too (I think you just want to keep both changes)?

yinkev:
Good catch, leaving it as a draft was accidental. I resolved the merge conflict by keeping both changes and marked it ready for review. Thanks for catching it.

CPBridge (approved):
Thanks @yinkev

PR #454, opened 2026-08-06
Add to_monai and from_monai methods

mccle:


src/highdicom/volume.py
```diff
@@ -4241,6 +4241,191 @@ def from_nibabel(
...
+
+        The Volume is converted to a ``monai.data.Metatensor``. Spatial metadata
+        is preserved through the affine array. By default, metatensors use "RAS"
+        convention, however they can support "LPS" similar to highdicom if
+        specified.
+
+        Parameters
+        ----------
+        convert_to_ras: bool
+            Whether to conver from 'LPS' to 'RAS' convention. Defaults to True.
```
```suggestion
            Whether to convert the affine matrix from 'LPS' to 'RAS' convention.
```

src/highdicom/volume.py
```diff
@@ -4241,6 +4241,191 @@ def from_nibabel(
...
+        meta[ImageStatsKeys.SPACING] = self.spacing
+        meta[MetaKeys.SPATIAL_SHAPE] = self.shape[:3]
+        meta[MetaKeys.ORIGINAL_AFFINE] = affine.copy()
+        meta[MetaKeys.AFFINE] = affine.copy()
+        meta[MetaKeys.ORIGINAL_CHANNEL_DIM] = (
+            -1 if self.ndim > 3 else float("nan")
+        )
+
+        metatensor = monai.data.MetaTensor(
+            self.array.copy(),
```
CPBridge, line 4311: Are you sure this copy is necessary? I imagine the conversion to metatensor copies anyway and the memcopy could be quite slow, so best to avoid it if possible

src/highdicom/volume.py
```diff
@@ -4241,6 +4241,191 @@ def from_nibabel(
...
+            space = monai.utils.enums.SpaceKeys.RAS
+
+        else:
+            space = monai.utils.enums.SpaceKeys.LPS
+
+        affine = self.get_affine(space.value)
+
+        meta[MetaKeys.SPACE] = space
+        meta[ImageStatsKeys.SPACING] = self.spacing
+        meta[MetaKeys.SPATIAL_SHAPE] = self.shape[:3]
```
CPBridge: There's already a property for this

```suggestion
        meta[MetaKeys.SPATIAL_SHAPE] = self.spatial_shape
```

src/highdicom/volume.py
```diff
@@ -4241,6 +4241,191 @@ def from_nibabel(
...
+            -1 if self.ndim > 3 else float("nan")
+        )
+
+        metatensor = monai.data.MetaTensor(
+            self.array.copy(),
+            meta=meta
+        )
+
+        if ensure_channel_first:
+            metatensor = monai.transforms.EnsureChannelFirst()(metatensor)
```
CPBridge, line 4316: If there is a channel dim in the Volume and ensure_channel_first is False, the channel is placed last in the metatensor. I'm not sure this makes sense since torch never uses this format. I think it would be more expected behavior to move it first

src/highdicom/volume.py
```diff
@@ -4241,6 +4241,191 @@ def from_nibabel(
             from_reference_convention='RAS'
         )
 
+    def to_monai(
+        self,
+        convert_to_ras: bool = True,
+        ensure_channel_first: bool = False,
+    ) -> 'monai.data.MetaTensor':  # noqa: F821
```
CPBridge, line 4248: I think we should probably add an option to squeeze the first spatial dimension (usually the slice dimension) if it is singleton

src/highdicom/volume.py
CPBridge: Also, this returns a tuple but i notice that the nibabel reader populates this as a 1D np.nparray, so we should probably match that behavior?

src/highdicom/volume.py
```diff
@@ -4241,6 +4241,191 @@ def from_nibabel(
...
+        )
+
+        array = metatensor.cpu().detach().numpy()
+
+        if array.ndim > 3:
+            if array.ndim > 4:
+                raise ValueError(
+                    'Monai conversion does not currently support'
+                    ' volumes with multiple channel dimensions.'
+                )
```
CPBridge, line 4399: this test could be moved before the detach and numpy conversion

tests/test_monai.py
```diff
@@ -0,0 +1,695 @@
...
+    DCM_QA_ME,
+    DCM_QA_PDT2,
+    TEST_DATA,
+    read_multiframe_ct_volume,
+    read_ct_series_volume,
+    urldownload_with_retry
+)
+
+
+try:
```
CPBridge, line 28: Do you need to test for ITK too, since you are using the ITKReader?

tests/test_monai.py
```diff
@@ -0,0 +1,695 @@
+import numpy as np
```
CPBridge, line 1: Should probably add a test with `ensure_channels_first=True` for both a volume with and without a channel dim to check it deals with them correctly

CPBridge (changes requested):

tests/test_monai.py
mccle: The tests are currently run with none of the optional dependencies or with all of them, so there should not be a situation where the test is run and monai is installed but itk is not.

src/highdicom/volume.py
mccle: This seems to be consistent with the default behavior that MONAI uses for data loading. I saved a 4D NIfTI with NiBabel since it is also channel last and the `LoadImage` transform also produces a channel last metatensor without specifying `ensure_channel_first=True`.

mccle: It is a more sensible default in my opinion, but I think it should follow the behavior of the corresponding MONAI features.

src/highdicom/volume.py
mccle: Ok, I will change this to:

```suggestion
        meta[MetaKeys.SPATIAL_SHAPE] = np.array(self.spatial_shape)
```

Issue #459, opened 2026-09-04
Clarify the meaning of TrackingUID in the docs


fedorov:
Explain how it can be used to track the same finding/lesion over time or across sequences.

CPBridge:
See https://github.com/ImagingDataCommons/highdicom/pull/460

PR #460, opened 2026-09-04
Improve description of tracking (U)ID

CPBridge:
Addresses #459 

The tracking ID is already described in SR and SEG pages of the User Guide, but this PR improves the docstrings in the API reference to make it clearer that tracking IDs are used to track ROIs across imaging studies.

### DIFF
--- a/src/highdicom/seg/content.py
+++ b/src/highdicom/seg/content.py
@@ -77,9 +77,13 @@ class SegmentDescription(Dataset):
             Information useful for identification of the algorithm, such
             as its name or version. Required unless the algorithm type is `MANUAL`
         tracking_uid: Union[str, None], optional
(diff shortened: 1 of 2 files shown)


fedorov:
Maybe it also makes sense to add an example somewhere? "Track segment across multiple studies" is a lot of terms that may not be very familiar. So, for example, we could say "As an example, segmentation of a lesion monitored for progression can be assigned the same TrackingUID, allowing to link and monitor segmentaton-derived tumor volume changes over time. As another example, TrackingUID and link segmentations of the same lesion in different imaging modalities within the same imaging study."

CPBridge:
Thanks @fedorov I adapted your suggestion to improve the user guide page on segmentations

fedorov (approved):

